PR #4, opened 2026-07-30
OSAC-1733: Consolidate fulfillment-service and osac-operator .github into root

eliorerz:
## Summary

After the subtree merge (#3), neither fulfillment-service's nor osac-operator's CI was actually running: GitHub Actions only scans workflows at the repo root, and the merge preserved each component's `.github/` exactly where it was, under its own subdirectory. This PR moves everything to root and fixes it up so it actually works against the merged tree.

- Move all workflow/script/composite-action files to root `.github/`, prefixed by component where they must coexist (`Integration Tests`, `Check Floating Image Tags` -- these are genuinely different per component), deduped to one copy where they were already redundant (`ok-to-test-label-cleanup`, `slash-command`, `scan-workflow-logs` -- none of these were ever component-specific; keeping two would have double-run the same bot logic on every PR/comment).
- Rewrite every `paths:`/`paths-ignore:` filter to be prefixed with `fulfillment-service/` or `osac-operator/` so they match real paths in the merged tree instead of firing on any change anywhere in the repo.
- Fix every job's working-directory/CWD assumptions -- `ginkgo run`, `ruff`, `buf generate`, `goreleaser workdir`, `docker build` context/file, `make -C`, `go-version-file`, `hashFiles` cache keys -- so they actually build/lint/test the right component's code, not the repo root.
- Rework the shared `setup-go` composite action to accept a `working-directory` input, since composite action steps don't inherit a calling job's `defaults.run.working-directory`.
- Remove fulfillment-service's and osac-operator's now-redundant pre-commit workflows/configs -- root's already covers both with correctly scoped excludes (from the earlier CI fix on #3).
- Merge `dependabot.yml` into one file with per-component `gomod`/`docker` entries.
- Consolidate the two near-duplicate `e2e-bmaas`/`e2e-vmaas` workflow pairs into one each, building **both** components' current code together instead of testing one against the other's stale pinned image -- this is the actual bug the mono-repo is supposed to fix. Needs the companion `osac-test-infra` PR below.

## Dependency

Requires **osac-project/osac-test-infra#283** (adding two-component build support to the reusable e2e workflows) to land first, or the new `e2e-bmaas-full-install.yml`/`e2e-vmaas-full-install.yml` here will fail on the unrecognized `component2-*` inputs.

## Deliberately out of scope

Per OSAC-1734/OSAC-3227: no workflow here is a required branch-protection check, and none of the path filters use `dorny/paths-filter`-style skip logic yet -- both are intentionally deferred until the full path-based CI matrix lands (after osac-aap/osac-installer also merge).

## Verification
- [x] `python -c "import yaml; yaml.safe_load(...)"` over every touched file -- all valid.
- [x] `pre-commit run --all-files` passes clean.
- [x] Grepped every renamed `name:`/`workflow_run.workflows:` pair to confirm producer/consumer names match exactly, and that `scan-workflow-logs.yml`'s watched names match the new consolidated e2e workflow names.
- [ ] CI on this PR itself is the first real test that each component's own suite actually fires and passes against current code -- watching it now.
- [ ] The new e2e workflows can't be verified end-to-end until osac-test-infra#283 merges and this is rebased/re-run.


eliorerz:
Follow-up commit: matrixed the three identical-shape per-component checks (`check-floating-tags`, `check-generated-code`, `helm-lint`) into one job each instead of duplicating per component -- scales to osac-aap/osac-installer/bare-metal-fulfillment-operator as matrix entries, not new files. Also added fulfillment-service's chart to helm-lint (previously unlinted). Left the image-build workflows and the genuinely-different-toolchain lint/test jobs (ruff/buf-lint/ginkgo vs golangci-lint/Makefile/Kind) split per component -- see the latest commit message for why.

eliorerz:
Follow-up: dropped component-name prefixes and merged the two cases where they existed only to avoid a collision (Integration Tests, Publish charts) into one file each. Everything else that never actually collided got its prefix stripped and filename reverted to the original name. Also reverted fulfillment-service from helm-lint.yaml -- its chart can't be linted bare (guards most values with `required(...)`), tracked as a separate follow-up rather than guessed at here.

PR #14, opened 2026-07-31
OSAC-1733: Split e2e-full-install.yml into 3 independently-runnable workflows

eliorerz:
## Summary

Splits e2e testing into three separate workflow files -- `e2e-bmaas-full-install.yml`, `e2e-vmaas-full-install.yml`, and a new `e2e-caas-full-install.yml` -- instead of one merged file with multiple jobs. Separate files can be independently re-triggered/re-run in the GitHub Actions UI, which matters because each e2e run reprovisions a whole cluster: with jobs sharing one file there's no way to retest just one failed suite without also re-running the others.

- Added the twice-daily `schedule: cron: '0 */12 * * *'` trigger to all three files (previously only VMaaS had it).
- Added the new `e2e-caas-full-install.yml`, calling `osac-project/osac-test-infra/.github/workflows/e2e-caas-full-install.yml@main`. CaaS-specific inputs (`cluster-template`, `caas-domain`, `flavor-image`/`flavor-name`, `namespace`, `installer-repo`/`installer-ref`, etc.) are left at the reusable workflow's own defaults since none are required.
- Switched all three callers from the old `component-repo`/`component-ref`/`component-containerfile`/`component-image-key`/`component2-*` inputs to the new `components` JSON-array input shape.
- Updated `scan-workflow-logs.yml`'s `workflow_run.workflows` list to the three workflow names.

**This PR depends on [osac-project/osac-test-infra#283](https://github.com/osac-project/osac-test-infra/pull/283) merging first** (adds the `components` input this PR uses) -- do not merge before that lands.

## Test plan
- [ ] `osac-test-infra#283` merges, adding the `components` input to the reusable e2e workflows
- [x] YAML syntax validated (`yaml.safe_load`) for all four touched/new files
- [ ] CI runs green for all three e2e workflows on this PR

## Summary by CodeRabbit

* **Tests**
  * Added automated end-to-end coverage for full CaaS installations across pull requests, merge queues, scheduled runs, and manual launches.
  * Improved full-install testing for BMaaS and VMaaS by consolidating component configuration.
  * Added clearer workflow status reporting so failed or cancelled test runs are surfaced reliably.

* **Chores**
  * Expanded automated log monitoring to include CaaS full-install test results.

### DIFF
--- a/.github/workflows/e2e-caas-full-install.yml
+++ b/.github/workflows/e2e-caas-full-install.yml
@@ -1,5 +1,5 @@
 ---
-name: E2E BMaaS Full Install
+name: E2E CaaS Full Install
 
 permissions:
   contents: read
@@ -13,9 +13,9 @@ on:
   workflow_dispatch:
     inputs:
       test-suite:
-        description: "Test suite (bmaas, or empty for all)"
+        description: "Test suite (caas, or empty for all)"
         required: false
-        default: "bmaas"
+        default: "caas"
       test-filter:
         description: "pytest -k filter"
         required: false
@@ -26,7 +26,7 @@ on:
         default: "main"
 
 concurrency:
-  group: e2e-bmaas-full-install-${{ github.event.pull_request.number || github.run_id }}
+  group: e2e-caas-full-install-${{ github.event.pull_request.number || github.run_id }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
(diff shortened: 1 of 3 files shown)


eliorerz:
@coderabbitai review

eliorerz:
Reviewed both actionable findings from the fresh `@coderabbitai review` pass against the current file content (not just the diff hunk):

1. **"Add a pinned SAST/SCA security gate"** (`e2e-caas-full-install.yml`): verified there is no SAST/SCA/CodeQL/dependency-scanning workflow anywhere in this repo today — checked all 15 files under `.github/workflows/`. This is a pre-existing repo-wide gap that predates this PR (the original single `e2e-full-install.yml` didn't have one either, and its sibling `e2e-bmaas-full-install.yml` / `e2e-vmaas-full-install.yml` still don't). Adding a security-scanning gate is out of scope for a pure workflow-split PR; if adopted it should land uniformly across all three files as its own change, not asymmetrically in just the new `caas` file.

2. **"Pin the reusable workflow to a full commit SHA"** (line 63, `@main`): verified `@main` is the pre-existing convention, not something this PR introduces — both `e2e-bmaas-full-install.yml` and `e2e-vmaas-full-install.yml` already called their reusable workflow via `@main` before this split (see commit c4805edc). The new `caas` file just follows the same established pattern for consistency with its two siblings. Changing only the new file to SHA-pin would create inconsistency across the three otherwise-identical callers; if this hardening is wanted, it should be applied uniformly to all three in a separate, deliberate change.

No code changes made — both findings are valid general suggestions but out of scope for this PR's purpose (splitting one workflow into three independently-runnable ones).

eliorerz:
@coderabbitai Re-verified both findings against the current file content directly (not just the diff):

1. **Pinned SAST/SCA security gate** — no SAST/SCA/CodeQL/dependency-scanning workflow exists anywhere in this repo today (checked all workflow files under `.github/workflows/`). This predates this PR — the original single `e2e-full-install.yml` never had one, and its siblings `e2e-bmaas-full-install.yml`/`e2e-vmaas-full-install.yml` still don't. Out of scope for a pure workflow-split PR; if adopted, it belongs in its own PR applied uniformly across all three files, not asymmetrically in just `caas`.

2. **Pin the `osac-test-infra` reference from `@main` to a commit SHA** — `@main` is the pre-existing convention here, not something this PR introduces. Both `e2e-bmaas-full-install.yml` and `e2e-vmaas-full-install.yml` already referenced their reusable workflow via `@main` before this split existed (commit c4805edc). The new `caas` file just matches its two siblings. Pinning only the new file would make the trio inconsistent — a separate, deliberate hardening change if wanted, applied to all three at once.

Not making code changes for either — both are out of scope for this PR's actual purpose (splitting one caller workflow into three independently-runnable ones).

PR #19, opened 2026-07-31
OSAC-3487: Consolidate per-component .ai-bot configs into repo root

eliorerz:
## Summary

`fulfillment-service/.ai-bot/` and `osac-operator/.ai-bot/` are leftovers from before the mono-repo merge, configuring the `jira-ai-issue-solver` bot. I checked the bot's actual source ([`flightctl/jira-ai-issue-solver`](https://github.com/flightctl/jira-ai-issue-solver), `repoconfig/config.go` + [`docs/repo-configuration.md`](https://github.com/flightctl/jira-ai-issue-solver/blob/main/docs/repo-configuration.md)) to confirm the discovery mechanism rather than guessing:

- `repoconfig.Load(dir)` reads exactly one `.ai-bot/config.yaml` at `dir`, where `dir` is the root of whatever the bot cloned for that repo entry (single-repo: the clone root; multi-repo: `workspace/<repo.Name>`, itself a full separate clone). There is **no** per-subdirectory or path-prefix-keyed config anywhere in the schema (`Config`, `RepoSettings`, `ProjectSettings`).
- Since fulfillment-service/ and osac-operator/ are now subdirectories of one cloned repo rather than separate clone roots, **neither per-component `.ai-bot/` directory is ever read by the bot** today.

This confirms the fix path the ticket anticipated: merge into a root-level `.ai-bot/`, since the bot has no monorepo-awareness to update on our side (that would require a change to `flightctl/jira-ai-issue-solver` itself).

## What changed

Merged both components' `config.yaml`, `container.json`, `instructions.md`, `new-ticket-workflow.md`, and `feedback-workflow.md` into one root `.ai-bot/`:

- `validation_commands` (a flat list of shell-command hints, no structured path-keying field) are scoped per component with explicit `cd fulfillment-service`/`cd osac-operator` prefixes — the same pattern the root pre-commit config already uses for its per-component local hooks.
- `instructions.md`, `new-ticket-workflow.md`, and `feedback-workflow.md` each open with a short dispatcher telling the AI to determine which component a ticket/PR touches and follow the matching section.
- `container.json` and the `pr`/`ai` sections of `config.yaml` were identical between components and merged without changes.
- The `imports` entry differed only in the install command's target argument (fulfillment-service passed `claude`; osac-operator passed none, defaulting to `cursor` per the install script despite osac-operator's own config targeting Claude). Kept `claude` since both components configure `ai.claude`, not Cursor — flagging this pre-existing mismatch for awareness, not fixing it as a separate concern.

## `$CLAUDE_PROJECT_DIR` — correction, this needs a live test (not resolved by this PR)

**Update:** the original version of this section claimed the launch-directory-only behavior for `CLAUDE_PROJECT_DIR`/committed `.claude/settings.json` was "documented behavior." That was wrong and has been corrected below — see [review discussion](https://github.com/osac-project/osac/pull/19#issuecomment-5138964067).

I re-fetched `code.claude.com/docs/en/hooks` and `code.claude.com/docs/en/settings` directly. What's actually documented:

- The Hooks page's entire definition of `CLAUDE_PROJECT_DIR` is "the project root" — no resolution algorithm, and it explicitly defers to the Settings page.
- The Settings page documents an explicit git-root resolution algorithm ("Claude Code reads and writes this file at the root of the git repository, resolved through worktrees to the main checkout...") — but only for `.claude/settings.local.json`, a different file than the committed `.claude/settings.json` these hooks actually live in.
- There is no official documentation stating a "launch directory only, no ancestor walk" rule for `CLAUDE_PROJECT_DIR` or the committed `settings.json` — my earlier claim to that effect was not actually backed by these docs and should not have been presented as confirmed.
- It's also undocumented whether a `.claude/settings.json` living in a subdirectory (e.g. `fulfillment-service/.claude/settings.json`) is even loaded at all by a session whose project root is the monorepo root — unlike `CLAUDE.md` and `.claude/rules/*.md`, which the docs explicitly describe as supporting monorepo-style nesting.

**This part of OSAC-3487 is therefore still open**, not resolved by this PR. It needs a live test: start a Claude Code session with cwd at the monorepo root, edit a file under `fulfillment-service/` to trigger one of its hooks, and observe whether the hook fires and what `$CLAUDE_PROJECT_DIR` actually expands to (repeat with cwd = `fulfillment-service/` as a control). Independently of the result, a more robust fix would stop relying on `$CLAUDE_PROJECT_DIR` in these hook commands at all — the hook's JSON payload already includes the tool's `file_path` (parsed via `jq` in both files), so the hook script could derive the correct component directory from that instead of trusting an environment variable whose monorepo resolution is unconfirmed.

## Test plan

- [x] `.ai-bot/config.yaml` parses as valid YAML
- [x] `.ai-bot/container.json` parses as valid JSON
- [x] `pre-commit run` (yamllint, check-json, etc.) passes on the changed files
- [ ] Manual: trigger the bot against this repo post-merge and confirm it picks up the root `.ai-bot/config.yaml`/`instructions.md` (can't be verified from within this repo alone)


eliorerz:
Reviewed the full diff plus the resulting tree on a checkout of this branch. The `.ai-bot/` consolidation itself is solid:

- Confirmed the core discovery-mechanism claim directly against `flightctl/jira-ai-issue-solver` source (`repoconfig/config.go`): `Load(dir)` really is a flat `dir + ".ai-bot/config.yaml"` join with no per-subdirectory or ancestor-walk logic, and no error surfaced when the file's absent (silent fallback to defaults). Matches this PR's header comment exactly.
- Diffed the merged `config.yaml`/`instructions.md`/`new-ticket-workflow.md`/`feedback-workflow.md`/`feedback-workflow.md` content against both deleted originals line-by-line — nothing lost, the component-dispatch structure is a faithful merge.
- `container.json` was identical between the two components and merged without changes (confirmed).
- Grepped the whole repo post-merge for `.ai-bot` — nothing left dangling outside the new root directory (no stale AGENTS.md/README/CI references to the old per-component paths).
- Both new files parse (`yaml.safe_load` / `json.load`). No merge conflicts against current `main`.

One finding on the second half of this PR — the `$CLAUDE_PROJECT_DIR` verification:

The description states as fact: *"Per Claude Code's documented behavior: project settings.json only loads from the exact directory a session is started in, and CLAUDE_PROJECT_DIR resolves to that same launch directory (it does not walk up to the git root)."*

I did a dedicated investigation into exactly this question earlier in this session (it's the other half of OSAC-3487) and pulled the two directly relevant official docs pages myself (`code.claude.com/docs/en/hooks` and `code.claude.com/docs/en/settings`). What I actually found:

- The Hooks page's entire definition of `CLAUDE_PROJECT_DIR` is "the project root" — it does not describe the resolution algorithm and explicitly defers to the Settings page.
- The Settings page *does* document an explicit git-root resolution algorithm — but for `.claude/settings.local.json` specifically ("Claude Code reads and writes this file at the root of the git repository, resolved through worktrees to the main checkout..."), which is a **different file** than the committed `.claude/settings.json` this PR's hooks actually live in.
- I found no official documentation stating the "exact launch directory only, no ancestor walk" rule for either `CLAUDE_PROJECT_DIR` or the committed `settings.json`. That specific phrasing traces to third-party/community pages, not an official source.

So this PR is stating an unverified (though plausible) claim as "documented behavior." It may well be correct in practice, but I couldn't confirm it against the actual docs, and if it's wrong, this PR's "no code change needed" conclusion for that half of OSAC-3487 would be leaving a live risk on the table while the PR record says it's resolved. Given the ticket itself asked to "verify" this, I'd suggest either citing the specific doc passage this claim is based on, or softening it to reflect that it's unconfirmed and still needs a live test (start a session at the monorepo root, edit a file under `fulfillment-service/`, check whether the hook fires and what `$CLAUDE_PROJECT_DIR` actually expands to).

Not blocking the `.ai-bot/` half, which is independently correct regardless of how the hooks question resolves — but worth fixing the description before this becomes the historical record on a claim I don't think holds up.

eliorerz:
You're right, and I've corrected the PR description.

I re-fetched both docs pages myself (`code.claude.com/docs/en/hooks` and `code.claude.com/docs/en/settings`) to check independently rather than take your word for it, and got the same result you did: the Hooks page defines `CLAUDE_PROJECT_DIR` only as "the project root" with no resolution algorithm, and the git-root-resolution language on the Settings page is scoped explicitly to `.claude/settings.local.json` — a different file than the committed `.claude/settings.json` these hooks live in. There's no official statement of a "launch directory only" rule anywhere I could find for `CLAUDE_PROJECT_DIR` or the committed `settings.json`.

That claim in my PR description was sourced from a subagent lookup that presented an inference (or a community-source claim) as "documented behavior" — I didn't independently verify it against the actual docs before writing it down, which is exactly the failure mode you caught. Your `/tmp/osac-worker-bot-configs-report.md` investigation is more careful and more correct than what I wrote; thanks for pulling the exact quotes.

Updated the PR description's second section to state plainly that this half of OSAC-3487 is unresolved and needs a live test (session at monorepo root, edit a file under `fulfillment-service/`, check whether the hook fires and what `$CLAUDE_PROJECT_DIR` expands to), plus your suggested more-robust fix (derive the component dir from the hook's `file_path` payload instead of trusting the env var). The `.ai-bot/` consolidation itself is unaffected by this and still stands as verified.

PR #28, opened 2026-07-31
OSAC-3481: Add CI dummy-values file for fulfillment-service chart

eliorerz:
## Summary

`fulfillment-service`'s Helm chart (`charts/service`) guards most values with `required "..." .Values.X` (`auth.controllerCredentials`, `auth.issuerUrl`, `certs.caBundle.configMap`, `idp.url`, `idp.credentials`, etc.), so `helm lint`/`helm template` fail immediately against its bare `values.yaml`. That's why `helm-lint.yaml`'s matrix currently only covers `osac-operator`'s charts, which have no such guards.

- Added `fulfillment-service/charts/service/ci-values.yaml`: CI-only placeholder values satisfying every `required` guard in the chart's templates. Not used for real deploys -- `values.yaml` remains the source of documented defaults/overrides.
- Also set a non-empty `database.connection` in that file: leaving it at its default empty list hits a separate, pre-existing bug in `grpc-server/deployment.yaml`'s `{{ else }} [] {{ end }}` branch (bad indentation under a block-sequence context breaks YAML parsing) -- unrelated to the `required`-value guards, but it blocks `helm template` all the same. Left the template itself untouched since fixing that bug is out of scope for this ticket; the dummy value routes around it.
- Extended `helm-lint.yaml`'s matrix from a flat `chart: [...]` list to a `component`+`chart` `include` list, and added `fulfillment-service`/`service` with `values-file: ci-values.yaml` passed via `-f` (only for that entry). Also added `fulfillment-service/charts/**` to the workflow's path trigger.

Part of OSAC-1732 (mono-repo consolidation) / OSAC-1733.

## Test plan
- [x] `helm lint charts/service/ -f charts/service/ci-values.yaml` passes (both default `variant: kind` and `--set variant=openshift`)
- [x] `helm template test charts/service/ -f charts/service/ci-values.yaml` renders successfully for both variants
- [x] Re-ran the two existing `osac-operator` matrix entries (`operator-crds`, `operator`) locally with the updated workflow's exact commands -- still pass unchanged
- [x] `python3 -c "import yaml; yaml.safe_load(...)"` on both touched files
- [x] `pre-commit run --files fulfillment-service/charts/service/ci-values.yaml .github/workflows/helm-lint.yaml` passes

### DIFF
--- a/.github/workflows/helm-lint.yaml
+++ b/.github/workflows/helm-lint.yaml
@@ -8,6 +8,7 @@ on:
       - 'osac-operator/config/crd/bases/**'
       - 'osac-operator/hack/sync-helm-crds.py'
       - 'osac-operator/Makefile'
+      - 'fulfillment-service/charts/**'
       - '.github/workflows/helm-lint.yaml'
 
 permissions:
@@ -35,17 +36,24 @@ jobs:
       working-directory: osac-operator
       run: make check-helm-crds
 
-  # fulfillment-service's chart isn't included here: it guards most values
-  # with `required "..." .Values.X` (auth/certs/idp config), so it can't be
-  # helm-lint'd with bare values.yaml the way osac-operator's charts can.
-  # Needs a CI-only dummy-values file before it can join this matrix.
+  # fulfillment-service's chart guards most values with `required "..." .Values.X`
+  # (auth/certs/idp config), so it needs its CI-only charts/service/ci-values.yaml
+  # passed via -f to lint/template cleanly -- osac-operator's charts have no such
+  # guards and lint fine bare.
   helm-lint:
-    name: Lint Helm chart (osac-operator/${{ matrix.chart }})
+    name: Lint Helm chart (${{ matrix.component }}/${{ matrix.chart }})
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        chart: [operator-crds, operator]
+        include:
+          - component: osac-operator
+            chart: operator-crds
+          - component: osac-operator
+            chart: operator
+          - component: fulfillment-service
+            chart: service
+            values-file: ci-values.yaml
     steps:
     - name: Checkout repository
       uses: actions/checkout@v7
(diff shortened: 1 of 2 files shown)


eliorerz:
Verified every claim in this PR against the actual diff, chart templates, and osac-installer's real usage — not just trusting the description or the green CI. Everything checks out, and I filed a follow-up ticket for the one thing that's genuinely out of this PR's scope but worth tracking.

**The claimed template bug is real — reproduced it directly.**

`grpc-server/deployment.yaml`'s database-connection volume source has its `{{ if }}/{{ range }}/{{ else }}[]{{ end }}` block indented at the *same* level as `sources:` rather than one level deeper. I confirmed this renders invalid YAML when `database.connection` is empty (the chart's own documented default):

```
$ helm template test fulfillment-service/charts/service -f <ci-values.yaml with database.connection stripped back to the chart default []>
Error: YAML parse error on fulfillment-service/templates/grpc-server/deployment.yaml: error converting YAML to JSON: yaml: line 40: could not find expected ':'
```

And confirmed the fix in this PR's `ci-values.yaml` (setting a non-empty `database.connection`) does route around it — `helm template` succeeds cleanly with the file as submitted.

**Is the workaround safe, or does it mask a real deployment risk?** Checked this carefully since it matters beyond CI:

- Every values file currently in actual use in `osac-installer` (`development`, `bmaas-ci`, `vmaas-ci`, `caas-ci`, `charts/osac/ci/*.yaml`, `values-example.yaml`) already sets `service.database.connection` to a real secret reference — so no currently-deployed environment hits this bug today.
- But `osac-installer/charts/osac/values.yaml`'s own top-level default *also* leaves `database.connection: []`, with just a commented-out example. Unlike the chart's other guarded fields (`auth.*`, `certs.*`, `idp.*`, which all use `required "..."` and fail with a clear message), `database.connection` has no such guard — so a future values file, or a bare `helm install` against defaults, would hit this same opaque YAML parse error instead of a clear validation failure.

That's a real, if currently dormant, bug — filed as **[OSAC-3495](https://redhat.atlassian.net/browse/OSAC-3495)** with the repro, impact analysis, and two suggested fixes (indentation fix, or add a `required` guard matching the chart's existing convention). Agree this PR was right to route around it rather than fix it here — routing around it for CI purposes doesn't mask anything, since it doesn't touch the template and every real deployment path already avoids the broken branch independently.

**Everything else checks out:**
- `ci-values.yaml` sets exactly the 8 fields covered by `required "..."` guards in this chart (`certs.issuerRef.name`, `certs.caBundle.configMap`, `externalHostname`, `internalHostname`, `auth.issuerUrl`, `auth.controllerCredentials`, `idp.url`, `idp.credentials`) plus the one documented `database.connection` workaround — nothing extra, nothing missing. Cross-checked against the chart's actual `required` call sites and values.yaml defaults (e.g. confirmed `certs.issuerRef.kind` correctly wasn't added since it has a real default, `ClusterIssuer`).
- `helm-lint.yaml`'s matrix restructuring is correct and doesn't disturb the existing `osac-operator` entries — confirmed both via reading the conditional `-f` logic and by checking the actual CI run: all three matrix legs (`fulfillment-service/service`, `osac-operator/operator`, `osac-operator/operator-crds`) pass.
- Re: comparing to `osac-operator`'s approach — `osac-operator` doesn't need an equivalent CI-values file because its one `required` guard (`certs.issuerRef.name` in `console-proxy/certificate.yaml`) is satisfied by a real default (`default-ca`) in its own `values.yaml`. Nothing inconsistent there; correctly not duplicated.

Nice catch on the template bug, and the right call keeping the fix out of this PR's scope.

PR #29, opened 2026-07-31
OSAC-3467: Scope fulfillment-service/osac-operator release-trigger tags

eliorerz:
## Summary

Fixes the tag-collision gap identified in OSAC-3467's comments after the OSAC-1733 merge: fulfillment-service's 4 tag-triggered workflows and osac-operator's 2 still trigger on a bare `v*`, so a single tag push right now would fire every one of them at once.

Convention adopted (already precedented by osac-operator's own `api/vX.Y.Z` submodule tags): `<component>/vX.Y.Z`, e.g. `fulfillment-service/v1.2.3`, `osac-operator/v1.2.3`.

**Trigger scoping** (`on.push.tags`):
- `publish-image.yaml`, `publish-binaries.yaml`, `publish-proto.yaml` -> `fulfillment-service/v*`
- `build-image.yaml` -> `osac-operator/v*`

**Downstream version handling** -- scoping the trigger alone isn't sufficient: several steps assumed a bare `vX.Y.Z` tag and would have silently broken (invalid Docker tags containing `/`, goreleaser/semver parsing failing on the prefix, `publish-charts.yaml`'s guard regex rejecting anything not starting with `v`). Updated all of these to strip the known prefix before using the value as a version:
- `publish-image.yaml`/`build-image.yaml`: replaced `docker/metadata-action`'s `type=semver,pattern=...` lines (which match the raw ref and would no longer match a prefixed tag) with a preceding step that strips the prefix and emits `RELEASE_VERSION`/`RELEASE_MAJOR_MINOR`/`RELEASE_MAJOR`, fed in via `type=raw,value=...,enable=...`.
- `publish-binaries.yaml`: strips the prefix and passes the bare version to goreleaser explicitly via `--current-tag` (goreleaser's own tag auto-detection would otherwise try to parse the full scoped tag as semver and fail).
- `publish-proto.yaml`: strips the prefix before using it as the buf push label (cosmetic, buf labels have no semver requirement, but keeps it clean).
- `publish-charts.yaml` (triggered via `workflow_run` from the two image-build workflows, not directly by the tag push): the shared `guard` job now recognizes either component's prefix, strips it to a `version` output (bare `vX.Y.Z`), and validates *that* against the semver regex. All three chart-publishing jobs and `create-operator-release` now derive chart/app versions from `needs.guard.outputs.version` instead of mis-stripping a literal `v` off the front of the full scoped tag (which was a no-op on a tag starting with a component name, not `v`). The full scoped tag is still used wherever the actual git ref matters (`gh release create`, the force-push/retag guard script).

**Stale tags from the subtree merge** -- verified directly against the live repo (`git tag -l` on this clone): **zero tags exist**. This matches what's already recorded in OSAC-3467's comments (subtree-add's internal fetch did not carry tag refs into either the fulfillment-service or osac-operator merge). Nothing to strip for this PR.

## Not included (flagging, not fixing here)
OSAC-3467's third sub-item -- dropping now-unnecessary go.mod version-pin tags now that `go.work` resolves cross-component imports locally -- is a real, separate finding: both `fulfillment-service/go.mod` and `osac-operator/go.mod` still `require github.com/osac-project/osac-operator/api v0.0.7` even though `go.work`'s `use` directive already makes that resolve to the local `./osac-operator/api` in this repo. Left untouched here since it's a Go-module-correctness change (needs `go mod tidy`/build verification, not YAML) rather than CI tag-trigger scoping -- recommend a separate follow-up.

## Validation
- [x] `yamllint -c .yamllint.yaml --strict` on all 5 changed files -- clean.
- [x] `python3 -c "yaml.safe_load(...)"` on all 5 files -- parses.
- [x] Full diff reviewed end-to-end for internal consistency (which value feeds which downstream step).
- [ ] Not able to exercise this live end-to-end (no tag has been pushed to this repo yet, and I didn't want to push a real component-scoped tag myself since that would trigger real image builds/chart publishes/GitHub releases). Recommend a real dry run with a test tag before relying on this for an actual release -- flagging the specific third-party-tool assumptions this makes: `docker/metadata-action`'s `type=raw`/`enable=` combination (well-established, low risk) and goreleaser's `--current-tag` flag overriding its own git-tag auto-detection (a real, documented goreleaser flag, but not exercised here).

## Summary by CodeRabbit

* **Release Automation**
  * Release workflows now support component-specific version tags for service and operator releases.
  * Docker images, binaries, charts, and protocol packages are published with clean, consistent version labels.
  * Chart releases validate version formats and use the extracted component version for packaging and release titles.
  * Builds and publishing are restricted to the appropriate component tags, reducing accidental cross-component releases.

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -13,7 +13,7 @@ on:
     branches:
     - main
     tags:
-    - 'v*'
+    - 'osac-operator/v*'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
@@ -80,6 +80,17 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      # Component-scoped tags (see the `tags:` trigger above) can't be used verbatim
+      # as image tags -- Docker/OCI tags can't contain the '/' the git tag does --
+      # so strip the prefix here to recover the bare vX.Y.Z for the actual image tags.
+      - name: Compute release version
+        if: startsWith(github.ref, 'refs/tags/osac-operator/')
+        run: |
+          version="${GITHUB_REF_NAME#osac-operator/}"
+          echo "RELEASE_VERSION=${version}" >> "$GITHUB_ENV"
+          echo "RELEASE_MAJOR_MINOR=${version%.*}" >> "$GITHUB_ENV"
+          echo "RELEASE_MAJOR=${version%%.*}" >> "$GITHUB_ENV"
+
       # Extract metadata (tags, labels) for Docker
       # https://github.com/docker/metadata-action
       - name: Extract Docker metadata
@@ -88,9 +99,9 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: |
-            type=semver,pattern=v{{version}}
-            type=semver,pattern=v{{major}}.{{minor}}
-            type=semver,pattern=v{{major}}
+            type=raw,value=${{ env.RELEASE_VERSION }},enable=${{ env.RELEASE_VERSION != '' }}
+            type=raw,value=${{ env.RELEASE_MAJOR_MINOR }},enable=${{ env.RELEASE_VERSION != '' }}
+            type=raw,value=${{ env.RELEASE_MAJOR }},enable=${{ env.RELEASE_VERSION != '' }}
             type=ref,event=branch
             type=ref,event=pr
             type=ref,event=tag
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -98,13 +98,17 @@ jobs:
         uses: docker/metadata-action@v6
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          # No `type=ref,event=tag` here: the `tags:` trigger above only ever
+          # matches this component's own scoped prefix, so the tag-triggered
+          # case is always covered by the three raw entries below already --
+          # keeping it would just add a redundant, slash-sanitized duplicate
+          # tag (e.g. "osac-operator-v1.2.3").
           tags: |
             type=raw,value=${{ env.RELEASE_VERSION }},enable=${{ env.RELEASE_VERSION != '' }}
             type=raw,value=${{ env.RELEASE_MAJOR_MINOR }},enable=${{ env.RELEASE_VERSION != '' }}
             type=raw,value=${{ env.RELEASE_MAJOR }},enable=${{ env.RELEASE_VERSION != '' }}
             type=ref,event=branch
             type=ref,event=pr
-            type=ref,event=tag
             type=sha
             type=raw,value=latest,enable=${{ github.ref == format('refs/heads/{0}', 'main') }}
 
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -87,9 +87,21 @@ jobs:
         if: startsWith(github.ref, 'refs/tags/osac-operator/')
         run: |
           version="${GITHUB_REF_NAME#osac-operator/}"
+          # Same SemVer rule as publish-charts.yaml's guard job -- rejects build
+          # metadata (+...) and validates the core/prerelease shape before it's
+          # trusted as an image tag.
+          semver_re='^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-((0|[1-9][0-9]*)|[0-9A-Za-z-]*[A-Za-z-][0-9A-Za-z-]*)(\.((0|[1-9][0-9]*)|[0-9A-Za-z-]*[A-Za-z-][0-9A-Za-z-]*))*)?$'
+          if ! [[ "${version}" =~ ${semver_re} ]]; then
+            echo "::error::Tag '${GITHUB_REF_NAME}' is not a valid semver release tag"
+            exit 1
+          fi
+          # Prerelease identifiers can themselves contain dots (e.g. v1.2.3-beta.1),
+          # so derive major/minor from the prerelease-stripped core version --
+          # otherwise a dotted prerelease would corrupt the major.minor alias.
+          core="${version%%-*}"
           echo "RELEASE_VERSION=${version}" >> "$GITHUB_ENV"
-          echo "RELEASE_MAJOR_MINOR=${version%.*}" >> "$GITHUB_ENV"
-          echo "RELEASE_MAJOR=${version%%.*}" >> "$GITHUB_ENV"
+          echo "RELEASE_MAJOR_MINOR=${core%.*}" >> "$GITHUB_ENV"
+          echo "RELEASE_MAJOR=${core%%.*}" >> "$GITHUB_ENV"
 
       # Extract metadata (tags, labels) for Docker
       # https://github.com/docker/metadata-action
--- a/.github/workflows/publish-binaries.yaml
+++ b/.github/workflows/publish-binaries.yaml
@@ -41,6 +41,7 @@ jobs:
     - uses: goreleaser/goreleaser-action@f06c13b6b1a9625abc9e6e439d9c05a8f2190e94  # v7
       with:
         workdir: fulfillment-service
-        args: release --clean --current-tag ${{ env.RELEASE_VERSION }}
+        args: release --clean
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GORELEASER_CURRENT_TAG: ${{ env.RELEASE_VERSION }}
--- a/.github/workflows/publish-binaries.yaml
+++ b/.github/workflows/publish-binaries.yaml
@@ -17,7 +17,7 @@ name: Publish binaries
 on:
   push:
     tags:
-    - 'v*'
+    - 'fulfillment-service/v*'
 
 jobs:
 
@@ -34,9 +34,13 @@ jobs:
     - uses: ./.github/actions/setup-go
       with:
         working-directory: fulfillment-service
+    - name: Compute release version
+      # Strip this component's tag prefix (see the `tags:` trigger above) --
+      # goreleaser expects a bare vX.Y.Z, not the scoped git tag.
+      run: echo "RELEASE_VERSION=${GITHUB_REF_NAME#fulfillment-service/}" >> "$GITHUB_ENV"
     - uses: goreleaser/goreleaser-action@f06c13b6b1a9625abc9e6e439d9c05a8f2190e94  # v7
       with:
         workdir: fulfillment-service
-        args: release --clean
+        args: release --clean --current-tag ${{ env.RELEASE_VERSION }}
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/publish-charts.yaml
+++ b/.github/workflows/publish-charts.yaml
@@ -46,20 +46,32 @@ jobs:
     permissions: {}
     if: >
       github.event.workflow_run.event == 'push' &&
-      startsWith(github.event.workflow_run.head_branch, 'v')
+      (startsWith(github.event.workflow_run.head_branch, 'fulfillment-service/v') ||
+       startsWith(github.event.workflow_run.head_branch, 'osac-operator/v'))
     outputs:
       tag: ${{ github.event.workflow_run.head_branch }}
+      version: ${{ steps.check.outputs.version }}
       sha: ${{ github.event.workflow_run.head_sha }}
     steps:
     - name: Check image build result
+      id: check
       env:
         CONCLUSION: ${{ github.event.workflow_run.conclusion }}
         HEAD_BRANCH: ${{ github.event.workflow_run.head_branch }}
       run: |
-        # No build-metadata (+...) suffix allowed: this tag is used verbatim as a
+        # Each component owns a scoped tag prefix (see publish-image.yaml/build-image.yaml's
+        # own `tags:` triggers) so a single tag push can't fire every component's release
+        # chain at once; strip it here to recover the bare vX.Y.Z used for artifact versions.
+        case "$HEAD_BRANCH" in
+          fulfillment-service/v*) VERSION="${HEAD_BRANCH#fulfillment-service/}" ;;
+          osac-operator/v*) VERSION="${HEAD_BRANCH#osac-operator/}" ;;
+          *) echo "::error::Tag '$HEAD_BRANCH' does not match a known component prefix"; exit 1 ;;
+        esac
+
+        # No build-metadata (+...) suffix allowed: this version is used verbatim as a
         # container image tag downstream, and Docker/OCI tags cannot contain '+'.
         semver_re='^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-((0|[1-9][0-9]*)|[0-9A-Za-z-]*[A-Za-z-][0-9A-Za-z-]*)(\.((0|[1-9][0-9]*)|[0-9A-Za-z-]*[A-Za-z-][0-9A-Za-z-]*))*)?$'
-        if ! [[ "$HEAD_BRANCH" =~ $semver_re ]]; then
+        if ! [[ "$VERSION" =~ $semver_re ]]; then
           echo "::error::Tag '$HEAD_BRANCH' is not a valid semver release tag"
           exit 1
         fi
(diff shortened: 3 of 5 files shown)


eliorerz:
Read the full diff and traced every downstream use of the stripped version through each file. Found one critical bug that CI can't catch (goreleaser never runs during PR checks, only on a real tag push), plus one minor cosmetic side-effect. Everything else checks out.

**Blocker: `publish-binaries.yaml`'s `--current-tag` is not a valid goreleaser flag.**

```yaml
- uses: goreleaser/goreleaser-action@f06c13b6...  # v7
  with:
    workdir: fulfillment-service
    args: release --clean --current-tag ${{ env.RELEASE_VERSION }}
```

I cloned `goreleaser/goreleaser` and read the actual `release` subcommand's flag registration (`cmd/release.go`) on current main (which matches this repo's `.goreleaser.yaml`'s `version: 2` schema, and `goreleaser-action`'s own default `version: '~> v2'`). The full flag list is: `config, release-notes(+tmpl), release-header(+tmpl), release-footer(+tmpl), auto-snapshot, snapshot, draft, fail-fast, clean, parallelism, timeout, deprecated, skip`. There is no `current-tag` flag anywhere in the CLI — I grepped the entire repo (source + docs) for "current-tag" and got zero hits outside of internal Go identifiers.

What *does* exist, and is what should be used here: `GORELEASER_CURRENT_TAG`, an **environment variable** (confirmed in `internal/pipe/git/git.go:269` — it's checked first, before git-based tag auto-detection, in goreleaser's own tag-resolution logic). Passing `--current-tag` as a CLI arg instead will make cobra reject it as an unknown flag and `goreleaser release` will exit non-zero immediately, before doing anything.

Net effect: this workflow only triggers on `push: tags: fulfillment-service/v*`, which never happens during PR checks — so this is fully invisible to CI (all green, as reported) and will only surface the first time someone actually pushes a real `fulfillment-service/vX.Y.Z` tag, at which point `publish-binaries.yaml` fails outright and no binaries/GitHub Release get published for fulfillment-service.

Fix: move it to the step's `env:` block instead of `args:`:
```yaml
- uses: goreleaser/goreleaser-action@...
  with:
    workdir: fulfillment-service
    args: release --clean
  env:
    GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
    GORELEASER_CURRENT_TAG: ${{ env.RELEASE_VERSION }}
```

**Minor/FYI, not blocking:** `build-image.yaml`/`publish-image.yaml` leave the pre-existing `type=ref,event=tag` docker/metadata-action rule untouched. Since the git ref is now e.g. `osac-operator/v1.2.3` (contains a slash), that rule will sanitize it (slash → dash per docker/metadata-action's documented normalization) and publish an extra, slightly redundant image tag like `osac-operator-v1.2.3` alongside the correctly-computed `v1.2.3`/`v1.2`/`v1` tags from the new `type=raw` lines. Not a functional break, just a cosmetic extra tag in the registry — up to you whether it's worth suppressing with `enable=false` on that line.

**Everything else verified correct:**
- `publish-charts.yaml`'s `guard` job case statement correctly maps both `fulfillment-service/v*`/`osac-operator/v*` prefixes and hard-errors (`exit 1`, not a silent pass-through) on anything else — confirmed this is unreachable in practice today only because the job's own top-level `if:` already gates on the same two prefixes, but it's a correct belt-and-suspenders check regardless.
- Every downstream consumer of `needs.guard.outputs.version` (chart version/app version in `publish-service-chart`, the `version` step in both operator chart jobs, the release title in `create-operator-release`) correctly reconstructs the same bare `vX.Y.Z` (and `X.Y.Z` where `v` is stripped for Helm's semver requirement) that the old unscoped-tag scheme produced — traced each one, no regressions.
- `osac-operator-verify-tag-matches-sha.sh` (untouched, correctly so) works fine with a slash-containing scoped tag — GitHub's `git/ref/tags/{ref}` API natively supports multi-segment ref names, verified by reading the script.
- Confirmed zero pre-existing tags on this repo (`git tag -l` / `git ls-remote --tags` both empty) — nothing to migrate, matches the PR's own claim.
- No documentation anywhere instructs maintainers to push a bare `vX.Y.Z` tag to cut a release, so there's no stale-docs cutover risk either.
- The diff's structure (case statement + `if:` prefix check) is cleanly additive for a future `osac-aap/v*` arm — no restructuring needed when that lands.
- The go.mod version-pin follow-up is accurately and explicitly documented as out-of-scope in the PR description, not silently dropped.

Good catch on identifying the real collision problem and the overall approach — just the one goreleaser flag needs fixing before this is safe to merge.

eliorerz:
Good catch, and good process for finding it -- confirmed against goreleaser's own flag registration, that's exactly right. `--current-tag` was never a real CLI flag; I conflated it with the actual `GORELEASER_CURRENT_TAG` env var. Fixed in the latest commit: moved it to `env:` on the goreleaser-action step, `args:` is back to plain `release --clean`.

Also took the non-blocking FYI: removed the now-fully-redundant `type=ref,event=tag` line from both publish-image.yaml and build-image.yaml. Since each workflow's `tags:` trigger only ever matches that component's own scoped prefix now, every tag-triggered run is already covered by the three `type=raw` entries -- the ref-based rule could only ever ADD a slash-sanitized duplicate (e.g. `osac-operator-v1.2.3`) alongside the clean version tags, never anything meaningfully different. Left a comment in both files explaining why it's gone, so it doesn't look like an accidental drop.

Validated both changes with a YAML parse check and `yamllint -c .yamllint.yaml --strict` on all 3 touched files -- clean.

eliorerz:
## Independent second review

Did this cold — pulled the fresh diff, formed my own conclusions first, then read the existing thread. Focused specifically on what only executes on a real tag push (same blind spot as the `--current-tag` bug already found and fixed).

### 🔴 New blocker: `GORELEASER_CURRENT_TAG` fix is directionally correct but still doesn't work

The fix applied for the earlier finding (moving `--current-tag` from `args:` to `env: GORELEASER_CURRENT_TAG: ${{ env.RELEASE_VERSION }}`) uses a real, documented goreleaser mechanism — but `goreleaser release` will still fail on a real `fulfillment-service/vX.Y.Z` tag push, for a different reason.

**Root cause**, traced through goreleaser v2's actual source (`internal/pipe/git/git.go`, current `main`):

- `(Pipe) Run()` unconditionally calls `validate(ctx)` at the end of every `release` invocation (line 59) — this isn't optional or skippable except via an explicit `--skip=validate` flag, which isn't set anywhere in this PR.
- `validate()` runs: `git describe --exact-match --tags --match ctx.Git.CurrentTag` (line 207).
- `ctx.Git.CurrentTag` comes from `getTag()`, which reads `GORELEASER_CURRENT_TAG` first via a plain `os.Getenv()` — **no existence check at that stage** (line 269, `getFromEnv`). So `CurrentTag` becomes the bare `v1.2.3` from `RELEASE_VERSION`.
- But the actual git tag that exists in the checkout is `fulfillment-service/v1.2.3` — nothing named literally `v1.2.3` exists. `git describe --match` does a glob match against tag names, not a substring/suffix match, so it will not find it.

I reproduced this empirically rather than trusting the source read alone:

```
$ git init -q && git commit -q --allow-empty -m x && git tag "fulfillment-service/v1.2.3"
$ git describe --exact-match --tags --match "v1.2.3"
fatal: No names found, cannot describe anything.
$ echo $?
128
```

That's the exact command `validate()` runs, with the exact values this workflow will produce, reproducing the exact failure. `validate()` turns that into `ErrWrongRef`, which `Run()` returns directly, so `goreleaser release --clean` exits non-zero — `publish-binaries.yaml` fails on every real release, no binaries or GitHub Release get published, and (same as the original finding) this is entirely invisible in CI/PR checks since it only executes on a tag push.

I also checked whether `goreleaser-action` itself does anything special with `GORELEASER_CURRENT_TAG` (e.g. auto-adding `--skip=validate`) — it doesn't; `main.ts` just does `exec.exec(\`${bin} ${inputs.args}\`)` verbatim, no interception.

**Suggested fix**: `validate()`'s tag-existence check is purely local (`git describe` against the checkout, not a remote/API call), so creating an unpushed local lightweight tag matching the bare version before invoking goreleaser satisfies it without touching the real tag:
```yaml
- name: Compute release version
  run: |
    version="${GITHUB_REF_NAME#fulfillment-service/}"
    echo "RELEASE_VERSION=${version}" >> "$GITHUB_ENV"
    git tag "${version}"   # local-only alias so goreleaser's git-describe validate step finds it
- uses: goreleaser/goreleaser-action@...
  with:
    workdir: fulfillment-service
    args: release --clean
  env:
    GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
    GORELEASER_CURRENT_TAG: ${{ env.RELEASE_VERSION }}
```
(`GORELEASER_CURRENT_TAG` becomes redundant once the local tag exists — goreleaser's own auto-detection via `git tag --points-at HEAD` would find it — but keeping it is harmless and more explicit.) The blunter alternative, `args: release --clean --skip=validate`, also works but additionally disables the dirty-worktree check, which is worth keeping.

### 🟡 Confirming CodeRabbit's finding, with one correction

CodeRabbit's point about `build-image.yaml`/`publish-image.yaml`'s bash-based `${version%.*}`/`${version%%.*}` mishandling prerelease/build-metadata suffixes is real — verified locally:
```
version="v1.2.3-rc.1"  →  RELEASE_MAJOR_MINOR="v1.2.3-rc"   (should be v1.2)
version="v1.2.3+build.1" → RELEASE_MAJOR_MINOR="v1.2.3+build" (should be v1.2)
```
One correction to the framing: I don't think this "fails image publication" as stated. `docker/metadata-action`'s own docs say invalid tag characters get sanitized (`+` → `-`), not rejected — so the likely effect is a **silently wrong** major/minor alias tag gets published (e.g. `v1.2.3-rc` instead of `v1.2`), not a hard failure. Build-metadata (`+...`) tags would additionally get rejected later by `publish-charts.yaml`'s guard job's semver regex (which has no `+build` group) — but only *after* `build-image.yaml`/`publish-image.yaml` have already pushed a mis-tagged image, since the guard runs in a separate, later `workflow_run`-triggered job. Real bug, correctly flagged, but lower severity than the goreleaser one above — this project's tags all appear to be plain `vX.Y.Z` in practice, and it doesn't block a normal release.

### ✅ Everything else I checked — verified clean, with evidence

Went through every external tool/action touch point the same way as the two issues above, plus the slash-interaction angle specifically:

- **`docker/metadata-action` `type=raw`/`enable=` syntax**: confirmed against the action's own README — `enable=<true|false>` genuinely skips the whole entry when false, matching the intent of `enable=${{ env.RELEASE_VERSION != '' }}` on PR/branch runs (where the compute-version step didn't run and `env.RELEASE_VERSION` correctly resolves to empty string, not an error).
- **Removing `type=ref,event=tag`**: confirmed the stated reasoning against the docs — its sanitization rule (`Replace invalid char sequences with '-' for tags`) would otherwise turn `osac-operator/v1.2.3` into a redundant `osac-operator-v1.2.3` tag; removing it is correct, not just cosmetic-neutral.
- **`buf push --label`**: no documented character/format restrictions, so stripping the prefix first (as done) is a style choice, not a correctness requirement, but no bug either way.
- **GitHub Releases + slash-containing tags** (the specific "does a tool assume no slashes" risk): checked `gh release create ... --verify-tag` and the `gh release edit` fallback in `create-operator-release` against `gh` CLI's actual source. `--verify-tag`'s existence check uses **GraphQL** (`ref(qualifiedName: "refs/tags/osac-operator/v1.2.3")` as a variable, not a URL path segment) — no slash-in-path risk at all. The release-by-tag REST lookup (`gh release edit`/`view`) builds its URL via `cli/cli`'s `safeurl` package, which percent-encodes each path component independently (`url.PathEscape`), so the tag's internal `/` becomes `%2F`. I confirmed empirically that GitHub's API decodes `%2F` back to a literal `/` for routing purposes (tested both a literal-slash and a `%2F`-encoded request against the same real ref on `osac-operator` — identical results). Both paths are safe.
- **`.github/scripts/osac-operator-verify-tag-matches-sha.sh`** (unchanged by this PR, but now consumes a slash-containing `TAG`): uses `gh api repos/${REPO}/git/ref/tags/${TAG}` — GitHub's `git/ref/{ref}` endpoint natively takes a multi-segment ref. Verified empirically against a real existing tag (`osac-operator` already has `api/vX.Y.Z` submodule tags) — resolves correctly.
- **Tag glob semantics**: confirmed via GitHub's own docs that a bare `*` in `tags:` patterns cannot cross `/` (would need `**`) — so `'osac-operator/v*'` correctly matches only this component's tags, and conversely a stray `'v*'` elsewhere would *not* accidentally start matching these scoped tags.
- **Other workflows in this repo that might still assume bare `v*` tags**: checked every workflow file in `.github/workflows/` — the four this PR touches are the only ones with a `tags:` trigger; nothing was missed or left orphaned. `publish-openapi.yaml` looks like a sibling of publish-binaries/image/proto but is intentionally branch-only (self-documented in its own header comment, unrelated to this PR).
- **Dependabot / branch protection**: no tag-pattern dependencies in `.github/dependabot.yml`, and no branch protection rulesets configured on `main` that reference tags.
- **Kubernetes label values / chart version strings**: every place a Helm chart version, app-version, or image tag gets set uses the already-stripped bare `VERSION`, never the raw scoped `TAG` — no k8s-label-illegal characters flow through anywhere.

### Bottom line

Not yet safe to merge — the goreleaser `validate()` issue above is a real, reproduced blocker on `publish-binaries.yaml` that would only surface on the first real `fulfillment-service/vX.Y.Z` tag push, same failure class (and same root cause: an untested-in-CI code path) as the flag bug already fixed. CodeRabbit's build-metadata/prerelease finding is also real but lower priority. Everything else — including every place I could find where the new slash-containing tags interact with an external tool or API — checks out.

eliorerz:
Reviewed commit ee72cf68 (CodeRabbit's SemVer-validation finding) with the same rigor as any worker PR — pulled the branch fresh, read the full current content of both changed steps (not the commit message), and hand-tested independently.

**(1) Read the actual current content.** `build-image.yaml`'s and `publish-image.yaml`'s `Compute release version` steps now both: validate with a SemVer regex before trusting the value, then compute `core="${version%%-*}"` (strip from the first hyphen onward) and derive `RELEASE_MAJOR_MINOR`/`RELEASE_MAJOR` from `core` instead of the raw `version`. `RELEASE_VERSION` itself is unchanged — still the full tag-derived version (correct: the exact-version Docker tag alias should keep any prerelease qualifier; only the rolling major/major.minor aliases need the prerelease stripped).

**(2) Regex consistency — confirmed byte-for-byte identical, programmatically, not by eye:**
```
$ diff <(grep -o "semver_re='[^']*'" build-image.yaml) <(grep -o "semver_re='[^']*'" publish-charts.yaml)
(no output — identical)
```
Same check against `publish-image.yaml` — also identical. All three files use the exact same pattern, character for character. No drift between the workflows that would let one accept a tag another rejects.

**(3) Hand-tested the regex + derivation against a fresh bash session (not trusting the PR's own claimed tests):**

| Input | Result | RELEASE_VERSION | MAJOR_MINOR | MAJOR |
|---|---|---|---|---|
| `v1.2.3` | accepted | `v1.2.3` | `v1.2` | `v1` |
| `v1.2.3-beta.1` (dotted prerelease) | accepted | `v1.2.3-beta.1` | `v1.2` ✅ | `v1` |
| `v1.2.3-rc1` (non-dotted prerelease) | accepted | `v1.2.3-rc1` | `v1.2` | `v1` |
| `v1.2.3+build5` (build metadata) | **rejected** | — | — | — |
| `v1.2`, `1.2.3`, `vfoo` (malformed) | **rejected** | — | — | — |

The dotted-prerelease case is the one that matters: before this fix, `${version%.*}` on `v1.2.3-beta.1` would strip only the last dot and produce `v1.2.3-beta` as the "major.minor" alias — silently wrong. With the fix, stripping from the first hyphen first (`core=v1.2.3`) before taking `${core%.*}` correctly yields `v1.2`. Build metadata is correctly rejected outright since the regex has no `+...` branch at all, so anything with a `+` suffix fails the full-string anchor match.

**(4) Confirmed nothing else changed.** `git show ee72cf68` — the diff is exactly 16 insertions / 4 deletions per file, entirely inside the `Compute release version` step's `run:` block. Nothing else in either file (triggers, docker/metadata-action config, the `RELEASE_VERSION` output itself, build/push steps) was touched.

Also independently re-validated YAML parses cleanly on all three files post-change.

Verdict: **correct, matches publish-charts.yaml's existing rule exactly, handles all the required cases correctly, and is scoped to exactly the lines that needed to change.** No new issues found. (Side note, not part of this ask: I can see `bfb78056` on this branch already addresses the goreleaser `--current-tag` flag issue I flagged earlier on this same PR — that's fixed too.)

eliorerz:
Disposition of CodeRabbit's latest review (commit ee72cf68):

- **SemVer validation finding (original review)**: fixed directly on this branch (commit ee72cf68) and independently re-verified — no longer flagged.
- **Cosign/Sigstore image signing** and **missing SAST/SCA gate before image push**: verified against `main` — both gaps predate this PR entirely (neither file's signing/scanning surface is touched by this diff, and neither exists on `main` today). Real findings, but out of scope for OSAC-3467 (tag scoping). Tracking separately rather than expanding this PR's scope.

No further changes needed on this PR for these findings.

PR #33, opened 2026-07-31
OSAC-3227: Phase A verification -- ci-poc path touched, FAIL_ME sentinel present

eliorerz:
## Purpose

Throwaway test PR for OSAC-3227 Phase A: verifies the `poc-path-skip-check` reference workflow (merged in #20) when its filtered path IS touched.

**Expected:** the `dorny/paths-filter` hits, the inner `check` job runs and fails (because `.github/ci-poc/FAIL_ME` is present), and the `poc-path-skip-check` aggregator concludes `failure`.

A follow-up commit will remove `FAIL_ME` to verify the same jobs then conclude `success`.

**This PR will be closed without merging once evidence is captured.**

### DIFF
--- a/.github/ci-poc/FAIL_ME
+++ b/.github/ci-poc/FAIL_ME
@@ -1,0 +1,2 @@
+This sentinel file's mere presence makes the PoC workflow's inner
+check job fail on purpose. Removing it lets the check pass.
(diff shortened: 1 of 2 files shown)


eliorerz:
Closing without merging — this PR did its job: it caught a real bug in the merged reference workflow (the 'check' job never checks out the repo, so the FAIL_ME sentinel test always passes regardless of the file's actual presence — confirmed via gh api job-step inspection on this PR's run).

Fix is in #34. Once that merges, I'll open a fresh Phase A test PR against the corrected workflow.

PR #44, opened 2026-07-31
OSAC-3511: Wire osac-aap into the tag-scoped release pipeline

eliorerz:
## Summary
Wires osac-aap into the now-proven [OSAC-3467](https://redhat.atlassian.net/browse/OSAC-3467)/OSAC-3490 tag-scoped release pattern already verified end-to-end on fulfillment-service and osac-operator. This was deliberately deferred, not forgotten -- per `publish-charts.yaml`'s own prior comment, wiring osac-aap in before PR #29 ([OSAC-3467](https://redhat.atlassian.net/browse/OSAC-3467)) settled its final shape would have needed rework regardless of merge order. PR #29 is merged and both other components have real tags fully verified through image build -> GHCR push -> chart publish -> GitHub Release.

## Changes

**`execution-environment.yml`**: trigger changed from unscoped `tags: - 'v*'` to `osac-aap/v*`, matching `build-image.yaml`/`publish-image.yaml`'s established convention. Added the same tag-prefix-stripping + SemVer-validation step (including the prerelease-safe major/minor derivation) and swapped `docker/metadata-action`'s `type=semver,pattern=...` entries for the same `type=raw,value=...,enable=...` pattern those two files use, for the same reason: a scoped tag with a literal `/` in it can't be matched by `type=semver`'s anchored pattern. Also dropped the now-redundant `type=ref,event=tag` entry, matching the reasoning already applied to the other two workflows.

**`publish-charts.yaml`**: added `osac-aap/v` to the shared `guard` job's `startsWith` condition and its tag-prefix-stripping `case` statement. Added a new `publish-osac-aap-chart` job, ported from the old standalone `osac-project/osac-aap` repo's own `publish-aap-chart` job (read in full, not guessed), adapted to this file's current conventions:
- `if: success() && github.event.workflow_run.name == 'Build execution environment'` (the actual name of this repo's build workflow), matching how the fulfillment-service/osac-operator jobs gate on their own workflow names.
- Sources the version from `needs.guard.outputs.version` (already prefix-stripped) instead of re-deriving it from the raw tag.
- Uses `.github/scripts/osac-aap-verify-tag-matches-sha.sh` (the script [OSAC-3491](https://redhat.atlassian.net/browse/OSAC-3491) already moved/renamed to root in preparation for this, previously unused).
- Kept as a **single job**, unlike osac-operator's chart-publish/release split -- osac-aap has only one chart, so there's no multi-chart parallelism to justify separating release creation into its own job. This also preserves the old job's double tag-race check (once before packaging, again immediately before the release), which is a real race guard given real wall-clock time passes between the two, not incidental duplication.

## Verification
Read all four current files fresh (`execution-environment.yml`, `build-image.yaml`, `publish-image.yaml`, `publish-charts.yaml`) rather than assuming their shape, and the old standalone repo's `publish-charts.yaml` in full before porting anything. YAML parse check + `yamllint -c .yamllint.yaml --strict` on both changed files -- clean.

## GHCR "Manage Actions access" gap -- checked proactively, looks like it's needed
Per the same class of gap that hit fulfillment-service/osac-operator twice already (container images, then chart packages): checked `osac-aap`'s and `charts/osac-aap`'s package linkage via the GitHub Packages API before attempting any real tag push.

Both packages currently show `repository: osac-project/osac-aap` (the old standalone repo) as their origin. Note this field doesn't reliably indicate whether a "Manage Actions access" grant for `osac-project/osac` already exists or not -- it reflects historical package origin, not the current grant list, and there's no REST API to read the actual grant list (this is UI-only). Given osac-operator/fulfillment-service needed this grant added manually before their real-tag tests could push successfully, and nobody would have had reason to add it for osac-aap before this ticket wired it into the pipeline, I'm flagging this now rather than discovering a 403 mid-test: **`osac-aap` (container) and `charts/osac-aap` likely both need `osac-project/osac` added under Package Settings -> Manage Actions access -> Add repository, with Write role**, same as was needed for the other two components.

## Testing plan (not yet done -- pending the access grant above)
Once merged to main (workflow file content is pinned to the tag's target commit, so testing against a stale/unmerged tag wouldn't exercise this fix) and the GHCR grant is confirmed, will push a real `osac-aap/vX.Y.Z` tag (next in sequence: `osac-aap/v0.0.12`, following the old repo's last release `v0.0.11`) and confirm the image builds and pushes to `ghcr.io/osac-project/osac-aap`, the chart publishes to `ghcr.io/osac-project/charts/osac-aap`, and a GitHub Release is created -- then report back and close [OSAC-3511](https://redhat.atlassian.net/browse/OSAC-3511).

Not merging this myself.

### DIFF
--- a/.github/workflows/execution-environment.yml
+++ b/.github/workflows/execution-environment.yml
@@ -10,7 +10,7 @@ on:
     branches:
       - main
     tags:
-      - 'v*'
+      - 'osac-aap/v*'
   pull_request:
     branches:
       - main
@@ -50,6 +50,29 @@ jobs:
                                             --tag "osac-ee"         \
                                             -f "./execution-environment/execution-environment.yaml"
 
+      # Component-scoped tags (see the `tags:` trigger above) can't be used verbatim
+      # as image tags -- Docker/OCI tags can't contain the '/' the git tag does --
+      # so strip the prefix here to recover the bare vX.Y.Z for the actual image tags.
+      - name: Compute release version
+        if: startsWith(github.ref, 'refs/tags/osac-aap/')
+        run: |
+          version="${GITHUB_REF_NAME#osac-aap/}"
+          # Same SemVer rule as publish-charts.yaml's guard job -- rejects build
+          # metadata (+...) and validates the core/prerelease shape before it's
+          # trusted as an image tag.
+          semver_re='^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-((0|[1-9][0-9]*)|[0-9A-Za-z-]*[A-Za-z-][0-9A-Za-z-]*)(\.((0|[1-9][0-9]*)|[0-9A-Za-z-]*[A-Za-z-][0-9A-Za-z-]*))*)?$'
+          if ! [[ "${version}" =~ ${semver_re} ]]; then
+            echo "::error::Tag '${GITHUB_REF_NAME}' is not a valid semver release tag"
+            exit 1
+          fi
+          # Prerelease identifiers can themselves contain dots (e.g. v1.2.3-beta.1),
+          # so derive major/minor from the prerelease-stripped core version --
+          # otherwise a dotted prerelease would corrupt the major.minor alias.
+          core="${version%%-*}"
+          echo "RELEASE_VERSION=${version}" >> "$GITHUB_ENV"
+          echo "RELEASE_MAJOR_MINOR=${core%.*}" >> "$GITHUB_ENV"
+          echo "RELEASE_MAJOR=${core%%.*}" >> "$GITHUB_ENV"
+
       - name: Extract metadata (tags, labels)
         if: github.event_name != 'pull_request'
         id: meta
@@ -60,11 +83,15 @@ jobs:
           # workflow post-merge (see OSAC-3490), colliding on mutable tags.
           # Keeping this component's pre-merge image identity stable instead.
           images: ghcr.io/osac-project/osac-aap
+          # No `type=ref,event=tag` here: the `tags:` trigger above only ever
+          # matches this component's own scoped prefix, so the tag-triggered
+          # case is always covered by the three raw entries below already --
+          # keeping it would just add a redundant, slash-sanitized duplicate
+          # tag (e.g. "osac-aap-v1.2.3").
           tags: |
-            type=semver,pattern=v{{version}}
-            type=semver,pattern=v{{major}}.{{minor}}
-            type=semver,pattern=v{{major}}
-            type=ref,event=tag
+            type=raw,value=${{ env.RELEASE_VERSION }},enable=${{ env.RELEASE_VERSION != '' }}
+            type=raw,value=${{ env.RELEASE_MAJOR_MINOR }},enable=${{ env.RELEASE_VERSION != '' }}
+            type=raw,value=${{ env.RELEASE_MAJOR }},enable=${{ env.RELEASE_VERSION != '' }}
             type=sha
             type=raw,value=latest,enable=${{ github.ref == format('refs/heads/{0}', 'main') }}
 
--- a/.github/workflows/publish-charts.yaml
+++ b/.github/workflows/publish-charts.yaml
@@ -29,16 +29,9 @@ jobs:
   # component's image-build workflow fired. Each publish job below then
   # gates on github.event.workflow_run.name to only publish its own
   # component's chart -- the image-build workflows themselves stay separate
-  # per component (see build-image.yaml/publish-image.yaml), so this is the
-  # one place their otherwise-independent tag-triggered releases converge.
-  #
-  # osac-aap's execution-environment.yml is NOT wired in here yet: PR #29
-  # (OSAC-3467) is reworking this guard job to use component-scoped tags
-  # (<component>/vX.Y.Z) and a shared `version` output instead of the bare
-  # `v*` matching below, and is still open/under review. Wiring osac-aap in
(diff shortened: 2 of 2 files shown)


eliorerz:
Reviewed from scratch — pulled the branch fresh, read the full current content of both files, and diffed the new SemVer/tag-stripping block character-for-character against build-image.yaml's and publish-image.yaml's already-proven versions, plus pulled the old standalone osac-aap repo's actual `publish-aap-chart` job from git history to compare against the new port line by line. Everything checks out.

**(1) The tag-stripping/SemVer block is verified equivalent, not just similar:**
```
$ diff <(execution-environment.yml's block) <(build-image.yaml's block)
1c1
<   if: startsWith(github.ref, 'refs/tags/osac-aap/')
---
>   if: startsWith(github.ref, 'refs/tags/osac-operator/')
3c3
<   version="${GITHUB_REF_NAME#osac-aap/}"
---
>   version="${GITHUB_REF_NAME#osac-operator/}"
```
Same result diffing against `publish-image.yaml`. The *only* differences anywhere are the component-prefix string itself — the SemVer regex, the prerelease-stripped `core=`/`RELEASE_VERSION`/`RELEASE_MAJOR_MINOR`/`RELEASE_MAJOR` derivation are byte-for-byte identical to the already-proven versions.

**(2) `type=semver` → `type=raw` conversion confirmed complete, no latent bug left.** Reasoned through why the old `type=semver,pattern=v{{version}}` entries would silently produce *zero* tags once the trigger became `osac-aap/v*`: docker/metadata-action's semver matcher needs the raw git ref to itself parse as semver, and `osac-aap/v1.2.3` (containing a `/`) never will — so if this conversion had been missed, every real tag push would build an image but push it with no version tags at all (only `sha-*`/`latest` would still work), a real would-be latent bug. Grepped the full file for any remaining `type=semver` or a live (non-comment) `type=ref,event=tag` — zero hits; both are fully gone, only survives as the explanatory comment text.

One thing I checked and confirmed is *not* a problem: `execution-environment.yml`'s "Extract metadata" step is gated `if: github.event_name != 'pull_request'` and its `tags:` list has no `type=ref,event=branch`/`type=ref,event=pr` entries, unlike the other two files. Confirmed via diff context that this gate and the missing entries both **predate this PR** (untouched lines) — osac-aap's build workflow has never pushed per-PR images, and this PR correctly leaves that pre-existing architecture alone rather than "fixing" something that was never broken for this component.

**(3) `publish-osac-aap-chart` verified against both the old standalone job and the current file's conventions, line by line.** Pulled the actual old `osac-aap/.github/workflows/publish-charts.yaml` from git history (commit before OSAC-3491's consolidation) and diffed the ported job against it. Every difference is a deliberate, necessary adaptation:
- `Extract version from tag`: old sourced `${TAG#v}` from the (pre-monorepo, always-bare) `tag` output; new correctly sources `${VERSION#v}` from `needs.guard.outputs.version` (already prefix-stripped) instead — using the old pattern against the new *scoped* tag (`osac-aap/v1.2.3`) would have been a no-op strip, corrupting the version. Matches the exact pattern already proven in this file's own `publish-operator-crds-chart`/`publish-operator-chart` jobs.
- `Create GitHub Release`: old used `--title "osac-aap ${TAG}"` (fine when TAG was bare); new correctly uses `--title "osac-aap ${VERSION}"` while still passing the *full* scoped `${TAG}` to `gh release create` itself (where the real git ref is required) — same split already proven in `create-operator-release`.
- `working-directory: osac-aap` added to the three steps that touch chart files (`Update image tag`, `Package chart`, `Push chart to GHCR`) — necessary now that the chart lives one level deeper in the monorepo; body of each step otherwise unchanged.
- Script path updated to `.github/scripts/osac-aap-verify-tag-matches-sha.sh` — confirmed this file exists and its content is byte-for-byte identical to the old repo's `verify-tag-matches-sha.sh` (pure rename via OSAC-3491, nothing behavioral changed).
- The double tag-race check (once before packaging, once immediately before release creation) is preserved intact — both copies present in the new job, matching the old job's structure exactly.

Confirmed `if: success() && github.event.workflow_run.name == 'Build execution environment'` matches this repo's actual workflow name — read `execution-environment.yml`'s `name:` field directly: `name: "Build execution environment"`. Exact match, not assumed.

Confirmed the guard job's `startsWith`/case-statement additions are correctly scoped: `osac-aap/v` added via `||` to the existing two-component `startsWith` condition, and `osac-aap/v*) VERSION="${HEAD_BRANCH#osac-aap/}" ;;` added as a new case arm before the `*)` fallback — consistent with the existing two entries.

**(4) Confirmed purely additive.** `git diff` shows zero lines touching `publish-service-chart`, `publish-operator-crds-chart`, `publish-operator-chart`, or `create-operator-release` — grepped explicitly for those job names in the diff, no matches. The only changes to pre-existing code are the guard job's condition/case-statement additions; everything else is either a new job appended at the end or the isolated `execution-environment.yml` changes.

**Not yet tested with a real tag — correctly flagged as such, not claimed otherwise.** The PR description is explicit that testing is pending a GHCR "Manage Actions access" grant it proactively identified as likely needed (same class of gap that hit the other two components), and states outright "Not merging this myself." Noting this as an open item per your instructions, not a review blocker — the static analysis above stands on its own regardless of when the real-tag test happens.

Also confirmed independently: both files parse as valid YAML, no merge conflicts against current `main`, and current CI (Analyze (python), CodeQL, dependency-review, pre-commit, all three e2e `changes` gates) is green; `Analyze (go)` and the three e2e suites are still running at review time, nothing failing so far.

**Verdict: correct, faithfully and carefully ported (not a blind copy-paste), and fully additive.** Only open item is the pending real-tag verification, which the PR already flags honestly rather than claiming done.

Issue #45, opened 2026-05-19
BareMetalPool controller needs to know network class


ajamias:
Currently, the BareMetalPool controller does not know what type of network class it wants to use when calling templates that create networks.

PR #48, opened 2026-07-31
OSAC-3395: Merge bare-metal-fulfillment-operator into osac mono-repo

eliorerz:
## Summary

Merges bare-metal-fulfillment-operator (BMF) into this mono-repo as `bare-metal-fulfillment-operator/`, following the same playbook as fulfillment-service+osac-operator (#3) and osac-aap (#30).

- **Full history preserved**: `git subtree add --prefix=bare-metal-fulfillment-operator https://github.com/osac-project/bare-metal-fulfillment-operator.git main` (no `--squash`). All 109 commits from BMF's `main` are reachable via this merge commit's second parent -- verified against `gh api repos/osac-project/bare-metal-fulfillment-operator/commits --paginate` for an exact count match (109/109).
- **No tags carried over**: 0 tags before and after.
- **bare-metal-fulfillment-operator/OWNERS** came through automatically via the subtree add, same as every prior component.
- **Blame-diff check**: diffed the merged `bare-metal-fulfillment-operator/` tree against a fresh clone of the source repo's `main` at the fetched commit -- zero unexpected differences, only the 3 files intentionally touched below.

## The real circular dependency (this merge's actual risk, per investigation on [OSAC-3395](https://redhat.atlassian.net/browse/OSAC-3395))

Unlike the prior two merges, this one involves a genuine 3-way circular Go module dependency: BMF requires `osac-operator` (+`/api`), and `osac-operator` requires BMF -- confirmed both directions have real source-level package imports (BMF's `api/v1alpha1` imports `osac-operator/api/v1alpha1`; `osac-operator`'s `cmd`/`internal/controller` import BMF's `api/v1alpha1`), but no *package-level* cycle (Go would refuse to build that regardless of workspace).

**Prototype, done before touching the real merge**: scratch-cloned all three repos, built a shared `go.work`, and:
- Confirmed `go list -m -f '{{.Dir}}'` resolves all cross-module refs (BMF from osac-operator, osac-operator+api from BMF, BMF from fulfillment-service) to the local workspace clones, not the registry-pinned versions.
- `go build ./...` succeeds cleanly in all 4 modules (bmf, fulfillment-service, osac-operator, osac-operator/api).
- **Positive control**: renamed an exported symbol (`AddToScheme`) in the local BMF copy that osac-operator's `cmd/main.go` actually calls, rebuilt osac-operator, and got a real compile failure (`undefined: bmfov1alpha1.AddToScheme`) -- proving the workspace resolution is live and correctly linked, not coincidentally matching cached/pinned versions. Reverted after.

Re-ran the identical `go build ./...` verification against the real merged tree (not just the scratch prototype) -- same clean result.

## go.work

Added `./bare-metal-fulfillment-operator` to the existing `use` block (it has no separate `api` submodule like osac-operator does -- its `api/v1alpha1` package lives in the single main module).

## Pre-commit/yamllint reconciliation

BMF's own `.pre-commit-config.yaml`/`.yamllint.yaml` weren't purely redundant with root's, same situation as the osac-aap merge:
- Its `charts/operator/templates/` uses Helm templating and its `config/` holds kubebuilder-generated CRDs/RBAC -- both already have an established exclude pattern for `osac-operator`'s identical structure; extended that same pattern (in both the yamllint hook's `exclude` and root `.yamllint.yaml`'s `ignore:` list) to cover BMF too, rather than inventing a new one.
- Added a `bare-metal-fulfillment-operator-golangci-lint` local hook mirroring `osac-operator-golangci-lint` exactly.
- Deleted BMF's now-redundant `.pre-commit-config.yaml`/`.yamllint.yaml`, matching the osac-aap precedent.
- `end-of-file-fixer` found and fixed one real pre-existing issue: a missing trailing newline in `bare-metal-fulfillment-operator/hack/boilerplate.go.txt`.
- Validated: `pre-commit run --all-files` passes clean across the whole merged tree (yamllint, generic hooks). `buf-lint`/`golangci-lint` hooks fail only on missing binaries in this sandbox -- same known, pre-existing, unrelated gap noted in the osac-aap PR.

## Minimal CI wiring (full release-pipeline wiring deferred, same as osac-aap/OSAC-3511)

- **helm-lint.yaml**: BMF has the *exact same* kubebuilder CRD-to-Helm sync mechanism as osac-operator (`hack/sync-helm-crds.py` + `make check-helm-crds`) -- this made the existing job's "osac-operator-only, nothing to matrix against yet" comment factually outdated, so converted `helm-crds-sync` from a single hardcoded job into a proper `[osac-operator, bare-metal-fulfillment-operator]` matrix, and added BMF's two charts (`operator`, `operator-crds`) to the `helm-lint` chart matrix. Verified locally: both charts `helm lint`/`helm template` clean bare (no values-file guard needed, same as osac-operator).
- **codeql.yml**: added BMF to the manual Go build step (`go build ./...` per module list) so CodeQL's autobuild-replacement covers it.
- **check-floating-tags.yaml**: added BMF to the matrix -- verified locally, `check-floating-tags.sh` passes clean against its chart values.yaml.
- **Deliberately NOT wired**: `check-generated-code.yaml` (buf/protobuf-specific; BMF uses controller-gen, not buf -- doesn't apply), any standalone unit/integration-test workflow (osac-operator's and osac-aap's actual test execution lives inside their own build-image.yaml release pipelines, not a separate CI-only workflow -- BMF's equivalent belongs in its own future build-image.yaml, deferred with the rest of the release wiring).
- Local-only caveat: couldn't fully exercise `make check-helm-crds` for BMF in this sandbox (a local Go-toolchain-auto-download quirk tied to BMF's pinned `controller-tools v0.18.0`, unrelated to this PR) -- confirmed instead that this exact check already passes repeatedly in BMF's own standalone-repo CI today.

## GitHub Issues

BMF had exactly one open issue. Transferred it via GitHub's native issue-transfer (same org, so no manual recreation needed) rather than leaving it stranded in a repo slated for eventual archival: `bare-metal-fulfillment-operator#16` -> `osac-project/osac#45`, title/body/state preserved.

## Flagging, not deciding (per [OSAC-3395](https://redhat.atlassian.net/browse/OSAC-3395))

- **`bmf.enabled` conditional in osac-installer**: BMF's chart is currently wired as `condition: bmf.enabled` (same tier as `osac-ui`), one tier below the unconditionally-required fulfillment-service/osac-operator/osac-aap. Now that BMF is joining the mono-repo, should it become unconditionally required too? Not changing this myself -- leaving it for a maintainer decision, out of scope for osac-installer anyway (this PR only touches osac-project/osac).
- **osac-aap's `create-api-token.yaml` hook** (`oc rollout restart deploy/bmf-operator-controller-manager`) depends on BMF's Helm fullname staying `bmf-operator`. Verified `fullnameOverride: "bmf-operator"` in `charts/operator/values.yaml` is untouched by this subtree merge (values.yaml content isn't modified by `git subtree add`), so this hook's target name doesn't change. No action needed, just confirming it stays stable through this specific PR.
- **Hardcoded `image.repository` values** for BMF (currently at `bare-metal-fulfillment-operator/charts/operator/values.yaml:4`, not line 19 as originally noted on the ticket -- the file has since been edited upstream) and the duplicate in osac-installer's `charts/osac/values.yaml:225` are the same pre-existing hardcoded-values-file pattern already tracked for other components (OSAC-3490-style). Both live outside this repo's scope for line 225 (osac-installer) or are simply carried over unchanged from BMF's own repo for line 4 -- not touched by this PR.

## Important: do not merge yet

This repo's branch protection has `required_linear_history: true`. A full non-squashed subtree merge is inherently full of merge commits, which GitHub's "Rebase and merge" cannot linearize -- same wall hit by both prior component merges. Merging this requires a brief, deliberate `required_linear_history: false` -> merge-commit -> `required_linear_history: true` toggle, done with a human watching, not autonomously. **Leaving this PR open, unmerged, pending that step.**

Part of [OSAC-1732](https://redhat.atlassian.net/browse/OSAC-1732) / [OSAC-3395](https://redhat.atlassian.net/browse/OSAC-3395).

## Test plan
- [ ] Toggle `required_linear_history` off, merge via merge commit, restore it immediately after
- [ ] Confirm CI passes post-merge on `main`
- [ ] Decide (separately) whether to make BMF's chart unconditionally required in osac-installer
- [ ] Decide (separately, later) when to archive the standalone `bare-metal-fulfillment-operator` repo, once CI is green here


eliorerz:
Investigated all 3 failing checks to root cause before touching anything, per the checklist:

**1. pre-commit** -- my own oversight, not a code issue. `.github/workflows/pre-commit.yaml` has a dedicated `make -C osac-operator golangci-lint` step that downloads `bin/golangci-lint` *before* the pre-commit hooks run (that's why `osac-operator-golangci-lint` passes in real CI despite failing in my own local sandbox, which lacks that setup step). I added the `bare-metal-fulfillment-operator-golangci-lint` pre-commit hook in the original merge but forgot the matching setup step. Fixed by adding `make -C bare-metal-fulfillment-operator golangci-lint` alongside the existing one.

**2. dependency-review** -- a genuine, valid finding, not a tooling gap: `google.golang.org/grpc@1.81.1` (indirect, in BMF's own `go.mod`) is below the patched 1.82.1 for GHSA-hrxh-6v49-42gf (high severity). Confirmed osac-operator (1.82.1) and fulfillment-service (1.83.0) are already past the fix -- this is specific to BMF's stale dependency, never caught because BMF's standalone repo never ran this check. Per convention, **not fixing this inside a code-relocation PR** -- filed [OSAC-3516](https://redhat.atlassian.net/browse/OSAC-3516) to track the actual bump. Added a scoped, commented `allow-ghsas` carve-out referencing that advisory (and pointing at the follow-up ticket) so this PR can merge without silently normalizing the vulnerability.

**3. CodeQL** -- likely a false positive in BMF's own pre-existing code, also not fixed here. `hack/sync-helm-operator.py:338` flags "clear-text storage of sensitive information" on a script that only ever rewrites Kubernetes Secret/ConfigMap *names* into Helm template *expressions* -- never touches actual secret payload values. BMF's standalone repo has no CodeQL workflow at all, so this was never previously evaluated by anyone (matches the hypothesis that this predates the merge). Rather than assume my own read is correct and add an inline suppression to BMF's source as a side effect of this PR, filed it in [OSAC-3516](https://redhat.atlassian.net/browse/OSAC-3516) for an actual security review, and added a commented `paths-ignore` carve-out for just this one file to unblock the merge.

Both carve-outs are scoped as narrowly as possible (one specific GHSA, one specific file) and documented in-line with why and where the real fix is tracked, so they're easy to find and remove once [OSAC-3516](https://redhat.atlassian.net/browse/OSAC-3516) is resolved.

PR #49, opened 2026-07-31
OSAC-3511: Add Build execution environment to publish-charts.yaml's workflow_run trigger

eliorerz:
## Summary
Found while live-testing PR #44's osac-aap wiring: pushing `osac-aap/v0.0.12` built and pushed the image successfully, but never triggered a `Publish charts` run at all.

## Root cause
PR #44 correctly added osac-aap's guard condition (`startsWith(head_branch, 'osac-aap/v')`) and the `publish-osac-aap-chart` job, but never added `Build execution environment` (osac-aap's actual image-build workflow name) to `publish-charts.yaml`'s top-level `on: workflow_run: workflows: [...]` list. `workflow_run` only fires for workflows explicitly named there -- without this, the entire osac-aap chart-publish path was dead code regardless of the job-level guard being correct.

## Fix
Added `"Build execution environment"` to the `workflows:` list. One line.

## Test plan
- [ ] Push a fresh `osac-aap/vX.Y.Z` tag against this fix and confirm a `Publish charts` run actually fires and completes the `publish-osac-aap-chart`/guard jobs this time.

### DIFF
--- a/.github/workflows/publish-charts.yaml
+++ b/.github/workflows/publish-charts.yaml
@@ -16,7 +16,11 @@ name: Publish charts
 
 on:
   workflow_run:
-    workflows: ["Container image", "Build container image"]
+    # OSAC-3511 added osac-aap's guard condition and publish-osac-aap-chart
+    # job, but missed adding its actual build workflow's name here --
+    # workflow_run only fires for workflows explicitly listed, so without
+    # this the osac-aap chart-publish path silently never triggers at all.
+    workflows: ["Container image", "Build container image", "Build execution environment"]
     types: [completed]
 
 env:
--- a/.github/workflows/publish-charts.yaml
+++ b/.github/workflows/publish-charts.yaml
@@ -16,10 +16,10 @@ name: Publish charts
 
 on:
   workflow_run:
-    # OSAC-3511 added osac-aap's guard condition and publish-osac-aap-chart
-    # job, but missed adding its actual build workflow's name here --
-    # workflow_run only fires for workflows explicitly listed, so without
-    # this the osac-aap chart-publish path silently never triggers at all.
+    # workflow_run only fires for workflows explicitly listed here -- the
+    # guard condition and publish-osac-aap-chart job alone aren't enough
+    # to trigger this workflow at all without osac-aap's build workflow
+    # name also present in this list.
     workflows: ["Container image", "Build container image", "Build execution environment"]
     types: [completed]
 


eliorerz:
Reviewed from scratch, same rigor as any worker PR — pulled the branch fresh (note: this one's from a branch in `osac-project/osac` itself, not a fork, unlike the recent run of PRs), read the actual pre- and post-fix file content, and cross-checked every workflow name referenced anywhere in this file against the real `name:` field of its source workflow.

**(1) Confirmed the bug is real, not just claimed.** Current `main`'s `publish-charts.yaml`:
```
workflows: ["Container image", "Build container image"]
```
`"Build execution environment"` is genuinely absent. Given `workflow_run` only fires for workflows explicitly named in this list, this confirms the whole osac-aap chart-publish path (guard condition + `publish-osac-aap-chart` job, both added correctly by #44) was dead on arrival regardless of any tag pushed — matches the live symptom described (image built and pushed, no `Publish charts` run at all).

**(2) Confirmed exact name match.** `execution-environment.yml`'s actual `name:` field: `name: "Build execution environment"` — character-for-character identical to what this PR adds to the list (including capitalization).

**(3) Checked every other workflow-name reference in the file for the same class of bug — none found; this was an isolated omission, not a symptom of a broader pattern:**
```
$ grep -n "workflow_run.name ==" publish-charts.yaml
84:   if: ... == 'Container image'
136:  if: ... == 'Build container image'
189:  if: ... == 'Build container image'
248:  if: ... == 'Build container image'
291:  if: ... == 'Build execution environment'   (added by #44)
```
Cross-checked against the real source files:
- `publish-image.yaml`'s actual `name:` → `Container image` — matches both the job-level check (line 84) and the top-level list. Correct and complete already.
- `build-image.yaml`'s actual `name:` → `Build container image` — matches all three job-level checks (136/189/248) and the top-level list. Correct and complete already.
- `execution-environment.yml`'s actual `name:` → `Build execution environment` — matches the job-level check (291, from #44); only the top-level list was missing it, which is exactly what this PR fixes.

So fulfillment-service's and osac-operator's entries were never broken — this really was a single missed line specific to osac-aap's onboarding, not evidence of a systemic gap in how this file tracks workflow names.

**(4) No ticket ID or other leftover anywhere in the current file.** `grep -n "OSAC-[0-9]" publish-charts.yaml` → no hits. Also pulled both individual commits to see the actual before/after: the first commit's comment did originally read "[OSAC-3511](https://redhat.atlassian.net/browse/OSAC-3511) added osac-aap's guard condition...", and the second commit correctly reworded it to describe the *mechanism* (`workflow_run only fires for workflows explicitly listed here...`) without referencing the ticket — a clean fix, not a partial one. The final diff against `main` is exactly the one-line list addition plus that 4-line ticket-ID-free comment, nothing else.

Also independently confirmed: the file parses as valid YAML and there are no merge conflicts against current `main`.

**Verdict: confirmed correct on all four points.** Genuine bug, exact and complete fix, isolated to osac-aap (not a broader pattern), and the ticket-ID follow-up cleanly resolved what it set out to fix. Good to merge — the PR's own test plan (push a real tag and confirm `Publish charts` actually fires this time) is the right next step and isn't something I can substitute for from static review alone.

PR #50, opened 2026-07-31
OSAC-3517: Add path-filter skip logic to unit-tests.yml and integration-tests.yml

eliorerz:
## Summary

`unit-tests.yml` and `integration-tests.yml` used a component-scoped `paths:` filter at the workflow-trigger level (e.g. `fulfillment-service/**` with OWNERS/LICENSE/*.md/docs/** exclusions). That's the exact anti-pattern [OSAC-3227](https://redhat.atlassian.net/browse/OSAC-3227) verified against: when a `paths:` filter lives on the trigger itself, a PR whose changed files never match means the whole workflow never gets a run registered at all — which permanently blocks merge if either job is ever made a required status check (the check just sits at "Expected" forever).

Replaced it with the same pattern already used (and already verified) by the 3 e2e workflows: an internal `changes` job running `dorny/paths-filter`, gating the actual test job(s) via `needs.changes.outputs.should-run` instead of gating the workflow trigger itself. Copied verbatim from `e2e-vmaas-full-install.yml`:

- Same pinned SHA: `dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706` (v4.0.2)
- Same `predicate-quantifier: 'every'`
- Same exclude list: `OWNERS`, `LICENSE`, `*.md`, `docs/**`
- Same `should-run` expression: `github.event_name != 'pull_request' || steps.filter.outputs.code == 'true'`

## Behavior note

This does change trigger scope slightly: previously these two workflows only ran for PRs touching their specific component directories (`fulfillment-service/`, or `fulfillment-service`+`osac-operator`+`osac-aap`). The ported filter (matching e2e's own filter exactly, per e2e's own comment about deferring true per-component scoping to the future path-based CI matrix in [OSAC-1734](https://redhat.atlassian.net/browse/OSAC-1734)) now runs them for any PR touching any non-doc file repo-wide, skipping only on doc-only changes (OWNERS/LICENSE/*.md/docs/**). This trades some unnecessary cross-component runs for closing the required-status-check risk class — consistent with how the e2e workflows already work.

push/schedule/workflow_dispatch triggers are unaffected — the `dorny/paths-filter` step only runs `if: github.event_name == 'pull_request'`, and `should-run` short-circuits to `true` otherwise, exactly as before.

## Verification

- Read both files' full current job structure directly (not assumed identical to the e2e workflows — `unit-tests.yml` has one job, `integration-tests.yml` has three independent per-component jobs since they use genuinely different test mechanisms).
- Traced the gating logic by hand for all three cases:
  - Doc-only PR (e.g. touches only `README.md`): fails the `!*.md` predicate under `every` → `code` output is `false` → `should-run` is `false` → job skipped.
  - Code-touching PR: passes all AND-conditions → `code` is `true` → `should-run` is `true` → job runs.
  - Non-`pull_request` trigger (schedule/workflow_dispatch): filter step doesn't run at all → `should-run` short-circuits to `true` via the `github.event_name != 'pull_request'` clause → job always runs.
- `actionlint` reports no issues on both files.
- `pre-commit run` (yamllint, etc.) passes.
- This PR's own diff (workflow files, not docs) should itself exercise the "should-run == true" path when its checks run.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -4,22 +4,6 @@ on:
   pull_request:
     branches:
     - main
-    paths:
-      - 'fulfillment-service/**'
-      - '!fulfillment-service/OWNERS'
-      - '!fulfillment-service/LICENSE'
-      - '!fulfillment-service/**/*.md'
-      - '!fulfillment-service/docs/**'
-      - 'osac-operator/**'
-      - '!osac-operator/OWNERS'
-      - '!osac-operator/LICENSE'
-      - '!osac-operator/**/*.md'
-      - '!osac-operator/docs/**'
-      - 'osac-aap/**'
-      - '!osac-aap/OWNERS'
-      - '!osac-aap/LICENSE'
-      - '!osac-aap/**/*.md'
-      - '!osac-aap/docs/**'
   workflow_dispatch:
   # 2x/day (every 12h). Offset 47 minutes past the hour, not on the hour --
   # GitHub Actions' shared cron scheduler queues everyone's on-the-hour
@@ -39,6 +23,26 @@ permissions:
 
 jobs:
 
+  changes:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    outputs:
+      should-run: ${{ github.event_name != 'pull_request' || steps.filter.outputs.code == 'true' }}
+    steps:
+      - uses: dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706  # v4.0.2
+        if: github.event_name == 'pull_request'
+        id: filter
+        with:
+          predicate-quantifier: 'every'
+          filters: |
+            code:
+              - '**'
+              - '!OWNERS'
+              - '!LICENSE'
+              - '!*.md'
+              - '!docs/**'
+
   # fulfillment-service, osac-operator, and osac-aap use genuinely different
   # integration test mechanisms (ginkgo directly, a Kind cluster + Makefile
   # target, and a Kind cluster + uv/pytest respectively), so these stay
@@ -46,6 +50,8 @@ jobs:
   # parameterize, only the workflow-level trigger/name were duplicated before.
   fulfillment-service:
     name: Run integration test (fulfillment-service)
+    needs: changes
+    if: needs.changes.outputs.should-run == 'true'
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
@@ -69,6 +75,8 @@ jobs:
 
   osac-operator:
     name: Run integration test (osac-operator)
+    needs: changes
+    if: needs.changes.outputs.should-run == 'true'
     runs-on: ubuntu-latest
     timeout-minutes: 30
     steps:
@@ -95,6 +103,8 @@ jobs:
 
   osac-aap:
     name: Run integration test (osac-aap)
+    needs: changes
+    if: needs.changes.outputs.should-run == 'true'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -38,10 +38,10 @@ jobs:
           filters: |
             code:
               - '**'
-              - '!OWNERS'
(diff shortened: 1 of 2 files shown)


eliorerz:
Reviewed from scratch — pulled the branch fresh, read the actual pre-change content of both files on `main` directly (not the PR description), and diffed the new `changes` job byte-for-byte against `e2e-vmaas-full-install.yml`'s already-proven version. (Note: this PR's branch is genuinely `BEHIND` current `main` by 117 commits — used `git diff $(git merge-base main <branch>) <branch>` to isolate this PR's actual change from unrelated churn since its branch point; confirmed that isolated diff is identical to `gh pr diff`'s output, so the review below is against the real, correct diff.)

**(1) Confirmed the old trigger-level `paths:` filter genuinely existed and has the claimed problem.** Read `main` directly:
- `unit-tests.yml`: `pull_request: paths: ['fulfillment-service/**', '!fulfillment-service/OWNERS', ...]`
- `integration-tests.yml`: same pattern, repeated for all three of `fulfillment-service/**`, `osac-operator/**`, `osac-aap/**`.

This is a real instance of the anti-pattern: a `paths:` filter on the trigger itself means GitHub registers *no run at all* for a PR that doesn't touch matching paths — not a skipped job, an absent one. If either job is ever made a required status check, such a PR's check sits at "Expected" forever with nothing that could ever resolve it. Confirmed this is exactly the class of problem [OSAC-3227](https://redhat.atlassian.net/browse/OSAC-3227) already established, not a fabricated pretext for the change.

**(2) Confirmed the new `changes` job is genuinely identical to the e2e workflows' proven version, byte-for-byte, not just similar:**
```
$ diff <(unit-tests.yml's changes job) <(e2e-vmaas-full-install.yml's changes job)
(empty)
$ diff <(integration-tests.yml's changes job) <(e2e-vmaas-full-install.yml's changes job)
(empty)
```
Same pinned SHA (`dorny/paths-filter@7b450...` / v4.0.2), same `predicate-quantifier: 'every'`, same `code:` filter list, same `should-run` expression. This is a real, verified port of the proven pattern, not a new implementation that merely resembles it.

**(3) Confirmed this is scoped to trigger/skip-logic only.** Grepped the isolated diff for anything outside the `paths:` removal, the new `changes:` job, and the `needs: changes`/`if: needs.changes.outputs.should-run == 'true'` additions on each existing job — nothing else touched. No changes to checkout steps, `setup-go`, `ginkgo run`, `make test-integration-kind`, or any other actual test-execution logic in either file.

**(4) Confirmed non-`pull_request` triggers are unaffected, and got live proof it actually works.** `workflow_dispatch:` and `schedule:` blocks are untouched in the diff; the `dorny/paths-filter` step is gated `if: github.event_name == 'pull_request'`, and `should-run` short-circuits to `true` via `github.event_name != 'pull_request' || ...` otherwise — identical to e2e's proven behavior. (Minor note: neither file actually has a `push:` trigger at all — only `pull_request`/`workflow_dispatch`/`schedule` — so there's no push case to separately worry about here.) Better than theory: checked this PR's own live CI run, since its diff touches workflow files (not docs), which should exercise the `should-run == true` path exactly as the PR predicts — confirmed via `gh pr checks`: `Run unit tests`, `Run integration test (fulfillment-service/osac-operator/osac-aap)`, and all `changes` jobs all **pass**. Real, live confirmation the gating logic works correctly, not just traced on paper.

Also independently confirmed: both files parse as valid YAML, and no merge conflicts against current `main` despite the branch being 117 commits behind.

**Verdict: confirmed correct on all four points, plus live CI evidence the new gating actually works.** Good to merge (worth a rebase given it's behind main, but no conflicts expected based on the merge-tree check).

eliorerz:
Pushed a follow-up commit fixing a bug found in [OSAC-3520](https://redhat.atlassian.net/browse/OSAC-3520) that this PR inherited: the ported exclude patterns (`'!OWNERS'`, `'!LICENSE'`, `'!*.md'`, `'!docs/**'`) only match repo-root paths, not nested per-component paths like `fulfillment-service/docs/AUTH.md` -- live-proven by #46, which touched only nested docs files yet still ran the 3 e2e suites in full. Same fix applied to the 3 e2e workflows directly in #54.

Now uses `'!**/OWNERS'`, `'!**/LICENSE'`, `'!**/*.md'`, `'!**/docs/**'` to match at any depth.

PR #51, opened 2026-07-31
OSAC-3519: Add bare-metal-fulfillment-operator to e2e-bmaas components array

eliorerz:
## Summary
`e2e-bmaas-full-install.yml`'s `components` array builds fulfillment-service, osac-operator, and osac-aap from the PR's own mono-repo source, but bare-metal-fulfillment-operator (BMF) was missing -- since [OSAC-3395](https://redhat.atlassian.net/browse/OSAC-3395) merged it in, e2e-bmaas has been exercising whatever stale/default BMF image `osac-installer`'s values pin, not the PR's actual BMF code. Same "testing old code, not the PR" gap already solved for the other 3 components.

This is the bmaas-only change, prioritized first per explicit instruction since it's the e2e suite that actually exercises BMF's own functionality (bare-metal-as-a-service). vmaas/caas follow-up covered separately below.

## Verified against actual current state, not guessed
- `bare-metal-fulfillment-operator/Containerfile` exists at the repo root -- confirmed directly.
- **imageKey**: cloned `osac-installer` and read `charts/osac/Chart.yaml` + `charts/osac/values.yaml` + `values/bmaas-ci/values.yaml` directly. BMF is wired into the umbrella chart as dependency alias `bmf` (`repository: file://../../base/bare-metal-fulfillment-operator/charts/operator`, `condition: bmf.enabled`), with its image at `bmf.image.repository` -- confirmed exact key path, not assumed from the osac-operator analogy alone.
- **Conditional wiring is not a blocker for bmaas**: `bmf.enabled: true` in both the umbrella chart's own defaults and `values/bmaas-ci/values.yaml` (the values file this e2e flow actually deploys with) -- BMF's chart is enabled, so the image override will actually apply to a running deployment.
- **`replace-installer-submodule.sh`'s `SUBMODULE_MAP`**: already has a `[bare-metal-fulfillment-operator]="bare-metal-fulfillment-operator"` entry (added ahead of this, presumably during [OSAC-3395](https://redhat.atlassian.net/browse/OSAC-3395)'s own merge prep) -- confirmed the mapped submodule directory (`base/bare-metal-fulfillment-operator`) also exists in `osac-installer`. No change needed in osac-test-infra for the bmaas case.
- Confirmed osac-test-infra's reusable `e2e-bmaas-full-install.yml` loops over the `components` array generically ("so any number of components") -- no hardcoded component count to work around.

## Change
Added a 4th `components` entry: `containerfile: "bare-metal-fulfillment-operator/Containerfile"`, `imageKey: "bmf.image.repository"` (no `buildType`, defaults to `containerfile`, matching fulfillment-service/osac-operator's entries -- only osac-aap uses `ansible-builder`). Updated the job's explanatory comment to mention BMF.

## Validation
YAML parse + `yamllint -c .yamllint.yaml --strict` clean. Also parsed the embedded `components` JSON string directly to confirm it's valid JSON with exactly 4 entries and the expected keys.

## Verification plan (not yet done)
Will trigger a real e2e-bmaas run (workflow_dispatch or a test PR) once this is up for review, and confirm from the actual logs that BMF's image is being built from this PR's source rather than pulled pre-built -- same standard used for the other 3 components' onboarding. Will report back with the run link.

## Follow-up (separate, not in this PR)
The ticket covers all 3 e2e workflows (vmaas/bmaas/caas) for consistency; bmaas is first per priority. One nuance worth flagging before I extend to the other two: `bmf.enabled: false` in both `values/vmaas-ci/values.yaml` and `values/caas-ci/values.yaml` -- BMF's chart isn't deployed at all in those flows today, so adding the same components entry there would be harmless but a functional no-op (no running BMF pod for the image override to ever reach). Will call this out explicitly when I get to those two rather than silently adding a no-op.

Not merging this myself.

### DIFF
--- a/.github/workflows/e2e-bmaas-full-install.yml
+++ b/.github/workflows/e2e-bmaas-full-install.yml
@@ -56,14 +56,14 @@ jobs:
               - '!*.md'
               - '!docs/**'
 
-  # Builds fulfillment-service, osac-operator, and osac-aap's execution
-  # environment from this PR's current code and installs them together in
-  # one cluster -- not one "component under test" plus stale pinned images
-  # for the others. Runs for any change touching any component (see
-  # `changes` above) rather than being path-scoped per component;
-  # splitting this by component to avoid running both suites when only one
-  # changed is deferred to the future full path-based CI matrix, once its
-  # skip-logic pattern is verified.
+  # Builds fulfillment-service, osac-operator, osac-aap's execution
+  # environment, and bare-metal-fulfillment-operator from this PR's current
+  # code and installs them together in one cluster -- not one "component
+  # under test" plus stale pinned images for the others. Runs for any
+  # change touching any component (see `changes` above) rather than being
+  # path-scoped per component; splitting this by component to avoid
+  # running both suites when only one changed is deferred to the future
+  # full path-based CI matrix, once its skip-logic pattern is verified.
   e2e-bmaas-full-install:
     needs: changes
     if: needs.changes.outputs.should-run == 'true'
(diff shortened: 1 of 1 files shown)


eliorerz:
Reviewed from scratch — pulled the branch fresh and independently verified every claim against the real osac-installer/osac-test-infra content, not the PR description.

**(1) New entry's shape confirmed identical to the existing 3, not just plausible.** `bare-metal-fulfillment-operator/Containerfile` confirmed present at repo root. Extracted every `"repo":"..."` and every `"ref":"..."` string across all 4 components-array entries with `grep -o ... | sort -u`: both collapse to exactly one distinct value each, meaning the new BMF entry uses the *exact same* `repo`/`ref` GitHub-expression template as the other three, character for character — not a plausible-looking but subtly different copy. `buildType` correctly omitted (defaults to `containerfile`, matching fulfillment-service/osac-operator; only osac-aap needs `ansible-builder`). Parsed the embedded JSON directly: valid, exactly 4 entries, expected keys.

**(2) `imageKey` verified against osac-installer's real schema, not the osac-operator analogy alone:**
- `charts/osac/Chart.yaml`: `bare-metal-fulfillment-operator` dependency has `alias: bmf`, `condition: bmf.enabled` — confirmed directly.
- `bare-metal-fulfillment-operator/charts/operator/values.yaml`: `image.repository: ghcr.io/osac-project/bare-metal-fulfillment-operator` at the top level — same shape as osac-operator's own chart.
- Combined, this confirms `bmf.image.repository` is the exact correct `--set-string` key at the umbrella level (subchart alias `bmf` + the subchart's own `image.repository` field) — same convention already proven for `operator.image.repository`. Not assumed from analogy; verified end-to-end through both the umbrella Chart.yaml and the BMF chart's own values.yaml.

**(3) `SUBMODULE_MAP` already had the entry — confirmed, no gap.** Read `osac-test-infra/.github/scripts/replace-installer-submodule.sh` directly: `[bare-metal-fulfillment-operator]="bare-metal-fulfillment-operator"` is already present, and the mapped `base/bare-metal-fulfillment-operator` directory exists in osac-installer. So this PR correctly needed zero osac-test-infra changes — confirmed rather than trusted. Also confirmed the reusable `e2e-bmaas-full-install.yml` loops generically via `COUNT=$(... | jq 'length')`, no hardcoded component count anywhere.

**(4) Conditional wiring confirmed not a blocker for bmaas, and the PR's own vmaas/caas no-op disclosure independently verified too:**
```
values/bmaas-ci/values.yaml:   bmf.enabled: true    <- this PR's flow
values/vmaas-ci/values.yaml:   bmf.enabled: false
values/caas-ci/values.yaml:    bmf.enabled: false
```
`bmf.enabled: true` in the values file this e2e flow actually deploys with, so the image override will reach a real running BMF pod, not a disabled subchart. I also independently confirmed the PR's own "Follow-up" note that vmaas/caas have `bmf.enabled: false` — accurate, not overstated; adding this same entry there today really would be a functional no-op exactly as flagged.

**(5) Not yet tested with a real e2e-bmaas run — correctly disclosed, not claimed otherwise.** The PR states this outright and says "Not merging this myself." At review time, this PR's own triggered `e2e-bmaas-full-install / e2e` check (which will exercise the new 4-component build as part of normal PR CI) is still `pending`; all other checks (Analyze go/python, CodeQL, dependency-review, pre-commit, `changes` gates) are green. Noting as an open item per your instructions, not a blocker — worth checking that run's logs once it completes to confirm BMF is actually built from PR source (not pulled pre-built), same standard already applied to the other 3 components.

Also confirmed independently: the diff is a single clean hunk (4th array entry + updated comment), valid YAML, JSON components string parses correctly.

**Verdict: confirmed correct on all four verifiable points** — exact entry-shape match, correct imageKey verified end-to-end through the real chart schema, no osac-test-infra gap, and accurate self-disclosure of both the pending real-tag test and the vmaas/caas scope boundary. Good, pending the e2e-bmaas run actually completing green.

PR #52, opened 2026-07-31
OSAC-3518: Wire bare-metal-fulfillment-operator into the tag-scoped release pipeline

eliorerz:
## Summary

bare-metal-fulfillment-operator (BMF) merged via #48 with minimal CI wiring only (build/lint checks) -- the actual tag-scoped release pipeline (image build, chart publish, GitHub Release) was deliberately deferred. This wires it in, mirroring [OSAC-3511](https://redhat.atlassian.net/browse/OSAC-3511)'s exact playbook for osac-aap (#44 + the #49 fix).

## build-bmf-image.yaml (new)

Ported from the old standalone repo's `build-image.yaml`, adapted to this repo's established per-component pattern (same shape as `build-image.yaml`/osac-operator and `publish-image.yaml`/fulfillment-service):
- Path-filtered PR trigger on `bare-metal-fulfillment-operator/**`.
- Tag trigger scoped to `bare-metal-fulfillment-operator/vX.Y.Z`, with the same tag-prefix-strip + SemVer-validation logic already proven on the other 3 components.
- `IMAGE_NAME` hardcoded to `osac-project/bare-metal-fulfillment-operator` (not `github.repository`) -- same [OSAC-3490](https://redhat.atlassian.net/browse/OSAC-3490) collision-avoidance pattern osac-operator/osac-aap already use.
- Test job runs `make test` only (BMF has no `test-kustomize`/`test-smoke` Makefile targets like osac-operator does, so this matches BMF's own old workflow rather than osac-operator's exact 3-step test job).
- Kustomize manifest-container generation/push, same as osac-operator's build-image.yaml.

**Important naming note:** BMF's *old standalone* build-image.yaml was named `"Build container image"` -- identical to osac-operator's existing mono-repo workflow name. Porting it verbatim would have made `publish-charts.yaml`'s `workflow_run.name == 'Build container image'` guard incorrectly also match BMF's builds, misrouting BMF's successful builds into osac-operator's chart-publish job. Renamed to `"Build bare-metal-fulfillment-operator image"` to avoid the collision.

## publish-charts.yaml

- Added `bare-metal-fulfillment-operator/v` to the guard job's `startsWith` condition and its tag-prefix-stripping `case` statement.
- Added `bare-metal-fulfillment-operator-verify-tag-matches-sha.sh` (identical content to the existing per-component scripts, matching the established convention of a dedicated script file per component rather than a shared one).
- Added three new jobs mirroring osac-operator's split (BMF has two charts -- `operator` and `operator-crds` -- same shape as osac-operator, unlike osac-aap's single-chart case): `publish-bmf-crds-chart`, `publish-bmf-chart` (includes the `sed` image-tag update in `charts/operator/values.yaml`, same placeholder pattern as osac-operator's chart), and `create-bmf-release`.
- **The critical thing explicitly called out in the ticket, verified rather than assumed:** added `"Build bare-metal-fulfillment-operator image"` to the top-level `on: workflow_run: workflows: [...]` list -- this exact list is what was missed for osac-aap in #44 (the job-level `if:` conditions alone don't make `workflow_run` fire at all) and had to be fixed live in #49 after osac-aap's chart-publish silently never ran. Confirmed by grep that the workflow's own `name:` field and all 4 references to it in `publish-charts.yaml` (the trigger list plus 3 job conditions) match byte-for-byte.

## Validation

- YAML parses clean, `yamllint --strict` clean on both changed workflow files.
- `pre-commit run` clean on all changed files (golangci-lint/buf-lint hooks skipped -- no matching files changed).
- Confirmed BMF's chart names (`bare-metal-fulfillment-operator`, `bare-metal-fulfillment-operator-crds`) and the `tag: latest` placeholder in `charts/operator/values.yaml` match what the new jobs assume.

## Not yet done: real end-to-end tag test

Per the same constraint hit during [OSAC-3511](https://redhat.atlassian.net/browse/OSAC-3511): GitHub Actions resolves tag-triggered workflow files from the tagged commit's own tree, not the default branch -- so a real `bare-metal-fulfillment-operator/vX.Y.Z` tag can only exercise this new pipeline once it's actually on `main`. I don't have direct push access to `osac-project/osac` (everything in this session has gone through fork+PR), so I can't push a test tag against an unmerged commit. Once this merges, I'll push a real tag (next version per `gh api repos/osac-project/bare-metal-fulfillment-operator/releases/latest`) and confirm the full chain: image build -> GHCR push -> chart publish -> GitHub Release.

Part of OSAC-1732, follow-on to OSAC-3395, mirrors [OSAC-3511](https://redhat.atlassian.net/browse/OSAC-3511).

## Test plan
- [ ] Merge (same `required_linear_history` toggle dance as every prior component merge/wiring PR, if needed -- this one has no merge commits of its own so may not need it, but worth checking)
- [ ] Push a real `bare-metal-fulfillment-operator/vX.Y.Z` tag once merged and confirm: image build succeeds and pushes to `ghcr.io/osac-project/bare-metal-fulfillment-operator`, both charts publish to GHCR, a GitHub Release is created

### DIFF
--- a/.github/scripts/bare-metal-fulfillment-operator-verify-tag-matches-sha.sh
+++ b/.github/scripts/bare-metal-fulfillment-operator-verify-tag-matches-sha.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+# Verifies that a git tag still resolves to the guarded commit SHA that was
+# validated by this workflow's guard job, protecting against a force-push/
+# retag race between the image build and any subsequent chart-publish or
+# release step.
+#
+# Required env vars: GH_TOKEN, REPO, TAG, GUARDED_SHA
+set -euo pipefail
+
+ref_json="$(gh api "repos/${REPO}/git/ref/tags/${TAG}" --jq '[.object.type, .object.sha] | @tsv')"
+read -r current_type current_sha <<< "$ref_json"
+if [[ "$current_type" == "tag" ]]; then
+  # Annotated tag: the ref's object.sha is the tag object, not the commit - peel it.
+  current_sha="$(gh api "repos/${REPO}/git/tags/${current_sha}" --jq .object.sha)"
+fi
+
+if [[ "$current_sha" != "$GUARDED_SHA" ]]; then
+  echo "::error::Tag '${TAG}' now points at ${current_sha}, not the guarded commit ${GUARDED_SHA} (force-pushed?). Refusing to proceed."
+  exit 1
+fi
+
+echo "Tag '${TAG}' still points at the guarded commit ${GUARDED_SHA}."
--- a/.github/workflows/build-bmf-image.yaml
+++ b/.github/workflows/build-bmf-image.yaml
@@ -1,0 +1,211 @@
+name: Build bare-metal-fulfillment-operator image
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - 'bare-metal-fulfillment-operator/**'
+      - '!bare-metal-fulfillment-operator/OWNERS'
+      - '!bare-metal-fulfillment-operator/LICENSE'
+      - '!bare-metal-fulfillment-operator/**/*.md'
+      - '!bare-metal-fulfillment-operator/docs/**'
+  push:
+    branches:
+    - main
+    tags:
+    - 'bare-metal-fulfillment-operator/v*'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+env:
+  # Use docker.io for Docker Hub if empty
+  REGISTRY: ghcr.io
+  # Hardcoded rather than github.repository: that now resolves to
+  # osac-project/osac for every component's build workflow post-merge,
+  # colliding on mutable tags with other components' images. Keeping this
+  # component's pre-merge image identity stable instead, same as
+  # osac-operator's build-image.yaml and osac-aap's execution-environment.yml do.
+  IMAGE_NAME: osac-project/bare-metal-fulfillment-operator
+
+
+jobs:
+  test:
+    name: Run Tests
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: 'bare-metal-fulfillment-operator/go.mod'
+
+      - name: Install kind
+        uses: helm/kind-action@v1
+        with:
+          install_only: true
+
+      - name: Run unit tests
+        working-directory: bare-metal-fulfillment-operator
+        run: make test
+
+  build:
+    needs: test
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
+
+      # Login against a Docker registry except on PR
+      # https://github.com/docker/login-action
+      - name: Log into registry ${{ env.REGISTRY }}
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v4
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      # Component-scoped tags (see the `tags:` trigger above) can't be used verbatim
+      # as image tags -- Docker/OCI tags can't contain the '/' the git tag does --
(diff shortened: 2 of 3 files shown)


eliorerz:
Reviewed from scratch — pulled the branch fresh, read the full diff, and diffed the new SemVer/tag-stripping block byte-for-byte against the already-proven versions. Special attention on point (2) since it's the exact bug class PR #49 had to fix live for osac-aap.

**(1) `build-bmf-image.yaml`'s tag-stripping/SemVer block confirmed equivalent, not a subtly different copy:**
```
$ diff <(build-bmf-image.yaml's Compute-release-version block) <(build-image.yaml's block)
2c2
<   if: startsWith(github.ref, 'refs/tags/bare-metal-fulfillment-operator/')
---
>   if: startsWith(github.ref, 'refs/tags/osac-operator/')
4c4
<   version="${GITHUB_REF_NAME#bare-metal-fulfillment-operator/}"
---
>   version="${GITHUB_REF_NAME#osac-operator/}"
```
Only the component-prefix string differs; the SemVer regex and the prerelease-stripped `core=`/`RELEASE_VERSION`/`RELEASE_MAJOR_MINOR`/`RELEASE_MAJOR` derivation are identical. The `docker/metadata-action` `tags:` block is likewise identical to `build-image.yaml`'s (`type=raw,value=...,enable=...` pattern, no leftover `type=semver`), aside from the example string in one comment.

**(2) This is the critical one — verified both places match exactly, and the naming-collision-avoidance claim itself checked against real history, not taken on faith.**

Confirmed the workflow's actual `name:` field and cross-referenced it against every place `publish-charts.yaml` mentions it:
```
$ head -1 build-bmf-image.yaml
name: Build bare-metal-fulfillment-operator image

$ grep -n "bare-metal-fulfillment-operator image" publish-charts.yaml
24:  workflows: ["Container image", "Build container image", "Build execution environment", "Build bare-metal-fulfillment-operator image"]
401: if: success() && github.event.workflow_run.name == 'Build bare-metal-fulfillment-operator image'
454: if: success() && github.event.workflow_run.name == 'Build bare-metal-fulfillment-operator image'
513: if: success() && github.event.workflow_run.name == 'Build bare-metal-fulfillment-operator image'
```
All four references (the top-level trigger list plus the three job-level `if:` conditions) are byte-identical to the file's actual `name:` field. This is exactly the thing #49 had to fix after it was missed for osac-aap, and it's correctly present in both places here.

Also independently verified the PR's stated *reason* for renaming away from BMF's old name, rather than trusting it: pulled BMF's old standalone repo's `build-image.yaml` from git history (`git show 18d43adb:.../build-image.yaml`) — its `name:` really was `Build container image`, identical to osac-operator's current mono-repo workflow name. Had this PR ported that name verbatim, `publish-charts.yaml`'s existing `workflow_run.name == 'Build container image'` guard (used by osac-operator's three jobs) would have *also* matched BMF's builds — misrouting BMF's successful image builds into osac-operator's chart-publish/release jobs. This is a real, substantive catch, not just mirroring the osac-aap playbook — it required actually noticing the name reuse, and the fix (renaming to `Build bare-metal-fulfillment-operator image`) genuinely avoids it.

**(3) Confirmed purely additive to the existing components.** Diffed against the actual merge-base (this PR's `main` hadn't moved since its branch point, unlike #50, so a plain `git diff main <branch>` was already accurate) — zero lines touch `publish-service-chart`, `publish-operator-crds-chart`, `publish-operator-chart`, `create-operator-release`, or `publish-osac-aap-chart`. The only changes to pre-existing lines are the shared guard's `startsWith`/case-statement additions and the top-level `workflows:` list — everything else is new jobs appended at the end plus the new standalone `build-bmf-image.yaml`/verify-script files. Also confirmed the new verify-tag script is byte-identical to the existing `osac-operator-verify-tag-matches-sha.sh` (pure rename, no logic drift), and that BMF's actual chart names (`bare-metal-fulfillment-operator`, `bare-metal-fulfillment-operator-crds`) and the `tag: latest` placeholder in `charts/operator/values.yaml` match exactly what the new jobs assume.

**(4) Not yet tested with a real tag — correctly and specifically disclosed, not glossed over.** The PR explains *why* it can't test end-to-end itself (tag-triggered workflow content resolves from the tagged commit's tree, and it has no direct push access to `osac-project/osac` outside fork+PR) and commits to doing the real tag push once merged. Worth noting: since this PR only touches `.github/` files (not `bare-metal-fulfillment-operator/**`), the new `build-bmf-image.yaml`'s own `pull_request: paths:` filter means it was never even triggered by this PR's own CI run — there's genuinely no CI signal yet on the new file, consistent with the PR's own disclosure rather than a gap in what should have run. Flagging as an open item, not a blocker.

Also independently confirmed: both changed/new files parse as valid YAML, and there are no merge conflicts against current `main`.

**Verdict: confirmed correct on all four points, including the one specifically flagged as critical** — the workflow_run trigger-list omission that bit osac-aap was not repeated here, and the PR's own naming-collision catch checks out against real history. Good, pending the real-tag verification once merged.

PR #54, opened 2026-07-31
OSAC-3520: Fix dorny/paths-filter exclude globs to match nested docs paths

eliorerz:
## Bug

The shared `dorny/paths-filter` exclude list (`'!OWNERS'`, `'!LICENSE'`, `'!*.md'`, `'!docs/**'`) used by the 3 e2e workflows only matches files/directories sitting directly at the **repo root** — these are unanchored, non-`**`-prefixed globs, so they never match nested per-component paths like `fulfillment-service/docs/AUTH.md`, which is how virtually all real documentation actually lives in this mono-repo.

**Live-proven, not theoretical:** #46 touched only `fulfillment-service/docs/AUTH.md` and `fulfillment-service/docs/INSTALL.md` (pure documentation), yet `e2e-vmaas-full-install`, `e2e-bmaas-full-install`, and `e2e-caas-full-install` all ran in full (43m-1h1m each, real cluster provisioning) because the filter's negation patterns never matched those changed paths.

## Verified against the actual tree, not assumed

Checked whether `OWNERS`/`LICENSE` genuinely only exist at repo root today: they don't. Confirmed real files at multiple depths:
- Root: `OWNERS`, `LICENSE`
- Per-component (1 level): `fulfillment-service/OWNERS`, `osac-operator/OWNERS`, `osac-aap/OWNERS`, `bare-metal-fulfillment-operator/OWNERS` + `LICENSE`
- Vendored, several levels deep: `osac-aap/vendor/ansible_collections/*/LICENSE`, `osac-aap/collections/ansible_collections/osac/templates/LICENSE`

So the `**/` prefix isn't just future-proofing — it's necessary right now for paths that already exist in the tree.

## Fix

Prefixed each exclude pattern with `**/` in all 3 e2e workflows (`e2e-vmaas-full-install.yml`, `e2e-bmaas-full-install.yml`, `e2e-caas-full-install.yml`):

```diff
 code:
   - '**'
-  - '!OWNERS'
-  - '!LICENSE'
-  - '!*.md'
-  - '!docs/**'
+  - '!**/OWNERS'
+  - '!**/LICENSE'
+  - '!**/*.md'
+  - '!**/docs/**'
```

`**/` matches zero-or-more path segments, so this still correctly excludes the root-level case these patterns already handled, while now also matching any nested depth.

## unit-tests.yml / integration-tests.yml

These two inherited the identical bug via #50 ([OSAC-3517](https://redhat.atlassian.net/browse/OSAC-3517)), which is still open/unmerged. Since their buggy content only exists on that PR's branch (not yet on `main`), I pushed a corresponding fix commit directly onto that PR rather than duplicating the change here against non-existent base content — see #50's latest commit.

## Verification

- `actionlint` reports no issues on all 3 changed files.
- `pre-commit run` (yamllint, etc.) passes.
- Confirmed the OWNERS/LICENSE tree depths above directly (`find . -name OWNERS -o -name LICENSE`), rather than assuming.
- **On live-testing the "after" behavior**: cleanly isolating a nested-docs-only diff against this fix specifically isn't possible within a single PR here, since this PR's own diff necessarily includes real code (the workflow YAML changes themselves), which alone makes the filter's `code` output `true` regardless of any docs file also present — the two effects can't be cleanly separated in one diff against a base that doesn't yet have the fix. Once this merges, I'll open a follow-up docs-only test PR (touching e.g. `fulfillment-service/docs/**`) against the now-fixed `main`, the same way #46 exposed the original bug, to close the loop with a genuine live "after" proof -- and close that test PR without merging once confirmed.

### DIFF
--- a/.github/workflows/e2e-bmaas-full-install.yml
+++ b/.github/workflows/e2e-bmaas-full-install.yml
@@ -51,10 +51,10 @@ jobs:
           filters: |
             code:
               - '**'
-              - '!OWNERS'
-              - '!LICENSE'
-              - '!*.md'
-              - '!docs/**'
+              - '!**/OWNERS'
+              - '!**/LICENSE'
+              - '!**/*.md'
+              - '!**/docs/**'
 
   # Builds fulfillment-service, osac-operator, and osac-aap's execution
   # environment from this PR's current code and installs them together in
--- a/.github/workflows/e2e-caas-full-install.yml
+++ b/.github/workflows/e2e-caas-full-install.yml
@@ -51,10 +51,10 @@ jobs:
           filters: |
             code:
               - '**'
-              - '!OWNERS'
-              - '!LICENSE'
-              - '!*.md'
-              - '!docs/**'
+              - '!**/OWNERS'
+              - '!**/LICENSE'
+              - '!**/*.md'
+              - '!**/docs/**'
 
   # Builds both fulfillment-service and osac-operator from this PR's current
   # code and installs them together in one cluster -- not one "component
--- a/.github/workflows/e2e-vmaas-full-install.yml
+++ b/.github/workflows/e2e-vmaas-full-install.yml
@@ -51,10 +51,10 @@ jobs:
           filters: |
             code:
               - '**'
-              - '!OWNERS'
-              - '!LICENSE'
-              - '!*.md'
-              - '!docs/**'
+              - '!**/OWNERS'
+              - '!**/LICENSE'
+              - '!**/*.md'
+              - '!**/docs/**'
 
   # Builds fulfillment-service, osac-operator, and osac-aap's execution
   # environment from this PR's current code and installs them together in


eliorerz:
Reviewed from scratch — pulled the branch fresh, confirmed the tree independently rather than trusting the PR's claims, and verified the glob semantics empirically against the actual matching library `dorny/paths-filter` uses (not reasoned about from memory).

**(1) Confirmed exact before/after patterns in all 3 files.** Current `main` has `'!OWNERS'`, `'!LICENSE'`, `'!*.md'`, `'!docs/**'` in all three — unanchored, non-`**`-prefixed. The PR's branch changes all three to `'!**/OWNERS'`, `'!**/LICENSE'`, `'!**/*.md'`, `'!**/docs/**'`. Matches the claim exactly.

**(2) Independently confirmed OWNERS/LICENSE exist at all three claimed depths — ran my own `find`, not trusted:**
```
$ find . -name ".git" -prune -o \( -name "OWNERS" -o -name "LICENSE" \) -print
./LICENSE                                                              <- root
./OWNERS                                                               <- root
./fulfillment-service/OWNERS                                          <- per-component
./osac-operator/OWNERS
./osac-aap/OWNERS
./bare-metal-fulfillment-operator/OWNERS
./bare-metal-fulfillment-operator/LICENSE
./osac-aap/collections/ansible_collections/osac/templates/LICENSE     <- vendored, nested
./osac-aap/vendor/ansible_collections/ansible/hub/LICENSE
./osac-aap/vendor/ansible_collections/ansible/utils/LICENSE
./osac-aap/vendor/ansible_collections/ansible_network/network_runner/LICENSE
./osac-aap/vendor/ansible_collections/infra/aap_configuration/LICENSE
./osac-aap/vendor/ansible_collections/kubernetes/core/LICENSE
./osac-aap/vendor/ansible_collections/nvidia/bare_metal/LICENSE
```
Matches the PR's claim precisely, down to the detail that only `bare-metal-fulfillment-operator` has both `OWNERS` *and* `LICENSE` at the component level (the other three components only kept `OWNERS`, consistent with earlier root-file consolidation work that already removed their per-component `LICENSE` files).

**(3) Glob semantics verified empirically against the real library, both before and after — not assumed:**

`dorny/paths-filter`'s `package.json` pins `picomatch@^2.3.1`. Installed that exact version and tested the actual patterns:
```
OLD patterns:
OWNERS      vs OWNERS                                    -> true   (root: matches)
OWNERS      vs fulfillment-service/OWNERS                -> false  (nested: MISSES -- the bug)
docs/**     vs fulfillment-service/docs/AUTH.md           -> false  (nested: MISSES -- the bug)

NEW patterns:
**/OWNERS   vs OWNERS                                     -> true   (root: still matches, not regressed)
**/OWNERS   vs fulfillment-service/OWNERS                 -> true   (nested: now matches)
**/OWNERS   vs osac-aap/vendor/.../ansible/hub/OWNERS      -> true   (deeply nested: matches)
**/docs/**  vs fulfillment-service/docs/AUTH.md            -> true   (matches)
```
This conclusively confirms both halves of the claim: the old patterns really do only match the root case (proving the bug is real, not theoretical), and the new `**/`-prefixed patterns correctly cover root, one-level, and deeply-nested cases without losing the root-level exclusion.

**(4) Confirmed pure exclude-list change, nothing else touched.** This PR's branch predates a recent `main` merge (PR #52), so a naive `git diff main <branch>` shows unrelated noise; isolated the actual change via `git diff $(git merge-base main <branch>) <branch>` and confirmed it's byte-identical to `gh pr diff`'s output (which already computes against the correct merge-base) — exactly 4 lines swapped per file, 3 files, nothing else.

**(5) Noted as intentionally deferred, not a gap in this review.** The PR is explicit that a live docs-only "after" proof isn't included here, and explains why it can't be cleanly isolated within this same PR (its own diff is workflow-YAML — real code — so the filter's `code` output would be `true` regardless of any docs file also touched). Confirmed this reasoning against the PR's own live CI: `e2e-bmaas-full-install / e2e` already passed and `e2e-vmaas`/`e2e-caas` are running, exactly as expected for a PR whose diff is workflow content rather than docs-only. The PR's plan (a follow-up docs-only test PR against fixed `main`, closed without merging once confirmed) is the right way to close the loop.

Also confirmed independently: all 3 files parse as valid YAML, no merge conflicts against current `main`.

**Verdict: confirmed correct on all five points**, with the glob-matching claim in particular verified against the actual library rather than assumed. Good to merge.

PR #55, opened 2026-07-31
OSAC-3519: Add bare-metal-fulfillment-operator to e2e-vmaas/e2e-caas components arrays

eliorerz:
## Summary
Final slice of [OSAC-3519](https://redhat.atlassian.net/browse/OSAC-3519) -- extends PR #51's bmaas fix (merged) to the other two e2e flows. Same change, same imageKey, same containerfile path:

- `e2e-vmaas-full-install.yml`: added bare-metal-fulfillment-operator (BMF) as a 4th `components` entry (alongside fulfillment-service, osac-operator, osac-aap).
- `e2e-caas-full-install.yml`: added BMF as a 3rd `components` entry. Note this file's own components array only has 2 entries today (no osac-aap) -- that's a pre-existing scope decision for this flow unrelated to BMF, left untouched.

Both use `containerfile: "bare-metal-fulfillment-operator/Containerfile"`, `imageKey: "bmf.image.repository"` (no `buildType`, same as fulfillment-service/osac-operator's entries), matching what was verified end-to-end in PR #51.

## Important: this is currently a no-op for both of these flows specifically
Re-confirmed against osac-installer's actual current values files before writing anything:
- `values/vmaas-ci/values.yaml`: `bmf.enabled: false`
- `values/caas-ci/values.yaml`: `bmf.enabled: false`

BMF's chart isn't deployed at all in either flow (unlike `values/bmaas-ci/values.yaml`, where it's `true`), so the `bmf.image.repository` override has nothing to apply to -- no running BMF pod exists in these clusters for the freshly-built image to reach. This is **not a bug in this change**, just a real, currently-inert side effect worth being explicit about rather than silent.

Added anyway, for the same reason the other 3 components are built from source in every flow regardless of what each flow exercises: consistency (all 4 mono-repo components build from the PR everywhere), and it costs nothing beyond one extra image build per run. If vmaas/caas ever need BMF enabled, this wiring is already in place and will start actually mattering with no further change needed here.

## Validation
YAML parse + `yamllint -c .yamllint.yaml --strict` clean on both files. Also parsed the embedded `components` JSON strings directly to confirm 4 entries (vmaas) / 3 entries (caas) with the expected keys.

## Verification plan
Will trigger real e2e-vmaas and e2e-caas runs and confirm from the logs that BMF builds from source (same evidence standard as PR #51 -- component build log entry, submodule replacement, image load), even though the override itself won't be exercised by a running BMF pod in either flow. Will report back with run links before this is ready to merge.

Not merging this myself. Once this merges, [OSAC-3519](https://redhat.atlassian.net/browse/OSAC-3519) is fully done (all 3 e2e flows) and I'll close the ticket.

## Summary by CodeRabbit

* **Chores**
  * Updated end-to-end installation workflows to build the bare-metal fulfillment operator alongside existing components.
  * Added the operator’s image configuration to both full-install workflows.
  * Documented that operator deployment remains disabled in the CAAS workflow.

### DIFF
--- a/.github/workflows/e2e-caas-full-install.yml
+++ b/.github/workflows/e2e-caas-full-install.yml
@@ -56,13 +56,21 @@ jobs:
               - '!**/*.md'
               - '!**/docs/**'
 
-  # Builds both fulfillment-service and osac-operator from this PR's current
-  # code and installs them together in one cluster -- not one "component
-  # under test" plus a stale pinned image for the other. Runs for any change
-  # touching either component (see `changes` above) rather than being
-  # path-scoped per component; splitting this by component to avoid running
-  # both suites when only one changed is deferred to the future full
-  # path-based CI matrix, once its skip-logic pattern is verified.
+  # Builds fulfillment-service, osac-operator, and bare-metal-fulfillment-operator
+  # from this PR's current code and installs them together in one cluster --
+  # not one "component under test" plus a stale pinned image for the others.
+  # Runs for any change touching any component (see `changes` above) rather
+  # than being path-scoped per component; splitting this by component to
+  # avoid running both suites when only one changed is deferred to the
+  # future full path-based CI matrix, once its skip-logic pattern is
+  # verified. (osac-aap is not built here -- that's a pre-existing scope
+  # decision for this flow, unrelated to BMF, not changed by this entry.)
+  #
+  # bare-metal-fulfillment-operator's image override (bmf.image.repository)
+  # is currently a no-op in this flow specifically: values/caas-ci/values.yaml
+  # sets bmf.enabled: false, so the chart isn't deployed here at all. Still
+  # built from source for consistency with the other components -- revisit
+  # if caas ever needs BMF enabled.
   e2e-caas-full-install:
     needs: changes
     if: needs.changes.outputs.should-run == 'true'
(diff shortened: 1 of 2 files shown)


eliorerz:
## Verification results

**e2e-caas**: https://github.com/osac-project/osac/actions/runs/30673628729
- First attempt failed at "Setup CaaS agent" -- `ERROR: Failed to create host type (HTTP 503)`, an Envoy upstream connection failure ("Connection refused") calling fulfillment-service, well after the component build step and completely unrelated to BMF (which isn't even deployed in this flow). Reran the job (`gh run rerun --failed`): passed cleanly end-to-end, confirming it was a transient infrastructure flake, not caused by this change.
- BMF confirmed built from source: `Build component[2]: eliorerz/osac@osac-3519-bmf-vmaas-caas (bare-metal-fulfillment-operator/Containerfile)`, real podman build, submodule correctly replaced.
- Final result: 2 passed, all jobs green.

**e2e-vmaas**: https://github.com/osac-project/osac/actions/runs/30673628790
- First attempt failed at "Run E2E tests" -- `test_compute_instance_api_fields` timed out waiting for a KubeVirt VM to transition Running -> Stopped (300s budget). Unrelated to BMF (also not deployed in this flow) and unrelated to osac-operator/fulfillment-service's images, which built fine -- the other 59 tests in the same run, including several other compute-instance lifecycle tests exercising the same controllers, all passed. Reran the job: passed cleanly, confirming it was a timing flake.
- BMF confirmed built from source: `Build component[3]: eliorerz/osac@osac-3519-bmf-vmaas-caas (bare-metal-fulfillment-operator/Containerfile)`, real podman build, submodule correctly replaced.
- Final result: 60 passed, all jobs green.

Both flows confirm the same evidence standard as PR #51 (bmaas): BMF genuinely builds from this PR's source and gets substituted into the installer, even though `bmf.enabled: false` in both `values/vmaas-ci/values.yaml` and `values/caas-ci/values.yaml` means the built image currently has no running pod to reach (as called out in the PR description).

Ready for review. Not merging this myself.

eliorerz:
Update on the two flaky-looking failures mentioned above: dug further and both turned out to be known, already-tracked issues, not generic flakiness:

- The e2e-caas failure is [OSAC-3505](https://redhat.atlassian.net/browse/OSAC-3505) (closed, fix merged in osac-test-infra) recurring here specifically because this repo's e2e caller workflows pin a stale osac-test-infra SHA that predates that fix by 13 commits. Filed [OSAC-3521](https://redhat.atlassian.net/browse/OSAC-3521) to track bumping the pin -- worth picking up separately since it'll keep causing recurrences of any future osac-test-infra orchestration-level fix, not just this one.
- The e2e-vmaas failure's symptom closely matches [OSAC-3403](https://redhat.atlassian.net/browse/OSAC-3403) (open). Added supporting evidence there.

Neither is caused by this PR -- BMF's build/substitution succeeded in both runs before either unrelated failure occurred -- but wanted to correct the record rather than leave "just a flake" as the final word.

PR #62, opened 2026-08-01
Remove duplicate bare-metal-fulfillment-operator/LICENSE

eliorerz:
## Summary

Follow-up to the earlier LICENSE consolidation (fulfillment-service/LICENSE and osac-operator/LICENSE removed, root LICENSE fixed to say "Copyright OSAC Project Contributors"). `bare-metal-fulfillment-operator/LICENSE` was missed since that component merged in via PR #48 around the same time.

Confirmed byte-identical to root `LICENSE` (`diff LICENSE bare-metal-fulfillment-operator/LICENSE` — no differences) before removing.

## Repo-wide check

Confirmed no other component-level project LICENSE duplicates remain:

```
./LICENSE                                                          <- root, keep
./bare-metal-fulfillment-operator/LICENSE                           <- removed by this PR
./osac-aap/collections/ansible_collections/osac/templates/LICENSE   <- vendored 3rd-party Ansible collection, not ours
./osac-aap/vendor/ansible_collections/*/LICENSE (5 files)           <- vendored 3rd-party Ansible collections, not ours
./osac-aap/collections/ansible_collections/netris/controller/roles/license  <- vendored 3rd-party role, not ours
```

The `osac-aap/vendor/` and `osac-aap/collections/` files are third-party vendored Ansible collections' own licenses (kubernetes/core, ansible/hub, etc.) — required to stay for license compliance, not duplicates of this project's own LICENSE.


eliorerz:
Reviewed from scratch. The fix itself is correct and minimal — one thing worth flagging that's changed on `main` in the last few minutes, not a defect in this PR.

**Confirmed byte-identical**: `diff LICENSE bare-metal-fulfillment-operator/LICENSE` — no output, genuinely identical to root LICENSE.

**Confirmed pure single-file deletion.** This branch predates a very recent `main` merge (osac-installer, PR #61 -- see below), so a naive `git diff main <branch>` shows ~23k lines of bogus "removed" content; isolated the real change via `git diff $(git merge-base main <branch>) <branch>` and confirmed it's exactly 201 lines removed, one file, matching `gh pr diff`'s output precisely.

**Confirmed no problematic leftover references.** `build-bmf-image.yaml` still has a `paths: ['!bare-metal-fulfillment-operator/LICENSE']` negation entry, but this is a harmless no-op once the file's gone (a `paths:` exclude for a nonexistent path just never matches, doesn't error) — and it's consistent with identical dangling entries already present for fulfillment-service/osac-operator/osac-aap's own already-removed LICENSE files across 6 other workflow files. Not something this PR introduced or needs to clean up.

**One thing worth flagging, timing-related, not a mistake in your check:** `osac-installer` merged into this repo via PR #61 at 03:20:12Z — about 6.5 minutes *after* this PR was opened (03:13:44Z). Your repo-wide duplicate check was accurate at the time you ran it, but `osac-installer/LICENSE` now also exists on `main` and is **also byte-identical** to root LICENSE:
```
$ diff LICENSE osac-installer/LICENSE
(no output -- identical)
```
So merging this PR as-is is completely safe and does what it says, but "no other component-level LICENSE duplicates remain" is no longer true the moment #61 landed. Worth a fast immediate follow-up (or folding into this PR before merge, your call) to also remove `osac-installer/LICENSE` while you're here, rather than letting this exact same investigation need repeating in a few days for a second cleanup PR.

No merge conflicts against current `main`.

**Verdict: correct and safe to merge as-is for what it claims to do (BMF's duplicate).** Flagging the newly-surfaced `osac-installer/LICENSE` duplicate as a fast-follow, caught only because of the ~6 minute race between this PR and #61 -- not something your own verification missed.

eliorerz:
Superseded by #64, which does this same LICENSE removal as part of a broader bare-metal-fulfillment-operator root-file consolidation ([OSAC-3525](https://redhat.atlassian.net/browse/OSAC-3525)). Closing to avoid duplicate/conflicting diffs — no need to merge this one separately.

PR #67, opened 2026-08-01
OSAC-3523: Deploy PR's own osac-installer for mono-repo-resident components

eliorerz:
## Summary

Passes `installer-repo`/`installer-ref` (pointing at this PR's own fork/ref) through to the e2e reusable workflows in `e2e-{vmaas,bmaas,caas}-full-install.yml`, mirroring the existing `components[].repo`/`.ref` pattern. Without this, a PR touching fulfillment-service, osac-operator, osac-aap, or bare-metal-fulfillment-operator would deploy against the reusable workflow's default standalone `osac-project/osac-installer` clone, which no longer has the submodules [OSAC-3524](https://redhat.atlassian.net/browse/OSAC-3524) removed and can't resolve the `file://../../../<name>/charts/...` sibling paths that now live inside this same mono-repo checkout.

Also deletes the two nested workflow files under `osac-installer/.github/workflows/` (`e2e-vmaas-full-install.yml`, `e2e-bmaas-full-install.yml`) left over from when osac-installer was a standalone repo. GitHub Actions only discovers workflows under the repo root's own `.github/workflows/`, so these were already inert duplicates of the root callers.

Paired with osac-project/osac-test-infra#291, which makes the reusable workflows' installer-clone/splice logic mono-repo-aware.

## Test plan

- [x] YAML parses for all 3 modified workflow files
- [x] `pre-commit run` (yamllint --strict + standard hooks) passes
- [x] Real e2e-bmaas run against a fork branch of osac-project/osac-test-infra#291 confirms the mono-repo installer path works end to end (see #68)

## Summary by CodeRabbit

* **Tests**
  * End-to-end installation checks now test the installer code from the pull request’s source branch when applicable.
  * Updated BMaaS, CaaS, and VMaaS workflow configuration to select the appropriate installer revision.
  * Removed redundant BMaaS and VMaaS installation workflows from the installer repository.


eliorerz:
Reviewed together with osac-test-infra#291 for consistency (per the paired-PR context). Verified from scratch: pulled both branches, diffed against merge-base, and independently ran yamllint/shellcheck/`python -c "import yaml"` rather than trusting the PR's own test-plan checkboxes.

**Verified correct:**
- `installer-repo`/`installer-ref` added to vmaas/bmaas mirror the exact existing `components[].repo`/`.ref` expression, so for a PR to osac-project/osac they resolve to the same repo+ref as the component images being built -- confirmed `e2e-caas-full-install.yml` already had these two inputs set before this PR (the diff there is comment-only), so vmaas/bmaas were genuinely the only two missing this, not a partial fix.
- Deleting the two nested `osac-installer/.github/workflows/e2e-{v,b}maas-full-install.yml` callers is safe: GitHub only discovers workflows under a repo's own root `.github/workflows/`, so these were already 100% inert (confirmed no `e2e-caas-full-install.yml` ever existed there either, consistent with only 2 files being deleted). Grepped the rest of the repo for references to these two file paths specifically -- none, other than one stale doc table (see below).
- yamllint, and `python3 -c "import yaml"` all clean on the 3 modified workflow files (one pre-existing, unrelated `truthy` warning on the `on:` key, not touched by this diff).

**One small finding, trivially fixable inline:** `osac-installer/AGENTS.md`'s "CI Workflows" table (lines 161-162) still lists `e2e-bmaas-full-install.yml` and `e2e-vmaas-full-install.yml` as files that exist under `osac-installer/.github/workflows/` -- this PR's own deletions just made those two rows factually wrong. Worth a 2-line removal in this same PR rather than waiting for the larger `.github/` consolidation (that consolidation covers the other 14 files' inertness, a much bigger job -- this is just "the table now describes files that don't exist").

**Cross-PR consistency:** confirmed the `installer-repo`/`installer-ref` values this PR threads through are exactly what #291's new `INSTALLER_CLONE_DIR`/directory-presence-detection logic expects, and confirmed #291's `repo == INSTALLER_REPO` skip-splice check will correctly match for all 4 mono-repo-resident components' `components[].repo` values as set here.

Full detailed review of the underlying clone/splice logic posted on osac-test-infra#291. One item there (the known CodeRabbit ref/commit-verification gap) is still pending the author's in-progress push -- I'm holding final sign-off on this pair until that lands, but everything else checks out.

PR #69, opened 2026-08-01
OSAC-3528: Convert osac-ui from git submodule to OCI chart + image reference

eliorerz:
## Summary

Converts osac-ui from a git submodule in osac-installer to a plain OCI Helm chart dependency + image-tag reference, following OSAC-3524's submodule-removal work for the mono-repo-merged components.

osac-ui already publishes real, versioned chart + image artifacts on every tag via its own `publish-image.yaml`/`publish-charts.yaml` (structurally identical to every merged component's own release pipeline). The submodule bought osac-installer nothing that a published OCI artifact reference doesn't already buy it. Unlike the 4 components merged under OSAC-3524, osac-ui has no full-install e2e workflow of its own and is never referenced in any `e2e-*-full-install.yml` `components[]` array -- the PR-source-substitution justification that legitimately keeps those 4 on a local-source pattern (tracked separately in [OSAC-3523](https://redhat.atlassian.net/browse/OSAC-3523)) doesn't apply here in practice.

## Sanity check performed before implementing (per the ticket's explicit ask)

- Latest osac-ui release: `v0.0.5`, published 2026-07-17 (94 commits behind current `main` -- flagging the drift, not fixing osac-ui's own release cadence here).
- **`ghcr.io/osac-project/osac-ui:v0.0.5`**: confirmed via a real registry token + manifest fetch (`HTTP 200`), not just a Packages API listing.
- **`ghcr.io/osac-project/charts/osac-ui:0.0.5`**: confirmed via a real `helm pull`, producing a genuine 2.5KB tarball with `name: osac-ui`, `version: 0.0.5` inside.
- osac-test-infra's netris-lab dev-refresh role has a `# Disable osac-ui (image not published yet)` switch that's now stale given the above -- flagged, not fixed here (different repo/concern, would need someone who owns that lab flow to confirm it's safe to remove).

## Changes

- **`charts/osac/Chart.yaml`**: `file://../../base/osac-ui/charts/ui` → `oci://ghcr.io/osac-project/charts`, pinned to `version: "0.0.5"`.
- **`values/bmaas-ci` and `vmaas-ci`**: `ghcr.io/osac-project/osac-ui:sha-d74aaaf` → `ghcr.io/osac-project/osac-ui:v0.0.5`. `values/development` keeps its floating `:latest` tag, unchanged.
- **`.gitmodules`**: removed `base/osac-ui` + its gitlink. Only `osac-installer/base/osac-csi-driver` remains a real submodule (hasn't merged into the mono-repo).
- **`scripts/sync-image-tags.sh`**: dropped osac-ui's submodule-status-based tag derivation entirely -- its version is a real release now, bumped deliberately like any other external chart dependency, not auto-synced by this script.
- **`.github/workflows/bump-submodules.yaml`** (still nested, not yet wired into root CI -- separate, already-tracked concern, [OSAC-3526](https://redhat.atlassian.net/browse/OSAC-3526)): removed the "check published images"/"update submodules" steps entirely, not just osac-ui's entry -- it was the last one, so those steps had nothing left to do. Renamed the workflow `Bump submodules and pinned refs` → `Bump pinned refs`, dropped the now-inaccurate "submodule" language from its PR title/body/branch name.
- **osac-test-infra companion PR**: [osac-test-infra#292](https://github.com/osac-project/osac-test-infra/pull/292) drops the now-dead `osac-ui` entry from `replace-installer-submodule.sh`'s `SUBMODULE_MAP`.

## Verification

- `helm dependency build charts/osac/` succeeds -- log shows a genuine OCI pull: `Downloading osac-ui from repo oci://ghcr.io/osac-project/charts` / `Pulled: ghcr.io/osac-project/charts/osac-ui:0.0.5`.
- `helm template` succeeds against **every** values file (`development`, `vmaas-ci`, `bmaas-ci`, `caas-ci`).
- Confirmed the rendered UI deployment (`vmaas-ci` values) uses `image: ghcr.io/osac-project/osac-ui:v0.0.5` as expected.
- `helm lint charts/osac/` reports the same pre-existing, unrelated schema error as OSAC-3524's PR (hostname `minLength`) -- not introduced here.
- `actionlint` and `pre-commit run` pass on all changed files.

Part of OSAC-1732. Implements [OSAC-3528](https://redhat.atlassian.net/browse/OSAC-3528).

## Summary by CodeRabbit

* **Enhancements**
  * Updated the UI deployment to use the released `v0.0.5` image and Helm chart.
  * Improved release automation to manage pinned workflow references more reliably.
  * Aligned component image versioning with shared repository releases.

* **Maintenance**
  * Removed the UI’s source submodule integration in favor of the published chart and image.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "osac-installer/base/osac-ui"]
-	path = osac-installer/base/osac-ui
-	url = https://github.com/osac-project/osac-ui.git
 [submodule "osac-installer/base/osac-csi-driver"]
 	path = osac-installer/base/osac-csi-driver
 	url = https://github.com/osac-project/osac-csi-driver.git
--- a/osac-installer/.github/workflows/bump-submodules.yaml
+++ b/osac-installer/.github/workflows/bump-submodules.yaml
@@ -1,4 +1,4 @@
-name: Bump submodules and pinned refs
+name: Bump pinned refs
 
 on:
   schedule:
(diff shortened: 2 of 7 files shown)


eliorerz:
Good catch — fixed. `nightly-build.yaml`'s `COMPONENTS` list (its single source of truth for the submodule-bump/republish/image-tag loops throughout that workflow) still had `osac-ui:osac-ui:base/osac-ui:inline:base/osac-ui/charts/ui`, which would have broken once `base/osac-ui` stopped existing as a submodule.

Dropped the `osac-ui` line entirely (confirmed it's the *only* real reference to osac-ui in that file — every downstream loop iterates over `COMPONENTS`, so removing it there is sufficient, no other step needed touching). It's the right fix, not just a workaround: osac-ui already publishes its own chart/image independently via its own `publish-image.yaml`/`publish-charts.yaml`, so this nightly pipeline republishing it under its own versioning was redundant even before this PR. The other 4 `COMPONENTS` entries are untouched. Validated with `actionlint`/`pre-commit`, amended into the same commit and pushed.

eliorerz:
Follow-up fold-in: found (per a second look) that \`nightly-build.yaml\`'s \`COMPONENTS\`/\`CRD_CHARTS\` lists still had the remaining 4 entries (osac-operator, fulfillment-service, osac-aap, bare-metal-fulfillment-operator) referencing \`base/<name>\` submodule paths removed by OSAC-3524. \`bump-submodules.yaml\` already got this cleanup in that PR; this file was the one leftover inconsistency.

Important distinction from the osac-ui fix above: these 4 aren't gone because they publish independently -- they're gone because they're mono-repo-resident directories with no submodule at all, sharing this repo's own commit history with osac-installer itself. Updated the comment to reflect that correctly rather than reusing the osac-ui reasoning.

Removed all 4 \`COMPONENTS\` entries and both \`CRD_CHARTS\` entries. Both env vars are now empty strings -- confirmed valid YAML (\`>-\` with no content lines parses to \`''\`) and that an empty \`for entry in \${COMPONENTS}\` loop safely iterates zero times, so nothing downstream breaks. Not live-breaking today (this workflow is still nested, not wired into root CI per [OSAC-3526](https://redhat.atlassian.net/browse/OSAC-3526)), but would have broken immediately once that consolidation lands if left stale.

Amended into the same commit, pushed.

PR #72, opened 2026-08-01
OSAC-3513: Repoint osac-aap AAP git-URI default and playbook paths at the mono-repo

eliorerz:
## Summary

The AAP Controller Project's default git-URI (`aap_project_git_uri` in `config.yml`) and osac-installer's CI value overrides still pointed at the standalone `osac-project/osac-aap` repo. That repo is now branch-locked following the mono-repo migration, so runtime playbook cloning against it is a functional risk, not just a stale reference: it can no longer receive updates, so any AAP-side project sync would drift from what actually ships.

## Design decision

AWX/Controller `Project` resources (`scm_type: git`) clone the **entire** repo at the configured `scm_url`/`scm_branch` — there's no native subdirectory-scoping mechanism (confirmed by reading `configure_project_scm.yml`). Job Template `playbook:` paths are simple paths relative to that checkout root.

Given that, three options were considered:
- **Option A (implemented):** Repoint the git-URI default at the mono-repo (`osac-project/osac`) and prefix every `playbook:` path in `controller.yml` with `osac-aap/`, since that's where the playbooks physically live post-migration. Lowest risk, consistent with how every other mono-repo-consolidated component already works.
- **Option B (rejected):** Unlock and keep mirroring content into the standalone `osac-aap` repo. Reintroduces the exact duplication the mono-repo migration was meant to eliminate.
- **Option C (rejected):** Switch to `scm_type: archive` with `aap_project_archive_uri` (an existing but unused code path) to fetch a subdirectory-scoped tarball. Requires new, untested infrastructure for no real benefit over Option A.

## Changes

- `config_as_code/playbooks/vars/config.yml`: default `aap_project_git_uri` → `https://github.com/osac-project/osac.git`
- `config_as_code/roles/aap/vars/controller.yml`: all 31 `playbook:` paths prefixed with `osac-aap/`
- `osac-installer/values/{bmaas,caas,vmaas}-ci/values.yaml`: `configAsCode.projectGitUri`/`projectGitBranch` overrides repointed the same way (these override the default via env var precedence, so fixing only the default would not have fixed live CI deployments)
- 7 `galaxy.yml` files + 4 docs files: stale `osac-project/osac-aap` repository/issue/documentation links updated to `osac-project/osac`

**Not changed (intentionally):** `ghcr.io/osac-project/osac-aap:*` image references — these name a published container image artifact, not a git clone source, and remain valid regardless of the source repo consolidation.

## Test plan

- [x] YAML syntax validated (`yaml.safe_load`) on all changed YAML/galaxy files
- [x] `yamllint -c .yamllint.yaml` clean on all changed YAML files
- [x] `ansible-lint` clean on the two changed Ansible files (0 failures; 1 pre-existing ignored warning, unrelated)
- [x] Confirmed via `find` that playbook files physically live at `osac-aap/<name>.yml` in the mono-repo, matching the new prefixed paths
- [x] Confirmed no live AAP instance available to test project-sync end-to-end — this PR is statically validated only; behavior should be spot-checked against a real Controller instance before/after rollout if possible

## Summary by CodeRabbit

* **Configuration**
  * Updated repository and issue-tracker links to use the main `osac` repository.
  * Updated project sources to use the `main` branch instead of a fixed commit.
  * Corrected automation paths to include the required repository prefix.

* **New Features**
  * Added synchronization checks to ensure required playbooks are discovered before job templates are created.
  * Added clear failure reporting when playbook discovery does not complete within five minutes.

* **Documentation**
  * Updated installation instructions, support links, and repository references for the consolidated project structure.


eliorerz:
Reviewed from scratch (pulled the branch, diffed against merge-base, independently verified every playbook path and both CodeRabbit findings rather than trusting either the PR description or CodeRabbit's own resolution annotations).

**Verified correct:**
- All 31 `playbook:` entries in `controller.yml` are consistently prefixed with `osac-aap/`, and every single one resolves to a real file at that path (checked programmatically, not sampled).
- `aap_project_git_uri`'s default in `config.yml` and all 3 CI `values.yaml` overrides (bmaas/caas/vmaas) are consistently repointed to `https://github.com/osac-project/osac` -- no inconsistency between the default and the overrides.
- Swept the whole repo for remaining `osac-project/osac-aap` references beyond what this PR touched: everything left is either a `ghcr.io/osac-project/osac-aap:*` image reference (correctly out of scope per the PR's own stated rationale -- these name a published artifact, not a git source) or historical PR links in `osac-operator/docs/vmaas-dedicated-cluster/README.md` (`.../osac-aap/pull/209`, `#213`) that still correctly resolve since the old repo is locked, not deleted -- retargeting those wouldn't even make sense, they document discussion that only ever happened in the old repo.
- CodeRabbit's first finding (project source vs. `osac-aap/`-prefixed playbook paths) is genuinely resolved by this same commit, confirmed independently -- its "✅ Addressed" annotation checks out.

**CodeRabbit's second finding is real and still open -- confirming and adding the concrete fix:** `projectGitBranch` changed from a pinned commit SHA (`8a48092a84ca9eb0f9be2fdf9796dce58366c0e7`, matching `eeImage`'s `sha-8a48092` tag exactly) to the literal string `"main"`, in all 3 CI values files. That's a real regression, not just a style nit: `eeImage`/`bootstrap.image` stay pinned to the old SHA while the config-as-code project source now tracks a moving branch, so the two drift apart on every future `main` commit.

I also checked whether the old SHA could simply have been kept (just swapping the URI) -- it can't: I confirmed `8a48092a84ca9eb0f9be2fdf9796dce58366c0e7` is still a reachable commit object in the monorepo's history (it's the exact commit the subtree-merge pulled in), but checking it out directly would give a tree *without* the `osac-aap/` prefix, breaking the very playbook paths this PR just fixed.

The actually-correct pin is the monorepo's own subtree-merge commit: `dc4352e3d6a294589dc4e54d6261bf1488ebd8b6` ("Add 'osac-aap/' from commit '8a48092...'"). I verified its `osac-aap/` tree object (`0c776b1f68634fe1bdcc1299e152a2dd1698bd2b`) is byte-identical to the old standalone commit's root tree -- so pinning `projectGitBranch` to that SHA instead of `"main"` gives exactly the same content as today's `eeImage`, *with* the correct `osac-aap/` prefix, fully restoring the pinning guarantee that existed before this PR. Worth fixing across all 3 values files before merge.

Everything else checks out. Holding this PR on the one open item above.

eliorerz:
Rebased onto latest main and pushed an update. Status on the three asks:

## CodeRabbit findings

1. **controller.yml SCM wiring**: added a comment clarifying that `scm_url`/`scm_branch` are set at runtime by `configure_project_scm.yml` (not in this file), and why every `playbook:` path carries the `osac-aap/` prefix. The underlying mechanism was already correct -- traced `configure_project_scm.yml`'s task ordering in `main.yml` to confirm it runs (and merges scm config into `controller_projects`) *before* the dispatch role creates the actual AWX objects. This was a readability gap, not a functional bug.

2. **projectGitUri/projectGitBranch pinning consistency**: superseded by PR #78 (OSAC-3367, merged since this PR opened). That PR dropped all image-tag pinning for mono-repo-resident components in these same CI values files (`bootstrap.image`/`eeImage` are now `latest`, not a pinned SHA). The inconsistency CodeRabbit flagged -- pinned image vs. unpinned git branch -- no longer exists post-rebase: both are now consistently unpinned, matching PR #78's established policy. No change needed.

## e2e-bmaas hang -- root cause investigated, **not yet fixed**

Pulled the real AAP bootstrap pod logs and AWX project-update records from a failed run's artifact bundle (not just the GitHub Actions log). Findings:

- Ruled out the original suspicion (mono-repo clone too slow): all 40 AWX project syncs across the run's 8 pod-restart attempts completed successfully in 6-9s each.
- The actual, 100%-reproducible failure happens immediately after every successful sync: `Unable to create job_template osac-create-hosted-cluster: {'playbook': ['Playbook not found for project.']}` -- on every single attempt, including the very first.
- Confirmed via side-by-side comparison against a real successful bmaas e2e run's own AAP job logs (pre-cutover, old unprefixed scm_url) that this is new and specific to this PR, not a pre-existing flake.

I initially proposed a fix (delete-and-recreate the AWX project when its scm_url changes, based on a suspected AWX caching bug) but couldn't corroborate that specific mechanism against real AWX documentation or a verifiable issue report, and it would be a destructive, production-affecting change (deletes job history/schedules). Rather than land a fix based on log inference alone, I'm leaving this as an open, documented finding -- see the latest commit message for the full writeup. This needs either a live AWX repro or a maintainer with more AWX-internals context to pick up.

**This PR is not yet ready to merge** -- the e2e-bmaas failure is real and unresolved, just no longer misattributed to clone speed.

eliorerz:
@coderabbitai full review

eliorerz:
Rebased onto current main to resolve merge conflicts (this PR's `projectGitUri`/`projectGitBranch` mono-repo repoint conflicted with OSAC-3367's eeImage pinning change + [OSAC-3545](https://redhat.atlassian.net/browse/OSAC-3545) note landing on the same lines in `values/{bmaas,caas,vmaas}-ci/values.yaml`). Resolution: kept the newer eeImage/OSAC-3545 comment from main, applied this PR's `projectGitUri`/`projectGitBranch` mono-repo values on top -- no semantic conflict, just adjacent-line churn. Diff shape vs. main is unchanged from before the rebase.

No new information from the rebase -- the AWX bootstrap-hang investigation stands exactly as documented in the previous comment/commit: root-caused but not fixed, needs live AWX access or expertise to resolve. Still not ready to merge.

eliorerz:
## Major correction to the root-cause investigation — new decisive evidence

Per the suggestion to use live SSH access to the runner host during an active e2e run (same technique as OSAC-3499's original triage), I queried the live AWX Controller API and inspected the actual deployed bootstrap image directly during a real bmaas run. This overturns my earlier "AWX playbook-indexing lag" theory entirely.

**What I found:**
- The live AWX Project's `scm_url`/`scm_branch` correctly point at the mono-repo (`osac-project/osac`/`main`) -- this comes from Helm *values* (`values/bmaas-ci/values.yaml`), which are read live from this PR's own checkout.
- The live Project's discovered playbook list **does** include `osac-aap/playbook_osac_create_hosted_cluster.yml` (confirmed via `GET /api/controller/v2/projects/{id}/playbooks/`), proving the sync and indexing are both fine.
- Yet job template creation still failed with `Unable to create job_template osac-create-hosted-cluster: {'playbook': ['Playbook not found for project.']}` -- because the actual `playbook:` value being requested, per the deployed bootstrap image's own `controller.yml`, is the **unprefixed** `"playbook_osac_create_hosted_cluster.yml"` -- confirmed by pulling `ghcr.io/osac-project/osac-aap:latest` (the image actually running the Job) and reading its baked-in `controller.yml` directly.
- That image also does not contain my new `wait_for_project_playbooks.yml` task at all.

**Root cause**: the bootstrap Job's container image is the *published* `ghcr.io/osac-project/osac-aap:latest` from main, not a freshly-built image from this PR's own commits. The e2e workflow's `COMPONENT_BUILD_RESULTS` image-override for `aap.bootstrap.image`/`aap.configAsCode.eeImage` isn't actually applying on a fresh Helm install (a gap I'd noted in an earlier investigation without realizing it was the direct cause of this specific failure). So this PR's own `osac-aap/` prefix fix in `controller.yml` has never actually been exercised by e2e -- the old (main) image's old (unprefixed) `controller.yml` is running against the new (mono-repo) project location, and the mismatch is 100% deterministic and permanent, not a timing issue at all.

**Consequence**: my `wait_for_project_playbooks.yml` polling fix (previous commit) is harmless/additive but doesn't address the real problem, and has never actually been deployed/tested. The real blocker lives in osac-test-infra's e2e workflow, not in this repo.

Leaving this PR exactly as-is pending direction on whether to pursue the osac-test-infra fix.

PR #74, opened 2026-08-01
OSAC-3526: Consolidate osac-installer's discoverable CI into root

eliorerz:
## Summary

osac-installer's own CI workflows still live at osac-installer/.github/workflows/, a leftover from before it was merged into this mono-repo (PR #61). GitHub Actions only discovers workflows under the repo root's own .github/workflows/, so none of these currently run on any PR, schedule, or dispatch. This handles 12 of the 15 files (14 workflows + dependabot.yml), grouped by what each actually needs.

**Deleted as already-redundant duplicates:** pre-commit.yaml, ok-to-test-label-cleanup.yml, slash-command.yml, and dependabot.yml are fully covered by root's own versions already. scan-workflow-logs.yml is covered too, but its one meaningful difference (a Nightly Build entry in its workflow_run listener list) was folded into root's copy first, so nothing is lost -- that entry is forward-compatible with the eventual root-level Nightly Build consolidation and inert until then.

**Relocated (no root equivalent):** cleanup-nightly-branches.yaml, mirror-envoy.yaml, secret-scanning.yaml, bump-submodules.yaml. The last two aren't actually osac-installer-specific and were already written repo-root-relative -- moving them closes a real, repo-wide gap: nothing currently runs gitleaks against any PR diff in this repo (despite a root .gitleaks.toml already existing), and nothing bumps the SHA-pinned osac-test-infra action refs already used in root's own scan-workflow-logs.yml (scan-and-purge-logs@..., notify-slack@...).

**Merged into root's existing file (different job shape than a simple matrix add would fix):** check-floating-tags.yaml gained osac-installer as a 4th matrix entry, identical shape to the existing three. helm-lint.yaml and integration-tests.yml each gained a new sibling job, since osac-installer's own versions test the umbrella chart's composition of every other component's chart together -- a different thing from linting/testing any one component's chart in isolation, not a parameterizable variant of the existing jobs.

**Intentionally excluded (3 files, blocked on [OSAC-3367](https://redhat.atlassian.net/browse/OSAC-3367)):** nightly-build.yaml and osac-installer's own publish-charts.yaml (distinct from root's already-existing multi-component publish-charts.yaml) still hardcode per-component logic that [OSAC-3367](https://redhat.atlassian.net/browse/OSAC-3367) is actively redesigning, itself blocked on an open pinning-strategy question.

check-image-tags.yaml was also moved here after testing the relocated version for real surfaced a structural issue: sync-image-tags.sh compares values/*-ci/values.yaml's pinned SHA tags against the mono-repo's own `git rev-parse HEAD`, but a pull_request-triggered run's actions/checkout defaults to a synthetic merge-commit ref -- a different, unpredictable SHA every time, that can never be known in advance and committed into a values file. This check can't pass as a PR gate under the current design regardless of relocation; it's the same open pinning-strategy question [OSAC-3367](https://redhat.atlassian.net/browse/OSAC-3367) is already blocked on, not something this ticket's relocation work can resolve on its own.

Part of OSAC-1732. See [OSAC-3526](https://redhat.atlassian.net/browse/OSAC-3526) for the full per-file breakdown.

## Test plan

- [x] YAML parses for every added/modified file
- [x] `yamllint --strict` passes
- [x] `actionlint` passes (one pre-existing, unrelated `osac-ci` custom-runner-label false positive on an untouched line)
- [x] Full `pre-commit run` passes
- [x] Confirmed real, passing runs on this PR itself: "Reject unmarked floating tags (osac-installer)", "Lint Helm charts (osac-installer)", "Scan PR diff for leaked secrets" all triggered and passed
- [x] Confirmed check-image-tags.yaml's structural issue by testing the relocated version live (see above) and reverted it
- [ ] "Run integration test (osac-installer)" still completing (Kind-cluster job takes longer)


eliorerz:
Reviewed from scratch (pulled the branch, diffed every one of the 15 files' claimed dispositions individually rather than trusting the summary table, and read the full before/after content of the two new/rewritten jobs).

**Verified correct:**
- `scan-workflow-logs.yml`: the claimed "Nightly Build" listener entry really was folded into root's copy (confirmed the exact 4 added lines, with a clear comment explaining why it's listed ahead of the workflow existing at root), and it was genuinely present in osac-installer's old copy first -- nothing invented, nothing lost.
- `check-floating-tags.yaml`: both the trigger `paths:` list and the `matrix.component` array got `osac-installer` added consistently -- not a partial update.
- `helm-lint.yaml`'s new `helm-lint-installer` job: every relocated step that touches repo content correctly got `working-directory: osac-installer` added (this is exactly the risk I flagged in my own earlier osac-installer root-audit -- none of osac-installer's original steps used explicit `working-directory:` since they assumed repo-root cwd; this PR handled that correctly and completely for this file).
- `integration-tests.yml`'s shared `changes` gate job uses an unscoped `'**'` filter (minus OWNERS/LICENSE/md/docs), so the new `osac-installer` job's `if: needs.changes.outputs.should-run` will correctly fire for osac-installer path changes -- no silent no-op risk there.
- `pre-commit.yaml`/`ok-to-test-label-cleanup.yml`/`slash-command.yml`/`dependabot.yml` deletions are justified: root's `pre-commit.yaml` triggers on all `pull_request` events repo-wide (no path scoping) and already covers osac-installer via the pre-commit/yamllint reconciliation that already landed with the original merge (`OSAC-3522`).
- `nightly-build.yaml` and osac-installer's own `publish-charts.yaml` are genuinely untouched, confirmed via diff -- the "intentionally excluded, blocked on OSAC-3367" claim is honest, not glossing over a partial job.

**Finding 1 -- one of the 15 files wasn't actually handled despite being listed as such:** `osac-installer/.github/workflows/check-image-tags.yaml` ("Check image tags match submodules", runs `scripts/sync-image-tags.sh`) is listed under "Relocated (no root equivalent)" in the PR description, but it doesn't appear anywhere in the actual diff -- confirmed it's still sitting untouched at the old path on this PR's branch, and root has no equivalent file. It'll remain exactly as inert after this merges as it was before. This PR handles 12 of the 15 files, not 13.

**Finding 2 -- pre-existing bug, but this PR is what makes it consequential for the first time:** the new `osac-installer` job in `integration-tests.yml` has `continue-on-error: true` on both the \"Deploy umbrella chart\" (`helm install`) and \"Test idempotency\" (`helm upgrade`) steps, with no downstream step checking their outcome and failing the job if either fails. I confirmed this is copied byte-for-byte from osac-installer's own old job, so it's not new -- but the old job only ever ran on `schedule`/`workflow_dispatch`, so this bug never actually mattered (nobody was watching a scheduled job's internals for silent partial success). Now that this PR wires it into `pull_request`-gated CI as a real merge signal, the same bug means this job will report green even if the actual chart install or upgrade fails outright -- only the preceding lint/template-dry-run steps can fail it. Worth adding `id:` to those two steps and a trailing `if: steps.deploy.outcome == 'failure' || steps.upgrade.outcome == 'failure'` gate (or just dropping `continue-on-error` and using `if: always()` on the "Check deployed resources" step, which already exists for diagnostics) before this becomes real CI signal people rely on.

Recommend fixing both before merge -- neither is large, but #1 leaves a stated goal silently unmet and #2 would make a genuine integration test structurally unable to catch the one thing it exists to catch.

PR #75, opened 2026-08-01
OSAC-3530: Merge osac-csi-driver into osac mono-repo

eliorerz:
## Summary

Merges [osac-csi-driver](https://github.com/osac-project/osac-csi-driver) into the mono-repo, following the same pattern established for fulfillment-service, osac-operator, osac-aap, and bare-metal-fulfillment-operator.

- `git subtree add --prefix=osac-csi-driver` (full history preserved, no `--squash`) — verified the merge commit's second parent has all 34 commits from the standalone repo, cross-checked against the live repo's commit count via the GitHub API.
- Standalone repo's `main` branch locked (branch protection API + durable Terraform change in `github-config`, so it survives the next auto-apply).
- Root-file audit: consolidated `.pre-commit-config.yaml`/`.pre-commit-config-ci.yaml`/`.yamllint.yaml`/`dependabot.yml` into root's, dropped the redundant per-component copies (and the stale "Innabox"-era `LICENSE`), added an `osac-csi-driver` matrix entry to `helm-lint.yaml`, merged `OWNERS` and `.gitignore` entries, added a README bullet.
- Scoped `publish-image.yaml`'s tag trigger to `osac-csi-driver/v*` and hardcoded `IMAGE_NAME` up front, to avoid the OSAC-3467/OSAC-3490 image-tag collision that hit the earlier merges. Renamed the workflow to `Build osac-csi-driver image` (its default name would've collided with fulfillment-service's `Container image` workflow, which `publish-charts.yaml` disambiguates on).
- Added the missing `publish-charts.yaml` wiring (guard job + `publish-csi-driver-chart`/`publish-csi-backends-chart`/release jobs), matching the shared per-component pattern.
- Repointed `osac-installer/charts/osac/Chart.yaml`'s `csi-driver`/`csi-backends` dependencies from the old `base/osac-csi-driver` submodule path to the mono-repo-local `osac-csi-driver/charts/{csi-driver,csi-backends}` paths, and removed the now-unused submodule from `.gitmodules`. Verified with a real `helm dependency build` + `helm template` (126 rendered manifests, including the csiDriver/csiBackends resources).
- Added `osac-csi-driver` to the root `go.work` workspace — required, since Go's workspace auto-discovery otherwise breaks `go build`/`go vet`/`golangci-lint` for any module living under a `go.work` root that isn't listed in it. Verified `go build`/`go vet` succeed from within `osac-csi-driver` afterward.
- No open GitHub Issues on the standalone repo to transfer.
- Disabled the standalone repo's `pre-commit` and `Container image` CI workflows post-merge (its branch is also lock-protected).

## Test plan

- [x] `git subtree add` history verified (commit count + merge-commit parentage)
- [x] `pre-commit run` clean on all changed files (yamllint, whitespace/EOF checks, osac-csi-driver golangci-lint with binary pre-downloaded)
- [x] `helm lint`/`helm template` clean for both `charts/csi-driver` and `charts/csi-backends`, standalone and via the umbrella `osac-installer/charts/osac` chart
- [x] `go build ./...` / `go vet ./...` clean from within `osac-csi-driver`, with `osac-operator`/`bare-metal-fulfillment-operator` builds unaffected
- [x] `publish-csi-driver-image.yaml` and `publish-charts.yaml` YAML-validated; workflow-name collision with fulfillment-service's `Container image` avoided
- [ ] Actual tag push (`osac-csi-driver/v0.0.1`-style) to confirm the end-to-end image-build → chart-publish → release chain, once this merges

## Summary by CodeRabbit

* **New Features**
  * Added the OSAC CSI driver for routing storage operations across vendor backends.
  * Added Helm charts for the CSI driver and Trident, VAST, and Pure Storage integrations.
  * Added container image and chart publishing for component releases.
* **Documentation**
  * Added deployment, configuration, build, and security guidance.
* **Tests**
  * Added end-to-end CSI sanity testing with a simulated vendor backend.
* **Chores**
  * Added automated linting, validation, dependency updates, and release safeguards.


eliorerz:
Triaged all 33 CodeRabbit findings on this PR against the actual diff (not just file location) to separate what this merge's own new content introduced from what's pre-existing in osac-csi-driver's carried-over driver/chart/test code (this is a full-history `git subtree add`, so every file under `osac-csi-driver/` shows as "added" relative to `main` regardless of which of the 34 original commits actually wrote it).

**5 findings are caused by this merge's own new content** (the wiring commit's workflow/pre-commit/README changes, not the subtree-added code) -- fixes ready to push to this branch, coordinating with the PR author first since some may already be in progress:
- `publish-charts.yaml`: the new csi-driver tag-placeholder `sed` step had no verification it landed (same class of gap `publish-osac-aap-chart`'s `replace_and_verify` already guards against).
- `publish-csi-driver-image.yaml`: missing `!osac-csi-driver/charts/**` exclusion (chart-only PRs shouldn't trigger a full image rebuild), plus all 4 actions this new workflow uses were unpinned to mutable tags.
- `.pre-commit-config.yaml`: the new `osac-csi-driver-golangci-lint` entry calls the binary directly, which won't exist on a clean checkout.
- `README.md`: the go.work workspace paragraph is missing osac-csi-driver (and, pre-existing, bare-metal-fulfillment-operator).

**The remaining 28 findings are real, but in osac-csi-driver's own pre-existing code** (Helm charts, Containerfile, Go driver logic, sanity tests) that this PR carries over unchanged from the standalone repo's history -- not something a migration PR should silently rewrite out from under the component's actual owners (zszabo-rh, wgordon17, avishayt, rgolangh per OWNERS). Filed for follow-up, not blocking this merge:
- [OSAC-3532](https://redhat.atlassian.net/browse/OSAC-3532) -- Helm chart / Containerfile security-hardening gaps (namespace-wide privileged PSA that should be scoped to the node DaemonSet only, missing resource limits/probes on several containers, overly-broad Trident ClusterRole, world-writable CSI socket, missing TLS for TCP vendor connections, `COPY . ./` + no HEALTHCHECK in the Containerfile, etc.)
- [OSAC-3533](https://redhat.atlassian.net/browse/OSAC-3533) -- driver correctness bugs and test reliability issues, including one **critical** item: `NodeUnstageVolume`/`NodeUnpublishVolume` currently collapse a real routing failure into a false success, which can leave a vendor mount/device attached on the node while Kubernetes believes it's unstaged. Also volume-ID/backend-resolution inconsistencies in `controller.go`/`node.go` that need a design decision (not backward-compatible with existing volumes), plus a few lower-severity code-quality/test-flakiness items.

Neither ticket blocks this PR -- they're pre-existing conditions in code that already exists and runs today in the standalone repo, unaffected by whether this merge lands.

eliorerz:
Follow-up review pass on CodeRabbit's post-push findings (submitted 21:46:53, after my earlier fixes). Verified each against current code rather than applying diffs blindly, per the same discipline as before.

**Fixed:** the prerelease/stable-alias overwrite bug in `publish-csi-driver-image.yaml`. Confirmed real: `RELEASE_MAJOR_MINOR`/`RELEASE_MAJOR` were derived from the prerelease-stripped core version, but their metadata-action entries were gated on `RELEASE_VERSION != ''` rather than their own value -- so a tag like `osac-csi-driver/v1.2.3-rc.1` would still publish/overwrite the `v1.2` and `v1` stable alias tags with a prerelease image. Traced through with test values (`v1.2.3` → aliases set; `v1.2.3-rc.1`/`v2.0.0-beta.1.2` → aliases correctly left empty) before pushing. Now only sets those two env vars when the tag has no prerelease suffix, and gates their tag entries on their own non-emptiness instead of `RELEASE_VERSION`'s.

**Rejected as a false positive:** the `helm-lint-installer` job "missing permissions" finding. Checked the actual file -- `helm-lint.yaml` already has a **workflow-level** `permissions: contents: read` block (line 24) that every job in the file inherits, including `helm-lint-installer`. Neither of the other two jobs in this same file (`helm-crds-sync`, `helm-lint`) has a job-level override either -- they all correctly rely on the same workflow-level default. Adding a redundant job-level block to only this one job would be a no-op change that also breaks consistency with its siblings. CodeRabbit's diff-scoped view of this PR likely didn't see the top-of-file block since it wasn't part of the diff.

**Deferred, not fixed here:** the missing Sigstore/cosign signing on published images/charts. Checked first, per instruction, whether this is unique to csi-driver -- it isn't: grepped every workflow in the repo for `cosign`/`sigstore`/`id-token` and found zero signing usage anywhere, across every component's publish-image/publish-charts jobs. This is already tracked epic-wide at [OSAC-1148](https://redhat.atlassian.net/browse/OSAC-1148) ("Sign all distributed artifacts... with cosign"), which already covers container images, Helm charts, and binaries -- I added a comment there linking this PR as a concrete current instance rather than filing a duplicate or bolting a one-off fix onto csi-driver alone.

Pushed the one real fix. Will keep watching this PR for further CodeRabbit rounds.

PR #77, opened 2026-08-01
OSAC-3484: Scope fulfillment-service's SRPM archive to its own subtree

eliorerz:
## Summary

`fulfillment-service/.copr/srpm.sh` builds the SRPM's source tarball with:

```
git -C "${here}/.." archive --prefix="${name}-${version}/" --output=... HEAD
```

This worked pre-merge because `${here}/..` (fulfillment-service's own root) *was* the repo root, so an unscoped `archive HEAD` correctly meant "the whole (small) repo." Post-merge, `${here}/..` is just a subdirectory of the mono-repo — the same command now archives the **entire mono-repo tree**, so the "fulfillment-service" SRPM's source tarball ends up containing every other component's full source too.

## Fix

- Archives `HEAD:fulfillment-service` (a subtree pathspec) instead of `HEAD`, so fulfillment-service's own directory becomes the archive root — matching what `%build`/`%install` in `osac-cli.spec.in` expect (bare `cmd/osac`, `go.mod`, etc.).
- `fulfillment-service/LICENSE` was itself deleted as a redundant duplicate when its root files were consolidated into the mono-repo's own `LICENSE` ([OSAC-1733](https://redhat.atlassian.net/browse/OSAC-1733)) — so the subtree archive alone is missing the `%license` file the spec requires. The tarball is now assembled in a scratch directory instead, with the root `LICENSE` copied back in before packing.
- Also scopes the version's `git describe` to `fulfillment-service/v*`: the mono-repo's tags aren't namespaced per directory, so an unscoped `git describe --tags` can just as easily resolve against the nearest tag from an unrelated component.

## Test plan

Verified end-to-end in a clean Fedora container (matching the real COPR builder environment):
- [x] `make -f .copr/Makefile srpm` produces a valid SRPM
- [x] Version resolves to fulfillment-service's own tag (`0.0.82`), not another component's
- [x] Tarball contains only fulfillment-service's own tree plus `LICENSE` (2.8M vs. 93M for the whole repo) — no other component directories present
- [x] `rpmbuild --rebuild` on the produced SRPM succeeds end-to-end (`go build ./cmd/osac`, `%license LICENSE`, `%doc README.md` all resolve)
- [x] The resulting `osac` binary runs correctly

## Summary by CodeRabbit

* **Chores**
  * Improved fulfillment-service source archive generation.
  * Version detection now uses fulfillment-service-specific tags.
  * Archives contain only the relevant service files and include the repository license.
  * Added safer staging and error handling during package creation.

### DIFF
--- a/fulfillment-service/.copr/srpm.sh
+++ b/fulfillment-service/.copr/srpm.sh
@@ -12,13 +12,17 @@
 # language governing permissions and limitations under the License.
 #
 
-set -e
+set -eo pipefail
 
 # Name of the project:
 name="osac-cli"
 
 # Directory containing this script:
 here="$(dirname "$(readlink -f "$0")")"
+
+# This component's own directory, and the mono-repo root above it:
+component_dir="$(readlink -f "${here}/..")"
+repo_root="$(readlink -f "${here}/../..")"
 
 # Output directory:
 outdir="${outdir:-$here}"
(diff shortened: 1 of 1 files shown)


eliorerz:
Reviewed from scratch — pulled the branch and actually ran the script end-to-end (up to the \`rpmbuild\` call, which isn't available in this sandbox) rather than just reading the diff.

**Verified correct, empirically:**
- Version computation: \`git describe --tags --always --match 'fulfillment-service/v*'\` on the real repo resolves to \`0.0.82^1.g8f3ecea\` — correctly scoped to fulfillment-service's own tag lineage. I independently hit this exact ambiguity bug during an unrelated investigation earlier (unscoped \`git describe\` on this repo currently resolves to a bare-metal-fulfillment-operator tag, not fulfillment-service's), so I can confirm this fix is necessary, not just defensive.
- Archive scoping: ran the script for real. The produced tarball is 2.91MB (matches the PR's claimed 2.8M), contains \`go.mod\`, \`cmd/osac/\`, \`README.md\`, and \`LICENSE\` at its root, and — checked explicitly — zero occurrences of any other component's directory (osac-operator, osac-aap, bare-metal-fulfillment-operator). The \`HEAD:fulfillment-service\` subtree pathspec correctly flattens the component's own directory to the archive root.
- Confirmed \`fulfillment-service/LICENSE\` genuinely doesn't exist anymore (deleted as a dup per [OSAC-1733](https://redhat.atlassian.net/browse/OSAC-1733)) and root \`LICENSE\` does — so the scratch-directory-plus-copy-back approach is solving a real, current gap, not a hypothetical one.
- \`set -eo pipefail\` upgrade is correct and needed here: the \`git archive | tar -x\` pipeline wouldn't fail loudly under plain \`set -e\` if the \`git archive\` half failed.

No issues found. Good to merge.

PR #83, opened 2026-08-02
OSAC-2156: Add length validation to ClusterVersion image and version fields

sk-ilya:
## Summary

- Add `buf.validate` string length constraints to `ClusterVersionSpec.image` (`min_len: 1`, `max_len: 512`) in the private proto and `ClusterVersionSpec.version` (`min_len: 1`, `max_len: 256`) in both private and public protos
- Add Go-level `max_len` validation for `spec.image` in `validateClusterVersionCreateRequest` (mirrors the existing `validateSemVer` length check for `spec.version`)
- Add test entries for both max-length constraints

The `image` field previously had no `buf.validate` annotations -- only `REQUIRED`/`IMMUTABLE` field-behavior hints and a Go-level empty check. The `version` field had a Go-level max-length check via `validateSemVer` but no proto-level constraint. Both now follow the API convention documented in `docs/API.md` (lines 122-124, 140): constrained fields should carry `buf.validate` rules so the protovalidate interceptor enforces them uniformly.

## Test plan

- [x] Existing "rejects empty image" test still passes
- [x] New "rejects image exceeding max length" test (513 chars) passes
- [x] New "rejects version exceeding max length" test (257 chars) passes
- [x] Full `ginkgo run -r internal` passes (84 suites, 0 failures)
- [x] `buf lint` passes

Assisted-by: Claude <noreply@anthropic.com>

---

Re-opened from osac-project/fulfillment-service#975, which is closed since that repo was archived in favor of this monorepo. No code changes from the original PR other than the `fulfillment-service/` path prefix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to prevent overly long image and version values.
  * Requests now return clear invalid-argument errors when limits are exceeded.
  * Version values must be between 1 and 256 characters; image values must be between 1 and 512 characters.

PR #84, opened 2026-08-02
OSAC-2162: Implement version-based cluster creation

sk-ilya:
Replace release_image with version_name on ClusterSpec so that users specify a cluster version instead of a raw image URL. The server resolves the version to a ClusterVersion object, validates it is usable, and falls back to the system default when no version is specified. The controller then looks up the release image from the resolved ClusterVersion at reconciliation time.

## Summary

* Added `--version` for cluster creation, selecting a ClusterVersion and driving `spec.version_name` (user input -> catalog/template defaults -> system default).
* Updated reconciliation to resolve and apply the ClusterVersion image to the resulting release image, including graceful handling when the version is missing.
* Added validation for `spec_defaults.version_name` and `field_definitions.version_name` (missing/disabled/obsolete now fail).
* Updated CLI help and admin/catalog docs, including "version not found" output guidance.
* Extended unit and integration coverage for version resolution precedence and failure scenarios.

## Test plan

- [x] Full `ginkgo run -r internal` passes (84 suites, 0 failures)
- [x] `buf lint` passes
- [x] `go build ./cmd/fulfillment-service ./cmd/osac` passes

---

Re-opened from osac-project/fulfillment-service#968, which is closed since that repo was archived in favor of this monorepo. No code changes from the original PR other than the `fulfillment-service/` path prefix.

## Summary by CodeRabbit

* **New Features**
  * Select cluster versions by name when creating clusters through the API or CLI.
  * Version selection follows explicit settings, template or catalog defaults, then the system default.
  * Added validation to prevent use of missing, disabled, deleted, or obsolete versions.
  * CLI errors now show the requested version and available alternatives.
* **Documentation**
  * Added guidance for managing cluster versions, lifecycle states, defaults, and version selection.
  * Updated catalog item and cluster examples to use `version_name` instead of `release_image`.
* **Bug Fixes**
  * Improved handling of version lookup failures and incomplete update requests.


fulfillment-service/internal/controllers/cluster/cluster_reconciler_function.go
```diff
@@ -334,6 +366,37 @@ func (t *task) addExplicitFields(spec *osacv1alpha1.ClusterOrderSpec) {
...
+			errClusterVersionNotFound, versionName,
+		)
+	}
+	if response.GetTotal() > 1 {
+		t.r.logger.WarnContext(ctx, "Multiple cluster versions found for name, using first",
+			"version_name", versionName,
+			"count", response.GetTotal(),
+		)
+	}
+	return response.GetItems()[0].GetSpec().GetImage(), nil
```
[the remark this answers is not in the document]
sk-ilya: it cannot be empty

fulfillment-service/proto/public/osac/public/v1/cluster_type.proto
```diff
@@ -146,11 +146,17 @@ message ClusterSpec {
...
-  optional string release_image = 6;
+  // The version is resolved with the following precedence:
+  //
+  //   1. Explicit `spec.version_name` provided by the user.
+  //   2. Template `spec_defaults.version_name` (template path) or catalog item field definition default (catalog-item
+  //      path).
+  //   3. System default (`is_default = true`).
+  //
+  // The referenced ClusterVersion must exist, be enabled, and not be in `OBSOLETE` state.
+  optional string version_name = 6;
```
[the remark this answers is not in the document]
sk-ilya: no need to preserve backward compatibility at this early stage of the project

fulfillment-service/proto/private/osac/private/v1/cluster_template_type.proto
```diff
@@ -58,9 +58,9 @@ message ClusterTemplateSpecDefaults {
...
   optional string ssh_public_key = 2;
 
-  // Default OCP release image URL.
-  optional string release_image = 3;
-
   // Default cluster networking configuration.
   optional ClusterNetwork network = 4;
+
+  // Default ClusterVersion name for clusters created with this template.
+  optional string version_name = 5;
```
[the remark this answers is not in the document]
sk-ilya: no need for backward  compatibility at this state of the project

sk-ilya:
@coderabbitai pause

fulfillment-service/internal/servers/private_clusters_server.go
```diff
@@ -546,22 +582,39 @@ func (s *PrivateClustersServer) validateTemplateImmutability(ctx context.Context
...
+		return err
+	}
+	if !newSpec.HasVersionName() {
+		newSpec.SetVersionName(existing.GetSpec().GetVersionName())
+		return nil
+	}
+	if newSpec.GetVersionName() == existing.GetSpec().GetVersionName() {
+		return nil
+	}
+	return s.ensureClusterVersion(ctx, request.GetObject())
```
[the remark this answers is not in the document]
sk-ilya: recheck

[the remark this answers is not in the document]
sk-ilya: added

fulfillment-service/docs/CATALOG_ITEMS.md
```diff
@@ -40,7 +40,7 @@ to end users via the public API (list, inspect, and use when creating resources)
 
 Catalog items control **both typed spec fields and template parameters**:
 
-- **Typed spec fields** (e.g., `pull_secret`, `ssh_public_key`, `release_image`, `network.pod_cidr`,
+- **Typed spec fields** (e.g., `pull_secret`, `ssh_public_key`, `version_name`, `network.pod_cidr`,
```
empovit, line 43: Does `version_name` correspond to `--name "4-17-0"`? Or to `--version "4.17.0"`? The field name contains both, so it's confusing.

sk-ilya: `version_name` corresponds to `ClusterVersion.metadata.name`, which is effectively the PK. 

The `--version "4.17.0"` is used **only** by the `osac create cluster ` CLI command as a convenience - dots are not allowed in the name field. It accepts both the `ClusterVersion` name or `ClusterVersion.spec.version` and then resolves the name.

fulfillment-service/it/it_suite_test.go
```diff
@@ -126,6 +127,19 @@ var _ = BeforeSuite(func() {
 		err := tool.Cleanup(ctx)
 		Expect(err).ToNot(HaveOccurred())
 	})
+
+	// Create a default cluster version for version resolution during cluster creation:
+	cvClient := privatev1.NewClusterVersionsClient(tool.InternalView().AdminConn())
```
empovit, line 132: I would expect more unit tests. E.g. `IsDefault` with other values. Is an empty `Image` accepted? `Version`?

fulfillment-service/docs/CATALOG_ITEMS.md
```diff
@@ -270,7 +270,7 @@ catalog item, the server rejects any spec field not listed in `field_definitions
 |------|----------|-------------|
 | `pull_secret` | `--pull-secret` | Credentials for container image repositories |
 | `ssh_public_key` | `--ssh-public-key` | SSH public key installed on worker nodes |
-| `release_image` | `--release-image` | OCP release image URL |
+| `version_name` | `--version` | ClusterVersion name for the OpenShift release |
```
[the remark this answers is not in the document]
sk-ilya: osac create cluster --help

fulfillment-service/internal/servers/private_clusters_server_test.go
```diff
@@ -1494,7 +1523,7 @@ var _ = Describe("Private clusters server", func() {
 								}.Build(),
 							},
 							SpecDefaults: privatev1.ClusterTemplateSpecDefaults_builder{
-								ReleaseImage: new("quay.io/openshift-release-dev/ocp-release:4.22"),
+								SshPublicKey: proto.String("ssh-rsa TEMPLATE_DEFAULT_KEY"),
```
empovit, line 1526: I'm not sure I understand why `ReleaseImage` was replace by `SshPublicKey` :/

fulfillment-service/internal/servers/private_clusters_server.go
```diff
@@ -546,22 +578,41 @@ func (s *PrivateClustersServer) validateTemplateImmutability(ctx context.Context
...
+}
+
 // validateNetworkAttachmentImmutability ensures that the subnet field within
 // network_attachment cannot be changed after cluster creation.
 func (s *PrivateClustersServer) validateNetworkAttachmentImmutability(ctx context.Context,
 	request *privatev1.ClustersUpdateRequest) error {
 	updateMask := request.GetUpdateMask()
 
-	subnetMayChange := false
```
empovit, line 555: nice improvement :)

fulfillment-service/internal/servers/private_clusters_server.go
```diff
@@ -546,22 +578,41 @@ func (s *PrivateClustersServer) validateTemplateImmutability(ctx context.Context
 	return nil
 }
 
+// validateVersionNameUpdate ensures version_name stays valid on update. Absent
+// version_name is normalized to the existing value (cannot be cleared once set).
+// Changed values are validated the same way as Create via ensureClusterVersion.
+func (s *PrivateClustersServer) validateVersionNameUpdate(ctx context.Context,
+	request *privatev1.ClustersUpdateRequest) error {
+	if !updateIncludesField(request.GetUpdateMask(), "spec.version_name") {
```
empovit, line 586: Can this be a constant? Something coming from where it's defined? Let's say the filed name changes, but the string does not (even AI can miss that, AFAIK).

fulfillment-service/internal/cmd/cli/create/cluster/create_cluster_cmd.go
```diff
@@ -404,6 +415,61 @@ func (c *runnerContext) findTemplate(ctx context.Context) (result *publicv1.Clus
...
+	// precedence:
+	if len(matches) == 2 {
+		for _, m := range matches {
+			if m.GetMetadata().GetName() == c.args.version {
+				result = m
+				return
+			}
+		}
+		// Both matched but neither has metadata.name == ref; fall back to the first match:
+		result = matches[0]
```
empovit, line 453: Meaning two `4.17.0` with different names?

fulfillment-service/it/it_suite_test.go
sk-ilya: Here it's just a fixture creation. 

Some additional validations to the `ClusterVersion` are being added and tested here: https://github.com/osac-project/osac/pull/83

fulfillment-service/internal/servers/private_clusters_server_test.go
sk-ilya: This test's purpose is to verify that generic `spec_defaults` from templates propagate through  catalog items. Using `VersionName` here would conflate it with version resolution logic (which has its own dedicated tests added right below)

fulfillment-service/internal/servers/private_clusters_server.go
sk-ilya: The constant value will need to be updated in this case anyway. We can probably build this from the proto definition, maybe there are some reflection utils for that... but this will make it inconsistent with other servers, and this functionality is covered by tests anyway. So I agree it's not ideal, but it should be a broader refactor, not something I'd like doing in this already huge PR

fulfillment-service/internal/cmd/cli/create/cluster/create_cluster_cmd.go
sk-ilya: Technically yes. But this will never happen - both name and version are globally unique (tenant=shared).

vladikr:
Just checking... There are no existing clusters we should worry about, right? :)

sk-ilya:
@vladikr We are not GA, and there are no changes to `ClusterOrder`. I already posted a notification in Slack about this change a few days ago, and I'll post another one after this is merged.

fulfillment-service/internal/servers/private_clusters_server.go
```diff
@@ -546,22 +578,41 @@ func (s *PrivateClustersServer) validateTemplateImmutability(ctx context.Context
...
+	}
+	newSpec := request.GetObject().GetSpec()
+	if newSpec == nil {
+		return nil
+	}
+	existing, found, err := s.getExistingCluster(ctx, request)
+	if err != nil || !found {
+		return err
+	}
+	if !newSpec.HasVersionName() || newSpec.GetVersionName() == "" {
```
vladikr, line 597: Great!

fulfillment-service/internal/servers/cluster_version_resolution.go
```diff
@@ -0,0 +1,121 @@
...
+		SetLimit(1).
+		Do(ctx)
+	if err != nil {
+		logger.ErrorContext(ctx, "Failed to look up default cluster version",
+			slog.Any("error", err),
+		)
+		return "", grpcstatus.Errorf(grpccodes.Internal,
+			"failed to look up default cluster version")
+	}
+	if len(response.GetItems()) == 0 {
```
vladikr, line 106: that's better than the getSize you had before.

vladikr:
@sk-ilya, everything looks great, one question on the fields numbering. 
The design explicitly says:
```     
  message ClusterSpec {
    // Field 6 was release_image — removed.
    optional string version_name = 9;
  }
```
But in the implementation, you're reusing the release_image's number
```
  -  optional string release_image = 6;
  +  optional string version_name = 6;
```
I don't have a concern, but just curious if it's intentional?

sk-ilya:
@vladikr Yes, I don't think there is much point in continuing to add backward compatibility code for a project at such an early stage. 

(the same applies to the DB migrations - we already have a bunch of `delete_xxx_up.sql` files even though we technically dont support in-place upgrades, but that's a separate issue that needs to be solved)

vladikr:
Perfect, thank you! 
Looks solid to me.
/lgtm

vladikr:
/hold 
@empovit would you like to have another look before this goes in or are you satisfied?

PR #85, opened 2026-08-02
OSAC-1330: Replace raw string reference fields with typed Reference messages

htayrie-rh:
## Summary

- Introduces `LocalReference` and `Reference` protobuf message types in both public and private APIs, replacing raw `string` fields for resource references (virtual_network, subnet, security_group, host_type, template, etc.)
- Adds a generic DAO-based lookup factory (`NewDAOLookupFunc`) and registers all reference lookups in the gRPC server, enabling the existing reference validator interceptor to validate typed references by convention
- Updates rendering table CEL paths, SQL data migrations for backfill/trigger updates, and osac-aap template role metadata to support the new reference structure

### Proto changes
- New `reference_types.proto` in both `proto/public/` and `proto/private/` with 20 typed reference messages each (e.g., `VirtualNetworkLocalReference`, `HostTypeReference`, `TemplateReference`)
- All resource `*_type.proto` files updated to use typed references instead of raw strings

### Server & database changes
- `internal/references/lookups.go` — Generic lookup factory with `NewDAOLookupFunc[O dao.Object]`
- `internal/cmd/service/start/grpcserver/reference_lookups.go` — Registers all lookups for both public and private proto names
- `internal/servers/name_derivation.go` — `toDNSLabel()` and `templateNameFromID()` helpers
- `internal/servers/catalog_item_validation.go` — `resourceRef` interface and `refKey()` helper
- All private server files, controller files, CLI command files updated for typed reference getters

### Migrations
- `89_update_triggers_for_typed_references.up.sql` — Backfills data, drops Z0002 triggers, updates Z0003 triggers, recreates indexes
- `90_backfill_catalog_resource_names.up.sql` — Backfills catalog resource names
- `91_backfill_template_reference_fields.up.sql` — Backfills template reference fields

### Rendering
- All table definition YAMLs updated from `this.spec.field` to `this.spec.field.id`

### osac-aap
- `find_template_roles.py` updated to propagate `metadata_name` and emit a `metadata` computed field
- All template role `meta/osac.yaml` files now include explicit `name:` field

## Test plan

- [x] Unit tests pass (84 suites, 1492+ specs)
- [x] Integration tests pass (241/241 specs)
- [x] `buf lint` passes
- [x] `go build ./...` compiles cleanly
- [ ] CI checks (pre-commit, linting, integration tests)

## Related

- Based on draft PRs: #986 (fulfillment-service), #457 (osac-aap)
- osac-test-infra changes tracked separately (PR #272)


htayrie-rh:
## Pre-merge checklist

These temporary overrides must be reverted before merging this PR:

- [x] **`e2e-vmaas-full-install.yml`** — revert `test-infra-repository` to default and `test-infra-ref` to `main` (lines 104-105)
- [x] **`e2e-caas-full-install.yml`** — revert `test-infra-repository` to default and `test-infra-ref` to `main` (lines 108-109)
- [x] **`e2e-bmaas-full-install.yml`** — revert `test-infra-repository` to default and `test-infra-ref` to `main` (lines 97-98)

### Merge order

1. Merge **osac-test-infra PR #296** first (typed reference changes in E2E test helpers)
2. Revert the 3 workflow overrides listed above in this PR
3. Merge this PR

.github/workflows/e2e-bmaas-full-install.yml
htayrie-rh, line 1: revert all workflows files before merge

fulfillment-service/internal/cmd/service/start/grpcserver/reference_lookups.go
```diff
@@ -0,0 +1,250 @@
...
+)
+
+func registerReferenceLookups(
+	validator *references.ReferenceValidator,
+	logger *slog.Logger,
+	tenancyLogic auth.TenancyLogic,
+	metricsRegisterer prometheus.Registerer,
+) error {
+	// Networking references
+	virtualNetworksDAO, err := dao.NewGenericDAO[*privatev1.VirtualNetwork]().
```
htayrie-rh, line 35: need to consult if those DAOs can be generically instantiate  on-demand(?)

fulfillment-service/internal/computeinstancespec/validate_network.go
```diff
@@ -36,7 +36,8 @@ func ValidateNetworkAttachments(networkAttachments []*privatev1.NetworkAttachmen
 		if att == nil {
 			return fmt.Errorf("network_attachments[%d]: attachment cannot be null", i)
 		}
-		if att.GetSubnet() == "" {
+		ref := att.GetSubnet()
```
htayrie-rh, line 39: these validations can covered by protovalidate if we remove the call by reconciler. (if we accept that reconciler trusts the DB data)

fulfillment-service/proto/public/osac/public/v1/reference_types.proto
CrystalChun, line 1: Doesn't the design state that each reference type is supposed to live in its _type proto? 
https://github.com/osac-project/enhancement-proposals/pull/121/changes#diff-787598f5e0aba343e18a03910a7d6eb11d7842631525daf5183172566a08a42aR97-R103

> For each referenceable resource
   type, two new messages are added to its `_type.proto` file:

go.work.sum
CrystalChun, line 1: Why is this needed?

fulfillment-service/internal/controllers/natgateway/nat_gateway_reconciler_function.go
```diff
@@ -367,9 +367,21 @@ func (t *task) removeFinalizer() {
...
+	GetName() string
+}
+
+func refKeyStr(ref refKeyer) string {
+	if ref.GetId() != "" {
+		return ref.GetId()
+	}
+	return ref.GetName()
+}
+
```
CrystalChun: Since this function is used repeatedly, can we refactor it into a general util and use it for every file?

CrystalChun (changes requested):
For the future, it'd be nice to have these changes in smaller PRs rather than one large one as it is difficult to review.

For example, can we create changes by type instead of for all types at once? 
Can this change be in a different PR: 
```
- Add generic DAO lookup factory (NewDAOLookupFunc) for reference
  validation
```

.github/workflows/e2e-bmaas-full-install.yml
CrystalChun, line 1: @omer-vishlitzky any recommendation on changes that require a test infra change to go in around the same time?

osac-aap/collections/ansible_collections/osac/service/plugins/filter/find_template_roles.py
```diff
@@ -302,6 +302,7 @@ class NetworkClassCapabilities(Base):
 class Metadata(Base):
     """Metadata about the template"""
 
+    metadata_name: str | None = pydantic.Field(None, validation_alias="name")
```
CrystalChun, line 330: How come this new arg is needed?

go.work.sum
htayrie-rh: this is a new mono repo thing

osac-aap/collections/ansible_collections/osac/service/plugins/filter/find_template_roles.py
htayrie-rh: those catalog items being inserted by AAP didn't have name. since now we reference by name we need them to have metadata::name field.

.github/workflows/e2e-bmaas-full-install.yml
htayrie-rh: (currently this PR is running with private test-infra branch)

fulfillment-service/internal/controllers/natgateway/nat_gateway_reconciler_function.go
htayrie-rh: great minds think alike :) done yesterday

osac-operator/buf.gen.yaml
```diff
@@ -25,7 +25,7 @@ managed:
 
 inputs:
 
-- module: buf.build/osac-project/private-api:v0.0.79
+- directory: ../fulfillment-service/proto/private
```
htayrie-rh: @CrystalChun do you think this is valid change since we now have mono-repo

CrystalChun: cc @ygalblum

go.work.sum
CrystalChun: I mean are the changes in this file needed for this PR?

osac-operator/buf.gen.yaml
CrystalChun: Actually I don't think we should be updating this in this PR - if it needs to be changed it should be a separate PR

osac-aap/collections/ansible_collections/osac/service/plugins/filter/find_template_roles.py
CrystalChun: But how come we can't use `name`? Why do we need a new argument?

fulfillment-service/proto/public/osac/public/v1/compute_instance_type.proto
```diff
@@ -381,3 +384,25 @@ enum ComputeInstanceConditionType {
...
+  bool shared = 4;
+}
+
+// Reference to a ComputeInstanceCatalogItem resource.
+message ComputeInstanceCatalogItemReference {
+  string id = 1;
+  string name = 2;
+  string project = 3;
+  bool shared = 4;
+}
```
CrystalChun, line 408: Shouldn't this be in proto/public/osac/public/v1/compute_instance_catalog_item_type.proto? https://github.com/osac-project/osac/pull/85/changes#diff-2af775be6ce33ef790c6bad89bc3851c5582cbfbb9dcd0b3afe2b78707f80aebR17

fulfillment-service/proto/private/osac/private/v1/project_membership_type.proto
```diff
@@ -83,3 +82,11 @@ enum ProjectMembershipState {
...
   PROJECT_MEMBERSHIP_STATE_FAILED = 3;
 }
+
+// Reference to a User resource.
+message UserReference {
+  string id = 1;
+  string name = 2;
+  string project = 3;
+  bool shared = 4;
+}
```
CrystalChun: I think this should be in the users proto

fulfillment-service/internal/servers/servers_tenancy_test.go
```diff
@@ -344,7 +344,7 @@ var _ = Describe("Tenancy logic", func() {
 		createResponse, err := clustersServer.Create(ctx, publicv1.ClustersCreateRequest_builder{
 			Object: publicv1.Cluster_builder{
 				Spec: publicv1.ClusterSpec_builder{
-					Template: "my-template",
+					Template: publicv1.ClusterTemplateReference_builder{Id: "my-template"}.Build(),
```
CrystalChun, line 362: The goal is to move away from using IDs and to using `name` instead right? Should we already have these tests using `name` in preparation for that?

osac-operator/buf.gen.yaml
htayrie-rh: same problem as with e2e, it won't pass without it.

htayrie-rh: Talked to Omer about this, turns out it's an open discussion. meanwhile, I will revert it once ci is green and then bump version on follow up PR. (same for the e2e)

*we should broadcast on the channel when merging on red ci

FYI @omer-vishlitzky

fulfillment-service/proto/public/osac/public/v1/compute_instance_type.proto
htayrie-rh: Can't move it — compute_instance_catalog_item_type.proto already imports compute_instance_type.proto, so adding the reverse import would create a circular dependency.

osac-aap/collections/ansible_collections/osac/service/plugins/filter/find_template_roles.py
htayrie-rh: Role directory names use underscores (e.g. `ocp_small`) which aren't valid Kubernetes names. We added an explicit `name:` field to each `meta/osac.yaml` with the valid name (`ocp-small`) so it's defined at the source rather than derived by code.

danielerez:
Looks good.
Just need to revert [workflows](https://github.com/osac-project/osac/pull/85/changes#r3702793680) before merging.
/hold

fulfillment-service/proto/public/osac/public/v1/compute_instance_type.proto
CrystalChun: Wait they reference each other? How could this circular dependency actually exist 

Honestly I see that the compute instance type is imported into it but that's only because it's using compute instance template reference which also should be in the compute instance template type proto, no?

osac-operator/buf.gen.yaml
CrystalChun: Gotcha, thanks for the update, please open a ticket in case you'll be out when we need to do this

fulfillment-service/proto/public/osac/public/v1/compute_instance_type.proto
htayrie-rh: Here's the visual:

**Today:**
```
compute_instance_type.proto
├── defines: ComputeInstanceSpec (uses TemplateRef, CatalogItemRef)
├── defines: ComputeInstanceTemplateReference
├── defines: ComputeInstanceCatalogItemReference
└── defines: ComputeInstanceImage, ComputeInstanceDisk

compute_instance_template_type.proto
├── imports: compute_instance_type.proto  ── needs Image, Disk
└── defines: ComputeInstanceTemplate

compute_instance_catalog_item_type.proto
├── imports: compute_instance_type.proto  ── needs TemplateReference
└── defines: ComputeInstanceCatalogItem
```

**If we move `TemplateReference` → `compute_instance_template_type.proto`:**
```
compute_instance_type.proto
├── imports: compute_instance_template_type.proto  ◄── NEW (needs TemplateRef)

compute_instance_template_type.proto
├── imports: compute_instance_type.proto           ◄── EXISTING (needs Image, Disk)
                                        
                              ❌ CIRCULAR
```

Root cause: `Image` and `Disk` live in `compute_instance_type.proto` but are used by the template file. Moving any reference type back creates a cycle. Breaking it requires extracting shared types to a common file — out of scope here.

fulfillment-service/proto/private/osac/private/v1/project_membership_type.proto
htayrie-rh: done

PR #89, opened 2026-08-02
OSAC-2481: add vip_prefix_length to NetworkClassSpec

ori-amizur:
Add optional int32 vip_prefix_length field to NetworkClassSpec
(public and private). This deployment-wide parameter reserves the
last /N of each subnet CIDR for VIP allocation. At subnet creation,
the fabric manager excludes this range from its DHCP pool, and the
k8s manager creates an IPAddressPool covering only this range.

The field name is infrastructure-agnostic (not MetalLB-specific) to
stay consistent with the pluggable manager design.

The ipv6_subnet_cidr field referenced in the ticket already exists
(added as subnet_ipv6_cidr in a prior change).

## Summary by CodeRabbit

* **New Features**
  * Added an optional setting to reserve the final portion of each subnet for virtual IP allocation.
  * The reservation size can be configured from 1 to 128 prefix bits.


fulfillment-service/proto/public/osac/public/v1/network_class_type.proto
```diff
@@ -78,6 +78,15 @@ message NetworkClassSpec {
...
   // to remove capabilities that the provider has explicitly disabled for this region.
   NetworkClassCapabilities disable_capabilities = 2;
+
+  // Prefix length that reserves a sub-range of each subnet for MetalLB VIPs. This is a deployment-wide
+  // parameter that applies to all subnet creation, not just defaults. The fabric manager uses this to
+  // exclude the VIP sub-range from the DHCP pool.
+  // Example: 28 means a /28 sub-range (16 IPs) is reserved for MetalLB VIPs in each subnet.
+  optional int32 metallb_vip_prefix_length = 3 [(buf.validate.field).int32 = {
+    gte: 1,
+    lte: 128
```
danmanor, line 90: **Rename: `vip_prefix_length`** instead of `metallb_vip_prefix_length`.

MetalLB is one implementation — the API should stay infrastructure-agnostic, consistent with the rest of the networking design (pluggable managers via ConfigMap, no implementation-specific names in the API). The proto comment can mention MetalLB as an example consumer without coupling the field name to it.

fulfillment-service/proto/public/osac/public/v1/network_class_type.proto
```diff
@@ -78,6 +78,15 @@ message NetworkClassSpec {
   // the class's supported capabilities. The operator uses this during capabilities intersection
   // to remove capabilities that the provider has explicitly disabled for this region.
   NetworkClassCapabilities disable_capabilities = 2;
+
+  // Prefix length that reserves a sub-range of each subnet for MetalLB VIPs. This is a deployment-wide
+  // parameter that applies to all subnet creation, not just defaults. The fabric manager uses this to
+  // exclude the VIP sub-range from the DHCP pool.
+  // Example: 28 means a /28 sub-range (16 IPs) is reserved for MetalLB VIPs in each subnet.
+  optional int32 metallb_vip_prefix_length = 3 [(buf.validate.field).int32 = {
+    gte: 1,
```
danmanor, line 89: **Document the range convention in the proto comment.** The prefix length tells the system how many IPs to reserve, but not _where_ in the subnet. Both the k8s manager (MetalLB IPAddressPool) and the fabric manager (DHCP exclusion) need to agree on which IPs are reserved.

Suggested comment:

```protobuf
// Prefix length that reserves the last /N of each subnet CIDR for VIP allocation (e.g., cluster
// API server and ingress endpoints). This is a deployment-wide parameter. At subnet creation, the
// fabric manager excludes this range from its DHCP pool, and the k8s manager creates an IPAddressPool
// covering only this range.
// Example: with a 10.0.1.0/24 subnet and vip_prefix_length=28, the last /28 (10.0.1.240–10.0.1.255)
// is reserved for VIPs; DHCP serves 10.0.1.1–10.0.1.239.
```

This makes the "last /N" convention explicit in the API contract so future manager implementations don't have to guess.

danmanor (commented):
Two suggestions on this field:

fulfillment-service/proto/public/osac/public/v1/network_class_type.proto
```diff
@@ -78,6 +78,17 @@ message NetworkClassSpec {
...
+  // Prefix length that reserves the last /N of each subnet CIDR for VIP allocation (e.g., cluster
+  // API server and ingress endpoints). This is a deployment-wide parameter. At subnet creation, the
+  // fabric manager excludes this range from its DHCP pool, and the k8s manager creates an IPAddressPool
+  // covering only this range.
+  // Example: with a 10.0.1.0/24 subnet and vip_prefix_length=28, the last /28 (10.0.1.240-10.0.1.255)
+  // is reserved for VIPs; DHCP serves 10.0.1.1-10.0.1.239.
+  optional int32 vip_prefix_length = 3 [(buf.validate.field).int32 = {
+    gte: 1,
+    lte: 128
+  }];
```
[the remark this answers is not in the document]
ori-amizur: `vip_prefix_length` is on `NetworkClassSpec` — a deployment-wide constant with no reference to any subnet CIDR. The actual validation against a specific subnet's address family and parent prefix happens at subnet creation time in the fabric manager (OSAC-2482), when both values are known together. At NetworkClass creation time, no subnet exists yet, so there's nothing to validate against. The `1..128` range covers the widest valid prefix for any address family (IPv4 max /32, IPv6 max /128).

danmanor (approved):

PR #127, opened 2026-08-04
OSAC-1755: Add capabilities intersection to NetworkClass reconciler

SiddarthR56:
## Summary by CodeRabbit

- **New Features**
  - NetworkClass capabilities now synchronize automatically with configured network managers.
  - Synchronization runs periodically, with a configurable interval defaulting to five minutes.
  - Capability results reflect supported IP modes and DPU functionality.

- **Configuration**
  - Added configuration for the synchronization interval.
  - Updated network manager discovery labels to use Kubernetes-compliant formats.

- **Bug Fixes**
  - Improved handling of missing, unresolved, or invalid network managers without blocking other resources.

### DIFF
--- a/osac-operator/.claude/rules/configuration.md
+++ b/osac-operator/.claude/rules/configuration.md
@@ -20,6 +20,13 @@ Config via environment variables from a Secret (see `config/samples/osac-config-
 - `OSAC_CLUSTER_ORDER_NAMESPACE`, `OSAC_COMPUTE_INSTANCE_NAMESPACE`
 - `OSAC_TENANT_NAMESPACE`, `OSAC_NETWORKING_NAMESPACE`
 
+## Networking Managers
+
+- `OSAC_NETWORK_CLASS_SYNC_INTERVAL` — periodic full resync interval for NetworkClass
+  capabilities intersection, in addition to the immediate resync triggered by manager
+  ConfigMap changes (default: 5m). Only active when the fulfillment-service gRPC
+  connection is configured and `OSAC_NETWORKING_NAMESPACE` is non-empty.
+
 ## Controller Enable Flags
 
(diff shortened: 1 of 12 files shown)


danmanor (approved):
Reviewed the full diff against the unified networking design (OSAC-1433).

**Design alignment — good:**

- Capabilities intersection logic directly implements the design: "Capabilities are inferred from the assigned managers and published in the NetworkClass status — the provider does not set them manually. The operator computes the intersection of capabilities declared by the fabric manager and k8sManager ConfigMaps and populates status.capabilities automatically."
- fabric ∩ k8s when both present, fabric-only when no k8s manager — correct per design
- All 4 capabilities covered: `supports_ipv4`, `supports_ipv6`, `supports_dual_stack`, `dpu_support`
- ConfigMap-triggered reconciliation + periodic resync (for NC-side spec changes with no k8s event source) is a sound approach
- Pagination on `listAllNetworkClasses` avoids the silent truncation from the server's default page size
- `disable_capabilities` subtraction deferred to OSAC-2030 — noted in code, matches the design's `spec.disableCapabilities` field
- `resyncMu` + `TryLock` on the periodic path avoids races without blocking context cancellation

**Label rename** (`osac.openshift.io/network/fabric-manager` → `osac.openshift.io/network-fabric-manager`): K8s labels allow at most one `/`, so the old format was never valid. This is a bug fix. The design doc examples (OSAC-1433) and any existing ConfigMap samples will need updating to match — but since the old labels would have silently failed selectors, no deployment could have been relying on them.

**Coordination note:** PR #126 (OSAC-1460, dispatcher wiring) shares the `dispatcher_resolver_helpers_test.go` file and the `ErrFabricManagerNotSet` sentinel. These two PRs need to merge in a coordinated order or rebase on each other to avoid conflicts.

PR #82, opened 2026-08-02
OSAC-985: Add metering service and Kafka infrastructure

omer-vishlitzky:
## Summary

Phase 1 of OSAC metering: establishes the infrastructure foundation and delivers the first end-to-end pipeline — VMaaS lifecycle events flowing from the fulfillment-service Watch stream through a new Metering Service to Kafka.

### What Phase 1 includes
- **Kafka infrastructure** in `osac-installer/`: AMQ Streams operator subscription (Phase 1) and Kafka cluster with KRaft mode, SCRAM-SHA-512 + TLS (Phase 2)
- **osac-metering component**: Watch Consumer connects to the fulfillment-service private gRPC Watch stream, maps ComputeInstance lifecycle events to CloudEvents 1.0, publishes to `osac.metering.lifecycle` Kafka topic
- **VMaaS state machine**: maps OBJECT_UPDATED events to billing-relevant types based on current state — RUNNING → `started.v1`, STOPPED/PAUSED → `suspended.v1`, other → `updated.v1`
- **Cobra CLI** with `serve` subcommand, matching fulfillment-service pattern
- **Structured JSON logging** with consistent fields: event_id, resource_id, tenant_id, topic, type
- **Helm chart** with KafkaTopic CRs and KafkaUser (service owns its Kafka resources; installer provides only the cluster)
- **52 unit tests** across 3 suites (events, kafka, watch)
- Registered in `go.work` as a workspace module

### Tickets covered

| Ticket | Summary | Coverage |
|--------|---------|----------|
| [OSAC-3417](https://redhat.atlassian.net/browse/OSAC-3417) | Add Kafka operator installation to osac-installer | Fully met |
| [OSAC-3423](https://redhat.atlassian.net/browse/OSAC-3423) | Scaffold Metering Service Go binary and Helm chart | Met except W3C traceparent (see gaps) |
| [OSAC-3426](https://redhat.atlassian.net/browse/OSAC-3426) | Implement VMaaS Watch Consumer with Kafka publishing | Met except in-memory state tracking and resumed.v1 (see gaps) |
| [OSAC-3429](https://redhat.atlassian.net/browse/OSAC-3429) | Set up CI pipeline | N/A — mono-repo root CI covers build/test/lint |
| [OSAC-3434](https://redhat.atlassian.net/browse/OSAC-3434) | VMaaS lifecycle event e2e tests | Not started — separate QE work |

### Gaps from design and ticket ACs

[...]

PR #135, opened 2026-08-04
OSAC-3408: State Projection, Heartbeat Generator, and Reconciliation (Phase 2)

omer-vishlitzky:
## Summary

Phase 2 of OSAC metering: adds durable state tracking to the Metering Service.

- **State Projection** — PostgreSQL-backed table tracking every metered resource's state, billability, and billing dimensions. Watch Consumer reads previous state before mapping, enabling `resumed.v1` (STOPPED/PAUSED→RUNNING) and `duration_seconds` enrichment. Stale-update rejection via `fulfillment_version`.
- **Heartbeat Generator** — 60-second tick publishes `osac.resource.heartbeat.v1` for all billable resources. All-or-nothing per tick. Bounds metering gap to one tick on pod restart.
- **Reconciliation** — Startup + hourly reconciliation diffs State Projection against `ListComputeInstances`. Emits correction events (`missed_creation`, `state_drift`, `missed_deletion`) with publish-first strategy. Stale heartbeat detection emits synthetic catch-up heartbeats for billable resources with no heartbeat in >120s. Readiness probe gates on startup reconciliation.
- **Multi-topic routing** — Publisher routes by CloudEvent type: lifecycle/heartbeat/corrections topics. `TopicFor` returns error (not panic) for unknown types. Prometheus metrics on publish.
- **Test Adapter** — Minimal Kafka consumer pod for E2E metering verification (disabled by default via `testAdapter.enabled`). Consumes all metering topics, exposes events via `GET /events` HTTP API. Solves the DNS resolution issue where CI test containers cannot reach cluster-internal Kafka. Reference implementation of the Provider Adapter consumer pattern.

### Tickets covered

| Ticket | Summary | Coverage |
|--------|---------|----------|
| [OSAC-3418](https://redhat.atlassian.net/browse/OSAC-3418) | State Projection with PostgreSQL and Watch Consumer integration | Fully met |
| [OSAC-3422](https://redhat.atlassian.net/browse/OSAC-3422) | Heartbeat Generator with heartbeat Kafka topic | Fully met |
| [OSAC-3425](https://redhat.atlassian.net/browse/OSAC-3425) | Startup and hourly reconciliation | Fully met |

### Correctness hardening (post-review)

- **Version replay safety** — same-version upserts allowed (`>` not `>=`), so a Watch event replayed after publish failure can re-process without being silently dropped by the stale version guard
- **BillableSince clearing** — reconciliation drift corrections clear `BillableSince` when correcting to non-billable state, preventing stale duration computations in heartbeats
- **Stale heartbeat accuracy** — synthetic heartbeat detection reloads projection from DB after corrections are committed, instead of iterating the pre-correction in-memory snapshot (prevents false heartbeats for resources just corrected to non-billable)

### Design choices

- **Upsert-first ordering** (Watch Consumer) — projection updated before Kafka publish, per design doc crash-safety analysis. If publish fails, heartbeats + reconciliation compensate for the lost lifecycle event.
- **Duration from transition time** — `duration_seconds` computed from `transitionTime - billableSince` (event time, not wall clock), preventing inflated durations on processing delay.
- **Consolidated billing dimensions** — single `billingDims()` source; `BillingDimensions()` and `BillingDimensionsMap()` both derive from it.

### What Phase 2 does NOT include (deferred)

- [OSAC-3431](https://redhat.atlassian.net/browse/OSAC-3431): E2E tests (separate osac-test-infra PR)
- Prometheus metrics endpoint authentication
- Watch resume token (reconciliation compensates)
- `AffectedInterval` in correction events (adapters not yet consuming it)

## References

- **Epic:** [OSAC-3408](https://redhat.atlassian.net/browse/OSAC-3408)
- **Feature:** [OSAC-985](https://redhat.atlassian.net/browse/OSAC-985)
- **Design:** [enhancement-proposals#131](https://github.com/osac-project/enhancement-proposals/pull/131)
- **Phase 1 PR:** osac-project/osac#82
- **E2E tests PR:** osac-project/osac-test-infra#301

## Test plan

- [x] `cd osac-metering/metering-service && go build ./...` passes (both binaries)
- [x] `cd osac-metering/metering-service && ginkgo run -r internal` — 97 specs pass (7 suites)
- [x] `helm lint osac-metering/charts/osac-metering/` passes
- [x] `helm lint osac-metering/charts/osac-metering/ --set testAdapter.enabled=true` passes
- [x] `golangci-lint run ./...` — 0 issues
- [ ] Deploy to dev cluster: verify heartbeats on `osac.metering.heartbeat` topic at ~60s intervals
- [ ] Verify `resumed.v1` emitted on STOPPED→RUNNING transition
- [ ] Restart metering pod, verify startup reconciliation runs before readiness
- [ ] Deploy test adapter, verify `GET /events` returns consumed events

## Summary by CodeRabbit

- **New Features**
  - Added PostgreSQL-backed resource state storage with automatic setup and migrations.
  - Added reconciliation for missed events, state changes, deletions, and stale heartbeats.
  - Added heartbeat and correction events, including billing-dimension updates.
  - Added Kafka routing for lifecycle, heartbeat, and correction events.
  - Added an optional metering test adapter with event storage and HTTP endpoints.
  - Added secure database connection configuration for deployments.

- **Monitoring**
  - Added Prometheus metrics for publishing, heartbeats, reconciliation, and processing latency.

- **Bug Fixes**
  - Improved handling of state transitions, duplicate updates, stale versions, and Kafka timeouts.


osac-installer/charts/osac/templates/bundled-postgres.yaml
```diff
@@ -16,6 +16,17 @@ type: Opaque
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: metering-db
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "osac.labels" . | nindent 4 }}
+type: Opaque
+stringData:
+  url: "postgres://metering@postgres:5432/metering?sslmode=verify-full&sslrootcert=/etc/metering/db/sslrootcert&sslcert=/etc/metering/db/sslcert&sslkey=/etc/metering/db/sslkey"
```
masayag, line 28: Shouldn't this be configurable? same as the DB URL above? w.g what if an external DB is used? unless this will be taken care of by a different infra/core task.

osac-metering/metering-service/cmd/test-adapter/main.go
```diff
@@ -0,0 +1,190 @@
...
+		if json.Unmarshal(raw, &ev) != nil {
+			continue
+		}
+		if eventType != "" {
+			if t, _ := ev["type"].(string); t != eventType {
+				continue
+			}
+		}
+		if resourceID != "" {
+			if rid, _ := ev["osacresourceid"].(string); rid != resourceID {
```
masayag, line 66: the common CE base event fields repeats many time in the codebase. pls extract and reuse them from a single definition.

masayag:
## Review Findings

### Important

**2. Missing composite index from design spec**

The design document specifies:
```sql
CREATE INDEX idx_billable ON metering_resource_state (is_billable, last_heartbeat_at)
    WHERE is_billable = TRUE;
```

The implementation only indexes on `is_billable` without the `last_heartbeat_at` column. The composite index is needed for the stale heartbeat detection query in the reconciler, which filters `WHERE is_billable = TRUE AND last_heartbeat_at < threshold`. Without it, the stale heartbeat detection during reconciliation will require a sequential scan of all billable rows for the `last_heartbeat_at` filter.

Suggest adding `last_heartbeat_at` to the partial index to match the design.

---

**5. Readiness probe semantics**

The design states: "Readiness probe reports not-ready until startup reconciliation completes." The implementation sets `health.ready.Store(true)` after reconciliation, which is correct. However, the readiness check also verifies `health.conn != nil` (gRPC connection) — if the gRPC connection is lost post-startup, readiness will toggle to not-ready.

This matches standard Kubernetes semantics (readiness reflects ability to serve) but differs slightly from the design's framing which only ties readiness to startup reconciliation completion. Is this intentional? If the fulfillment-service connection drops, should the metering service become not-ready (triggering traffic removal) or stay ready since heartbeats and reconciliation can still function from the State Projection?

---

### Suggestions

**2. Test Adapter ring buffer size**

The test adapter uses a fixed `maxEvents = 10000` ring buffer. For E2E tests that exercise many lifecycle cycles (e.g., the performance test at 10K concurrent resources), this could wrap and drop events needed for assertions. Consider making `maxEvents` configurable via an environment variable (e.g., `MAX_EVENTS`) for CI tunability.

---

**3. Copyright year in new files**

The license headers in new files say `Copyright (c) 2025` but these are authored in 2026. Minor but should be updated for accuracy.

---

**4. Silent fallback in `parseDurationOrDefault`**

If `HEARTBEAT_INTERVAL` or `RECONCILIATION_INTERVAL` is set to an invalid value (e.g., `"abc"` or a negative duration), `parseDurationOrDefault` silently falls back to the default without logging. This could mask configuration errors in deployment. Consider logging a warning when a configured value is unparseable or non-positive, e.g.:

```go
func parseDurationOrDefault(s string, fallback time.Duration) time.Duration {
    if s == "" {
        return fallback
    }
    d, err := time.ParseDuration(s)
    if err != nil || d <= 0 {
        // Consider: logger.Info("ignoring invalid duration, using default", "input", s, "default", fallback)
        return fallback
    }
    return d
}
```

osac-installer/charts/osac/templates/bundled-postgres.yaml
omer-vishlitzky: this is only used by the ci currently

masayag:
## Go code review — follow-up on `ec19437`

`ec19437` nicely resolves the composite index gap, the `parseDurationOrDefault` silent-fallback, and the test adapter's hardcoded ring buffer size, plus a good DRY extraction (`events.SetOSACExtensions`) that also fixes a latent inconsistency (correction events now carry `osacproject` like every other event type). Nice cleanup.

A few items from an implementation-focused pass are still open in the current diff:

---

### Critical: time-precision bug in `shouldSkipUpdate`

```go
// internal/watch/consumer.go
if existing.TransitionTime != transitionTime.UTC() {
    c.logger.Info("skipping replayed event (upserted but likely unpublished)", ...)
} else {
    c.logger.V(1).Info("same state and dimensions, skipping", ...)
}
```

`transitionTime` comes from the incoming protobuf `Timestamp` (nanosecond precision); `existing.TransitionTime` was read back from Postgres `TIMESTAMPTZ` (microsecond precision, truncated on the prior write). A genuine replay of the same transition will almost never satisfy `==` here, so the "skipping replayed event" WARN-ish log will fire far more often than intended, and the log message's diagnostic value (distinguishing true no-ops from replayed-but-unpublished events) is lost. Suggest truncating both sides to a common precision (e.g. `.Truncate(time.Microsecond)`) or using a tolerance-based comparison before branching.

### Important: dead error-handling path for the heartbeat generator

```go
// cmd/metering-service/main.go
if err := hbGen.Run(ctx); err != nil {
    logger.Error(err, "heartbeat generator exited with error")
    runCancel()
}
```

`Generator.Run` only returns when `ctx` is done, and always returns `nil` at that point — it logs and continues on any per-tick error rather than propagating. So this branch is unreachable in practice, and a persistently failing heartbeat loop (e.g. broken DB connection) will degrade silently forever instead of crash-and-restart via the k8s supervisor. Either make `Run` surface fatal conditions (e.g. N consecutive failures) or drop the dead error-handling here to avoid the misleading impression that it's a safety net.

### Important: duplicated "is billable" logic

```go
// internal/reconciliation/reconciler.go
func isBillableState(state string) bool {
    return state == "RUNNING"
}
```

This re-implements what `computeInstanceMapper.IsBillable()` already determines from the proto enum, via a string comparison in a second location. If a new billable state is ever added, it's easy to update one and miss the other, silently diverging billing behavior between the watch path and the reconciliation path. Worth consolidating into a single source of truth.

### Important: silent skip when fulfillment is behind projection

```go
// internal/reconciliation/reconciler.go
if fs.version < ps.FulfillmentVersion {
    continue
}
```

No log or metric here. If fulfillment gets stuck behind the projection for a given resource (e.g. an upstream caching bug), reconciliation will quietly no-op on it every cycle with no observable signal. Suggest at least a debug log or a counter so this is discoverable.

### Minor: `ProjectID` stored as empty string instead of `NULL`

```go
// internal/projection/postgres.go — Upsert
state.ProjectID,                      // no nullIfEmpty(...)
nullIfEmpty(state.PreviousState),     // ...but this field gets it
```

`project_id` is documented as nullable, and `PreviousState` on the same row goes through `nullIfEmpty`, but `ProjectID` is written as-is. Harmless for round-tripping through the Go struct, but any future `WHERE project_id IS NULL` query (or a downstream consumer relying on NULL vs `''`) will behave inconsistently.

### Minor: `omitempty` on `BillingDimensions` in correction events

```go
// internal/reconciliation/correction.go
BillingDimensions map[string]any `json:"billing_dimensions,omitempty"`
```

`omitempty` elides the field for both `nil` and zero-length maps, whereas the design's correction event schema shows `billing_dimensions` always present. Low risk in practice today (compute instance dimensions are effectively never empty), but worth aligning with the schema for adapters that might parse this strictly.

### Minor: stale `AffectedInterval` TODO

`buildCorrectionEvent`'s `interval *AffectedInterval` parameter is threaded through everywhere but never populated in this phase. A short `// TODO(OSAC-xxxx): populate AffectedInterval once...` at the call sites would make the deferred scope explicit for the next reader.

masayag (approved):
/lgtm
/approve

PR #137, opened 2026-08-04
OSAC-3503: Wire image source_type through BMI provisioning path

mennyaboush:
## Summary

- Add `imageSourceType` to the template parameters map in the BMI reconciler, so the `--image-source-type` CLI flag actually affects provisioning
- Update the Metal3 `build_bmh_patch.yaml` to use `imageSourceType` when present for checksum decisions, falling back to URL-prefix detection for backward compatibility
- Add unit tests for the new parameter: presence, absence, empty source_type, and system-override of user-provided template_parameters

## Problem

The `--image-source-type` flag on `osac create baremetalinstance` was accepted, defaulted, validated, and stored by the API — but never forwarded to the Ansible provisioning pipeline. The reconciler only extracted `imageURL` from the image spec, silently dropping `source_type`. This meant setting `--image-source-type` to any value had zero effect on how the bare metal instance was provisioned.

### fulfillment-service (reconciler)
- `baremetalinstance_reconciler_function.go`: Extract `GetSourceType()` into `params["imageSourceType"]` when non-empty (3 lines added)
- `baremetalinstance_reconciler_function_test.go`: Added assertions to 4 existing tests + 2 new tests covering override and empty source_type scenarios

### osac-aap (Ansible)
- `build_bmh_patch.yaml`: Updated checksum condition to prefer explicit `imageSourceType` over URL-prefix detection, with `| default('')` fallback for backward compatibility

## Rollout note

Existing BMIs with `source_type` set will receive an updated `TemplateParameters` on the next reconcile (adding the `imageSourceType` key), which may trigger a config version bump in the BMFO controller.

## Test plan

- [x] `go build ./...` passes
- [x] `ginkgo run -r internal/controllers/baremetalinstance/` — 71 tests pass
- [ ] `ansible-lint` on the changed role
- [ ] E2E: Create BMI with `--image-source-type oci` and verify `imageSourceType` appears in BMFO CR's `TemplateParameters`
- [ ] E2E: Verify BMH patch skips checksum for OCI images

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced bare-metal instance provisioning parameters by including `imageSourceType` when an image source type is provided, and omitting it when not.
  * Prevented checksum injection for OCI images while keeping existing URL-based OCI handling for empty source types.
  * Ensured system-provided image values override any user-provided `imageSourceType`.
* **Tests**
  * Extended reconciler tests to validate `imageSourceType` behavior and precedence rules across image and system overrides.

### DIFF
--- a/osac-aap/collections/ansible_collections/osac/templates/roles/bm_host_provisioning/tasks/build_bmh_patch.yaml
+++ b/osac-aap/collections/ansible_collections/osac/templates/roles/bm_host_provisioning/tasks/build_bmh_patch.yaml
@@ -8,7 +8,10 @@
         online: true
 
 - name: Add checksum to BMH patch for non-OCI images
-  when: not template_params.imageURL.startswith('oci://')
+  when: >-
+    (template_params.imageSourceType | default('') != 'registry')
+    and
+    (template_params.imageSourceType | default('') != '' or not template_params.imageURL.startswith('oci://'))
   ansible.builtin.set_fact:
     bmh_patch: >-
       {{ bmh_patch | combine({
--- a/fulfillment-service/internal/controllers/baremetalinstance/baremetalinstance_reconciler_function.go
+++ b/fulfillment-service/internal/controllers/baremetalinstance/baremetalinstance_reconciler_function.go
@@ -595,6 +595,11 @@ func (t *task) mutateBMI(ctx context.Context, object *bmfov1alpha1.BareMetalInst
 	}
 	if t.bareMetalInstance.GetSpec().HasImage() {
 		params["imageURL"] = t.bareMetalInstance.GetSpec().GetImage().GetSourceRef()
+		if st := t.bareMetalInstance.GetSpec().GetImage().GetSourceType(); st != "" {
+			params["imageSourceType"] = st
+		} else {
+			delete(params, "imageSourceType")
+		}
 	}
 	if len(params) > 0 {
 		paramsJSON, err := json.Marshal(params)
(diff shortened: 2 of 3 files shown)


osac-aap/collections/ansible_collections/osac/templates/roles/bm_host_provisioning/tasks/build_bmh_patch.yaml
```diff
@@ -8,7 +8,10 @@
         online: true
 
 - name: Add checksum to BMH patch for non-OCI images
-  when: not template_params.imageURL.startswith('oci://')
+  when: >-
+    (template_params.imageSourceType | default('') != 'oci')
```
adriengentil:
```suggestion
    (template_params.imageSourceType | default('') != 'registry')
```

mennyaboush: Thanks for the review Adrien:)

If I change line 12 from `!= 'oci'` to `!= 'registry'`, then `imageSourceType=oci` would now ADD checksum - but the original code skipped checksum for `oci://` URLs. That would be a behavioral change for the oci flow.

Should I:
1. Replace `oci` with `registry` everywhere (skip checksum for `registry` only, add checksum for `oci`)
2. Skip checksum for both `registry` and `oci` (i.e. `not in ['oci', 'registry']`)
3. Something else — should the URL-prefix fallback also change?

Want to make sure I get the Metal3 semantics right.

adriengentil: "registry" is the right key to point to an OCI registry, it's already used on VMaaS side

adriengentil (approved):

adriengentil (approved):

PR #144, opened 2026-08-04
OSAC-3500: Make osac CLI help colorless by default

achuzhoy:
Color is now opt-in: default rendering uses a plain ASCII style
    regardless of terminal type. Add --color=yes|true to enable styled output;
    
    Clear all heading prefixes (H1-H6) so plain-mode output does not show
    raw Markdown syntax. Defer terminal background detection until color
    is actually enabled. Use fmt.Fprint for colored output and
    lipgloss.Fprint for plain output to avoid leaking ANSI codes.
    
    Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
    Signed-off-by: Alexander Chuzhoy <achuzhoy@redhat.com>

### DIFF
--- a/fulfillment-service/internal/cmd/cli/help/help_setup.go
+++ b/fulfillment-service/internal/cmd/cli/help/help_setup.go
@@ -17,6 +17,7 @@ import (
 	"bytes"
 	"embed"
 	"encoding/json"
+	"fmt"
 	"log/slog"
 	"os"
 	"path/filepath"
@@ -35,6 +36,26 @@ import (
 //go:embed templates
 var templatesFS embed.FS
 
+const colorFlag = "color"
+
+func colorEnabled(c *cobra.Command) bool {
+	// NO_COLOR always wins (https://no-color.org/), including over FORCE_COLOR and --color.
+	if os.Getenv("NO_COLOR") != "" {
+		return false
+	}
+
+	flags := c.Root().PersistentFlags()
+	if flags.Changed(colorFlag) {
+		enabled, err := flags.GetBool(colorFlag)
+		if err == nil {
+			return enabled
+		}
+	}
+
+	// FORCE_COLOR enables color for environment-based callers when the flag is not set.
+	return os.Getenv("FORCE_COLOR") != ""
+}
+
 // Setup configures the given command and all its subcommands to render their help output as styled Markdown.
 func Setup(cmd *cobra.Command) {
 	// Create a silent logger for the templating engine, as help rendering happens before the persistent pre-run
@@ -52,19 +73,22 @@ func Setup(cmd *cobra.Command) {
 		return
 	}
 
+	cmd.PersistentFlags().Bool(
+		colorFlag,
+		false,
+		"Enable colored help output. NO_COLOR and --color=false override FORCE_COLOR.",
+	)
+
 	// Set the help function for the command and all its subcommands. The renderer is created each time the
 	// help is displayed, so that it can adapt to the current terminal width and color capabilities.
 	cmd.SetHelpFunc(func(c *cobra.Command, args []string) {
+		// If the output is a terminal, we want to adjust the width of the terminal, but never more than the
+		// maximun width that we consider readable:
 		out := c.OutOrStdout()
-
-		// Determine if the output is a terminal and get its size:
 		var width int
-		isTTY := false
-		file, ok := out.(*os.File)
-		if ok {
+		if file, ok := out.(*os.File); ok {
 			fd := int(file.Fd())
 			if term.IsTerminal(fd) {
-				isTTY = true
 				width, _, err = term.GetSize(fd)
 				if err != nil {
 					c.PrintErrln("Error getting terminal size:", err)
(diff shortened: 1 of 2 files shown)


ygalblum:
You wrote:
> there was no NO_COLOR support

But, the code does support it. If I run:
```bash
NO_COLOR= osac help
```
I get an output with no color

In addition, not sure why we need to change the default behaviour.

achuzhoy:
> You wrote:
> 
> > there was no NO_COLOR support
> 
> But, the code does support it. If I run:
> 
> ```shell
> NO_COLOR= osac help
> ```
> 
> I get an output with no color
> 
> In addition, not sure why we need to change the default behaviour.

@ygalblum 
Previously there was no support at all. Now the support was added but it's not the default. IMO it should be the default. Making colorful output should be done with COLOR=

ygalblum:
> Previously there was no support at all. Now the support was added but it's not the default. IMO it should be the default. 

Not sure I agree. The color option is for humans running the CLI. The won't know about the color option and I'm not sure they should. Furthermore, having to set the environment variable it a bit cumbersome for a human using a terminal. Someone piping the output already has a more complex line, so it's simpler to add it.

Having said that, if you strongly believe defaulting to no color, I won't argue again :). But, please see below 

> Making colorful output should be done with COLOR=

This is not what the current PR suggests. It uses no color by default but then has two environment variables with precedence. If we go in this path, just have a default and one variable that adds color (as you wrote here). In addition, document the variable in the help's output

fulfillment-service/internal/cmd/cli/help/help_setup.go
```diff
@@ -74,35 +131,34 @@ func Setup(cmd *cobra.Command) {
 		}
 		width = min(width, maxReadableWidth)
 
-		// Select the style based on terminal capabilities. Use colored styles only when the output is
-		// a terminal and the NO_COLOR environment variable is not set (https://no-color.org/).
-		_, noColor := os.LookupEnv("NO_COLOR")
+		// Color is opt-in: default is plain/no-color. Pass --color=true or --color=yes to enable styled output.
+		useColor := colorEnabled(c)
```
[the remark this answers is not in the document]
achuzhoy: Where does the requirement for FORCE_COLOR and/or NO_COLOR come from?

fulfillment-service/internal/cmd/cli/help/help_setup.go
```diff
@@ -35,6 +38,54 @@ import (
 //go:embed templates
 var templatesFS embed.FS
 
+const colorFlag = "color"
```
ygalblum, line 39: No need for this complex boolean from string type. Just have a boolean parameter with default of `false`. The user can set it by setting `--color` or not pass anything

fulfillment-service/internal/cmd/cli/help/help_setup.go
```diff
@@ -130,7 +186,11 @@ func Setup(cmd *cobra.Command) {
 			c.Print(buffer.String())
 			return
 		}
-		_, err = lipgloss.Fprint(out, text)
+		if useColor {
+			_, err = fmt.Fprint(out, text)
```
ygalblum, line 157: Why is this change needed? Before, the color output used `lipgloss`, why can't it continue doing that?

fulfillment-service/internal/cmd/cli/help/help_setup.go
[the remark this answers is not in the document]
achuzhoy: The commit message was changed. Update the description per the commit message

fulfillment-service/internal/cmd/cli/help/help_setup.go
achuzhoy: Good question. Colored output could use lipgloss.Fprint on a real TTY — that’s what the old code did. The split isn’t about glamour or lipgloss “rendering” differently; it’s about what happens on write.

What each call does
glamour already renders markdown into a string (text) with ANSI sequences baked in when using DarkStyleConfig / LightStyleConfig.

Then:

fmt.Fprint(out, text) — writes the string as-is.
lipgloss.Fprint(out, text) — wraps the writer with colorprofile.NewWriter, which inspects the output destination and environment (NO_COLOR, CLICOLOR, CLICOLOR_FORCE, TTY detection) and may strip or downsample ANSI.
From lipgloss’s Fprint:


writer.go
Ln 67–68
func Fprint(w io.Writer, v ...any) (int, error) {
	return fmt.Fprint(colorprofile.NewWriter(w, os.Environ()), v...)
And when the profile is non-TTY, colorprofile strips all ANSI:


writer.go
Ln 42–44
case w.Profile <= NoTTY:
	_, err := io.WriteString(w.Forward, ansi.Strip(string(p)))
	return len(p), err
Why the split exists
Mode	Writer	Reason
Plain (useColor == false)
lipgloss.Fprint
Safety net: strips any stray ANSI (pipes, non-TTY, NO_COLOR, etc.) even if glamour left some behind
Colored (useColor == true)
fmt.Fprint
User explicitly opted in (--color, FORCE_COLOR); write glamour’s ANSI through without colorprofile stripping it
The important case is colored help to a non-TTY — tests use bytes.Buffer, users pipe help, CI captures stdout. With lipgloss.Fprint, colorprofile sees non-TTY and strips ANSI even when --color or FORCE_COLOR says color is wanted.

Also, lipgloss/colorprofile honors CLICOLOR_FORCE, not FORCE_COLOR. Your colorEnabled() logic treats FORCE_COLOR as opt-in, but lipgloss.Fprint would ignore that on non-TTY output.

Could colored go back to lipgloss.Fprint?
Only if you’re fine with colored help never showing color when:

output is piped or redirected
FORCE_COLOR is set but output isn’t a TTY
unit tests assert on ANSI in a buffer
For interactive terminal-only color, lipgloss.Fprint would likely look the same. The fmt.Fprint branch is there so explicit opt-in actually produces colored bytes in all those other cases.

Bottom line: glamour still does the styling in both paths; the change is about whether the final write pass-through preserves or strips ANSI. Plain → strip; explicitly colored → preserve.

fulfillment-service/internal/cmd/cli/help/help_setup.go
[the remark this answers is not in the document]
achuzhoy: recheck please

fulfillment-service/internal/cmd/cli/help/help_setup.go
achuzhoy: Updated. Thanks.

ygalblum (approved):
/approve
/lgtm

PR #147, opened 2026-08-05
OSAC-1459: Extend provisioning lifecycle for multi-job tracking

SiddarthR56:
## Summary by CodeRabbit

* **New Features**
  * Added support for provisioning and deprovisioning resources through multiple targets.
  * Job statuses now identify the target handling each operation.
  * Added independent tracking, retries, polling, and error handling for each target.
  * Preserved existing behavior for single-target resources.

* **Validation**
  * Added checks for missing, duplicate, or invalid target configurations.

### DIFF
--- a/osac-operator/pkg/provisioning/provision_lifecycle.go
+++ b/osac-operator/pkg/provisioning/provision_lifecycle.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"hash/fnv"
 	"time"
@@ -48,7 +49,14 @@ type JobsExtractor func(client.Object) []v1alpha1.JobStatus
 
 // EvaluateAction determines the next provisioning action based on job history and config versions.
 func EvaluateAction(provState *State, checkAPIServer func() bool) (Action, *v1alpha1.JobStatus) {
-	latestJob := FindLatestJobByType(*provState.Jobs, v1alpha1.JobTypeProvision)
+	return evaluateActionForTarget(provState, "", checkAPIServer)
+}
+
+// evaluateActionForTarget is EvaluateAction scoped to a single job target, so
+// concurrent targets on the same resource are evaluated from only their own
+// job history.
+func evaluateActionForTarget(provState *State, target string, checkAPIServer func() bool) (Action, *v1alpha1.JobStatus) {
+	latestJob := FindLatestJobByTypeAndTarget(*provState.Jobs, v1alpha1.JobTypeProvision, target)
 
 	if !HasJobID(latestJob) {
 		// No provision job exists — trigger one.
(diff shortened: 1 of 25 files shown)


osac-operator/pkg/provisioning/provision_lifecycle.go
```diff
@@ -254,38 +418,46 @@ func ComputeDesiredConfigVersion(spec any) (string, error) {
...
+
+// triggerDeprovisionJobForTarget is TriggerDeprovisionJob scoped to a single
+// job target: the appended deprovision JobStatus is tagged with target.
+func triggerDeprovisionJobForTarget(ctx context.Context, provider ProvisioningProvider, resource client.Object,
+	jobs *[]v1alpha1.JobStatus, target string, maxHistory int, pollInterval time.Duration) (ctrl.Result, error) {
 	log := ctrllog.FromContext(ctx)
-	log.Info("triggering deprovision job")
+	log.Info("triggering deprovision job", "target", target)
 
 	result, err := provider.TriggerDeprovision(ctx, resource, *jobs)
```
ori-amizur, line 431: **Bug: unfiltered jobs passed to provider.**

This passes the entire `*jobs` slice (all targets) to `provider.TriggerDeprovision`. The `ProvisioningProvider.TriggerDeprovision` interface takes `provisionJobs []v1alpha1.JobStatus` so providers can inspect provision job history (e.g., to cancel a running provision before deprovisioning). A fabric provider receiving k8s jobs — or vice versa — would see irrelevant history and could make wrong decisions based on it.

Suggestion: filter `*jobs` to only those matching `target` before passing:

```go
targetJobs := filterJobsByTarget(*jobs, target)
result, err := provider.TriggerDeprovision(ctx, resource, targetJobs)
```

The test mock currently discards the `provisionJobs` argument in `triggerDeprovisionFunc`, so this bug has no failing test. Consider widening the mock callback signature to accept and assert on `provisionJobs`.

osac-operator/pkg/provisioning/provision_lifecycle.go
```diff
@@ -183,42 +208,181 @@ func RunProvisioningLifecycle(
...
+	provider ProvisioningProvider,
+	resource client.Object,
+	provState *State,
+	target string,
+	maxHistory int,
+	pollInterval time.Duration,
+	callbacks *PollCallbacks,
+	checkAPIServer func() bool,
+) (ctrl.Result, error, bool) {
+	action, latestJob := evaluateActionForTarget(provState, target, checkAPIServer)
```
ori-amizur, line 240: **Nit: `error` should be the last return value.**

`runLifecycleCore` returns `(ctrl.Result, error, bool)`. Go convention (and `revive`'s `error-return` rule) places `error` last. Since this function is unexported with only two call sites, swapping to `(ctrl.Result, bool, error)` is a contained change.

osac-operator/pkg/provisioning/provision_lifecycle.go
```diff
@@ -183,42 +208,181 @@ func RunProvisioningLifecycle(
...
+// target has reached Skip), and joins errors from all targets rather than
+// short-circuiting on the first one — one target's failure never prevents
+// another target's attempt in the same call. statusFlush is called at most
+// once per call, only if at least one target triggered a new job.
+//
+// Every target's manager must have actually been dispatched to before being
+// included here — a target that was never dispatched (e.g. a resource with
+// no k8s manager configured) has no job history to evaluate and should be
+// omitted by the caller rather than passed in.
+func RunMultiTargetProvisioningLifecycle(
```
ori-amizur, line 317: **Missing migration path for legacy untagged jobs.**

Existing resources have jobs with `Target: ""`. When a controller switches from `RunProvisioningLifecycle` to `RunMultiTargetProvisioningLifecycle`, those untagged jobs won't match any non-empty target name — `FindLatestJobByTypeAndTarget(jobs, Provision, "fabric")` returns nil for a job with `Target: ""`. Each target will see no prior job history and trigger a fresh provision, potentially re-provisioning something that's already running.

The test suite starts from either empty jobs or already-tagged jobs; there's no test for the upgrade path. I'd suggest at minimum:
1. Document the expected migration behavior (e.g., does the caller need to backfill `Target` on existing jobs before switching to multi-target? Or should `evaluateActionForTarget` treat untagged jobs as belonging to a designated legacy target?).
2. Add a test with pre-existing `Target: ""` jobs to encode whichever strategy you choose.

osac-operator/pkg/provisioning/provision_lifecycle.go
```diff
@@ -77,28 +85,44 @@ func EvaluateAction(provState *State, checkAPIServer func() bool) (Action, *v1al
...
+
+// CheckAPIServerForNonTerminalProvisionJobAndTarget is
+// CheckAPIServerForNonTerminalProvisionJob scoped to a single job target —
+// required when populating JobTarget.CheckAPIServer for a non-"" target,
+// since the untargeted form only ever checks untagged (target == "") job
+// history.
+func CheckAPIServerForNonTerminalProvisionJobAndTarget(ctx context.Context, apiReader client.Reader, key client.ObjectKey, fresh client.Object, extract JobsExtractor, target string) bool {
 	log := ctrllog.FromContext(ctx)
 	if err := apiReader.Get(ctx, key, fresh); err != nil {
 		return false
```
ori-amizur, line 99: **Silenced API server read error.**

When `apiReader.Get` fails, this returns `false` (no non-terminal job found), which bypasses the duplicate-job guard and lets a new job be triggered. The error is silently discarded. Consider logging it so operators can diagnose why the guard was bypassed:

```go
if err := apiReader.Get(ctx, key, fresh); err != nil {
	log.Error(err, "failed to read resource from API server; proceeding without duplicate-job check", "target", target)
	return false
}
```

ori-amizur (commented):
Four issues flagged — two I consider blocking (#1 and #3), two are cleanup (#2 and #4). Details in inline comments.

danmanor (approved):

PR #157, opened 2026-08-05
OSAC-2158: Implement ClusterVersion delete protection

sk-ilya:
## Summary
- Add PostgreSQL BEFORE UPDATE trigger (Z0003) on `cluster_versions` that blocks soft-deletion when active clusters, templates, or catalog items reference the version by name.
- Inbound (forward-reference) validation is handled by the existing gRPC reference validation interceptor (PR #183), not by database triggers.

## Test plan
- [ ] Migration tests: 7 specs covering outbound delete protection (blocked and permitted scenarios)
- [ ] Server tests: 3 specs exercising the full gRPC path (server.Delete -> DAO -> trigger -> ErrInUse -> FailedPrecondition)
- [ ] Full migration suite and servers suite pass with no regressions


fulfillment-service/internal/database/migrations/93_add_cluster_version_delete_protection_trigger.up.sql
```diff
@@ -0,0 +1,195 @@
...
+  execute function check_cluster_version_ref('spec', 'version_name');
+
+-- Attach the trigger so that it fires on update of active clusters, but only when the version_name reference
+-- actually changes (avoids a FOR SHARE lock on every status update):
+create trigger check_cluster_version_name_ref_on_update
+  before update on clusters
+  for each row
+  when (new.deletion_timestamp = 'epoch' and
+    (new.data->'spec'->>'version_name') is distinct from (old.data->'spec'->>'version_name'))
+  execute function check_cluster_version_ref('spec', 'version_name');
```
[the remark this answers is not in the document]
sk-ilya: the resource can be restored  - how?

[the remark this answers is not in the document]
sk-ilya: are there any other cases of such a reference protection in this codebase

fulfillment-service/internal/database/migrations/94_add_cluster_version_delete_protection_trigger.up.sql
```diff
@@ -0,0 +1,89 @@
...
+--
+-- Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+-- an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+-- specific language governing permissions and limitations under the License.
+--
+
+-- This migration adds a BEFORE UPDATE trigger on cluster_versions that prevents soft-deleting a version while an
+-- active cluster, cluster template, or cluster catalog item still references it. Inbound (forward-reference)
+-- validation is handled by the gRPC reference validation interceptor, not by database triggers.
+--
```
tzvatot, line 17: 🟡 **Inbound reference validation — does it exist?**

This comment states inbound validation is handled by the gRPC reference validation interceptor. However, I couldn't find a `ClusterVersionReference` lookup registered with the `ReferenceValidator` in the gRPC server setup (`start_grpc_server_cmd.go:436`). The validator is created but `.Register()` is never called for any reference type in production code.

Without inbound validation, users could create clusters/templates referencing non-existent or deleted cluster versions — the outbound protection would be one-sided.

The established pattern (migrations 56, 76, 81) uses bidirectional DB triggers with FOR SHARE locking to eliminate TOCTOU races. Is the inbound side handled elsewhere (separate PR, server-level check), or is this a known gap?

fulfillment-service/internal/database/migrations/94_add_cluster_version_delete_protection_trigger.up.sql
```diff
@@ -0,0 +1,89 @@
...
+-- Unlike other Z0003 triggers (migration 90) which match on ->>'id', this trigger matches on ->>'name' because
+-- the ClusterVersionReference stores the version name, and migration 93 backfilled only the name field.
+
+-- Index to speed up the JSONB text extraction used by the outbound trigger:
+create index clusters_version on clusters
+  ((data->'spec'->'version'->>'name'))
+  where data->'spec'->'version'->>'name' is not null;
+
+-- Trigger function that checks whether any active cluster, cluster template, or cluster catalog item references the
```
tzvatot, line 26: 💡 **Consider adding expression index for cluster_templates**

The migration indexes the `clusters` table lookup but not `cluster_templates` (on `data->'spec_defaults'->'version'->>'name'`). While version deletion is rare and the table is small, adding the index would be consistent. The `cluster_catalog_items` lateral unnest can't be served by a plain expression index, so that one is fine without.

tzvatot (changes requested):
Clean, focused change — the Z0003 trigger follows the established pattern, `exists(select 1 ...)` avoids leaking referencing IDs (consistent with migration 72), and the tests cover all three referencing resource types at both migration and server layers.

| Category | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟡 Important | 2 |
| 💡 Suggestion | 1 |

### 🟡 PR description is stale

The description mentions "inbound triggers (Z0002) with FOR SHARE locking", `ensure_active_cluster_version()`, `check_cluster_version_ref()`, and "23 specs" — none of which exist in the current diff. The actual PR contains only the outbound Z0003 trigger and 10 test specs (7 migration + 3 server). Should be updated to match the actual scope.

fulfillment-service/internal/database/migrations/94_add_cluster_version_delete_protection_trigger.up.sql
sk-ilya: I don't think adding the index would necessarily make this more consistent, since, as you mentioned, `cluster_catalog_items` doesn't have one either. Given how small `cluster_templates` is, I don't think there is much benefit in adding an index here.

fulfillment-service/internal/database/migrations/94_add_cluster_version_delete_protection_trigger.up.sql
sk-ilya: https://github.com/osac-project/osac/blob/b465d11f352219647e0c24fd7cd8cb1bdfae7915/fulfillment-service/internal/cmd/service/start/grpcserver/reference_lookups.go#L249-L259

Merged [recently](https://github.com/osac-project/osac/pull/183/changes#diff-b8ede1153377952feb25ca94d084b7f2804e04a009b3cb8fba626bb18510439cR249)

tzvatot (approved):
**Re-review** - PR has been significantly reworked since my last review.

## Previous Findings Status

| # | Finding | Status | Reason |
|---|---------|--------|--------|
| 1 | 🟡 Inbound reference validation - does it exist? | **ACCEPTED** | PR #183 added `ClusterVersionReference` lookup to `reference_lookups.go`, now merged. PR description correctly updated. |
| 2 | 💡 Expression index for cluster_templates | **ACCEPTED** | Author's reasoning is sound - table is small, consistent with `cluster_catalog_items` also being unindexed. |

## New Issues

None. The reworked PR correctly narrows scope to outbound-only Z0003 trigger. Trigger function follows established patterns, JSONB path matching is correct, catalog item array scan handles edge cases, and the test coverage (7 migration + 3 server specs) is comprehensive.

/lgtm

PR #161, opened 2026-08-05
OSAC-3669: Fix CLI hangs when OAuth token/logout endpoints stall

wgordon17:
## Summary
- `sendForm()` (login/token refresh) and `logout_cmd.go`'s `terminateSession`/`refreshForIdToken` all sent requests via `http.Client.PostForm`/`.Get`, which build their request with `context.Background()` internally, so the caller's context could never cancel a stalled request; combined with falling back to an unbounded-timeout client, any stall talking to the OAuth server hung the CLI forever
- Requests are now built explicitly with `http.NewRequestWithContext` so the caller's context is honored, and a bounded 30s timeout is set on every client constructed by this code instead of relying on an unbounded default
- Audited the rest of the `osac` mono-repo (osac-operator, osac-csi-driver, bare-metal-fulfillment-operator, and the rest of fulfillment-service) for the same pattern — no other instances found; server-side JWKS/API-client code already builds context-aware requests correctly
- Adds regression tests for all three fixed call sites that point their respective endpoint at a handler that never responds and assert the call fails promptly once the context deadline is reached

## Summary by CodeRabbit

- **Bug Fixes**
  - OAuth token requests now honor caller-provided cancellation and deadlines.
  - Requests to unresponsive token endpoints fail promptly instead of hanging indefinitely.
  - Logout and token-refresh requests now stop promptly when cancelled or when deadlines expire.
  - Requests without a deadline use a 30-second timeout.
  - Form requests explicitly provide the expected content type.
  - TLS connections now require TLS 1.3 or later.

### DIFF
--- a/fulfillment-service/internal/cmd/cli/logout/logout_cmd.go
+++ b/fulfillment-service/internal/cmd/cli/logout/logout_cmd.go
@@ -217,16 +217,20 @@ const httpRequestTimeout = 30 * time.Second
 
 func (c *runnerContext) httpClient(caPool *x509.CertPool, insecure bool) *http.Client {
 	tlsConfig := &tls.Config{
-		RootCAs: caPool,
+		RootCAs:    caPool,
+		MinVersion: tls.VersionTLS13,
 	}
 	if insecure {
 		tlsConfig.InsecureSkipVerify = true
 	}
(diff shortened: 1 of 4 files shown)


CrystalChun (approved):
/lgtm

PR #167, opened 2026-08-05
OSAC-3609: restrict public filters to public object fields

wgordon17:
## Summary
- `FilterTranslator` is now descriptor-driven instead of generic over the private proto type, so public-fronted resources configure it with the public descriptor and reject tenant filters referencing private-only fields at CEL compile time instead of leaking them through list-result timing/presence
- Wired the fix through all 29 public-fronted resources (26 from the original archived design + 3 gaps: `secrets`, `cluster_versions`, `bare_metal_instance_types`), and classified filter-translate failures as `InvalidArgument` instead of `Internal`
- Added a reflection-driven regression test that discovers every public/private message pair's private-only fields and asserts rejection automatically for any future resource, plus fixed an unrelated map-typed-field misclassification bug in the same translator

Relates to [OSAC-2819](https://redhat.atlassian.net/browse/OSAC-2819)

## Summary by CodeRabbit

* **New Features**
  * Added centralized registration for public and private resource services.
  * Improved filtering with public resource schemas, including nested and map fields.
* **Bug Fixes**
  * Filters referencing private-only fields now return clear `InvalidArgument` errors instead of internal failures.
  * Improved validation and safety for JSON-based filter expressions.
* **Tests**
  * Added regression coverage for filtering, streaming events, users, and resource server registration.


fulfillment-service/internal/database/dao/filter_translator.go
```diff
@@ -997,6 +1011,14 @@ func (t *FilterTranslator[O]) translateSelectJsonField(operandSql string, msgDes
 		return
 	}
 	fieldDesc := msgDesc.Fields().ByName(protoreflect.Name(fieldName))
+	// Map fields are stored as JSONB objects — use -> to preserve the object structure, so that bracket-index
+	// and 'in' translate against it instead of falling into the generic message-kind case below:
+	if fieldDesc.IsMap() {
+		result.sql = fmt.Sprintf("%s->'%s'", operandSql, fieldName)
+		result.kind = filterTranslatorMapKind
+		result.precedence = filterTranslatorMaxPrecedence
+		return
```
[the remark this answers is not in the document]
wgordon17: Overriding the `No-Injection-Vectors` pre-merge check for this finding (`Ignore` checked in the summary comment).

`fieldName` here only ever originates from a CEL field selector (`expr.FieldName()`); CEL's own identifier grammar restricts it to `[a-zA-Z_][a-zA-Z0-9_]*`, and this codebase's earlier review pass on this same line already conceded "CEL descriptor validation limits this value to schema fields, so this is not currently a direct injection path." That hasn't changed.

I did add identifier quoting here (reusing the same `translateString` escaping `translateIndex` already uses for map bracket-index keys), which the check's first-pass resolution text listed as an acceptable alternative to full parameterization. The re-scan after that change now insists on true bound-parameter SQL instead.

Getting there for real means changing `FilterTranslator.Translate()`'s return contract to carry query arguments alongside the generated SQL text, threading that through all of its `translate*` methods, and updating the DAO's list-query execution path to bind placeholders instead of consuming one fully-formed SQL string — an architectural change to the query-building layer, not something scoped to OSAC-3609 (the public/private filter field boundary). Since the value is provably non-attacker-controlled today, that refactor belongs in its own dedicated change rather than folded into this PR.

wgordon17:
@coderabbitai full review

wgordon17:
@coderabbitai review

wgordon17:
@coderabbitai full review

wgordon17:
@coderabbitai approve

fulfillment-service/internal/database/dao/filter_translator.go
```diff
@@ -176,34 +178,51 @@ var binaryOps = map[string]binaryOpInfo{
...
-	var thisTempl O
-	thisDesc := thisTempl.ProtoReflect().Descriptor()
-
-	var projectTempl *privatev1.Project
-	projectDesc := projectTempl.ProtoReflect().Descriptor()
+	// A message is project-shaped if it is either the public or the private Project message: both configure the
+	// translator with their own descriptor depending on which server (public or private) built it.
+	publicProjectDesc := (*publicv1.Project)(nil).ProtoReflect().Descriptor()
+	privateProjectDesc := (*privatev1.Project)(nil).ProtoReflect().Descriptor()
+	isProject := b.thisDesc == publicProjectDesc || b.thisDesc == privateProjectDesc
```
sk-ilya, line 225: I understand it was already implemented this way, but having a concrete generic field translator inspect the underlying type does not feel right... I'm wondering whether we should instead allow the type of the `name` field to be overridden and set it to `ltree` for `Project`, perhaps smh like `SetNameKind(filterTranslatorLtreeKind)`.


--
(same comment as  https://github.com/osac-project/fulfillment-service/pull/921#discussion_r3610293987)

sk-ilya:
This was initially handled in https://github.com/osac-project/fulfillment-service/pull/921 @jhernand  PTAL

fulfillment-service/internal/database/dao/filter_translator.go
jhernand: The filter translator was initially made generic (using Go generics) only to have a way to pass the type of the subject object. This change removes that, as it isn't really needed. So the filter translator is no longer generic in the sense of Go generics. It is still generic in that it knows how to handle any kind of object, via gRPC reflection. I don't see a problem with that. It needs to know some details about the subject type, like the type of the "name" field, but I don't think that is too bad. Doing as you suggest wouldn't be too bad either, but it would require exposing the `filterTranslatorLtreeKind` constant (and other constants of the type) to the users of this component, which I think isn't good.

jhernand (approved):

jhernand (approved):

PR #169, opened 2026-08-05
NO-ISSUE: Restore e2e-caas-netris-full-install as on-demand only

eliorerz:
## Summary

PR #110 ([OSAC-2801](https://redhat.atlassian.net/browse/OSAC-2801)) added e2e-caas-netris-full-install.yml (provisions an ephemeral EC2 c5n.metal box, runs a full Netris Spectrum-X simulated lab + OCP SNO + OSAC install, then the CaaS suite), auto-triggered on every PR. It started failing on every PR -- including totally unrelated ones -- with `gh: Resource not accessible by personal access token (HTTP 403)` when registering the ephemeral EC2 instance as a JIT GitHub Actions runner. That's a separate, still-open token/permissions problem, not addressed by this PR. #166 reverted the whole workflow rather than leave every PR red.

This restores the original file (`git show 4416ff26:.github/workflows/e2e-caas-netris-full-install.yml`) verbatim, with exactly one substantive change: the `changes` job's `should-run` output now also requires `github.run_attempt != '1'` on `pull_request` events, so the suite is skipped (not run) on a PR's automatic first check run. `schedule`/`workflow_dispatch`/`merge_group` triggers are untouched -- those always run for real. Also adds a log step (pull_request + attempt 1 only) explaining how to actually trigger it.

**Why this works with zero shared-code changes**: `github.run_attempt` is `1` for a workflow's original run and `2+` for each `gh run rerun`. `osac-test-infra`'s existing shared `slash-command-handler.yml` (untouched) already finds an existing PR-triggered run for a given workflow at the PR's current head SHA and reruns it on `/test <workflow>` -- that's exactly attempt 2+. Keeping `on: pull_request` means there's always a run for the handler to find and rerun; no new trigger mechanism, and zero effect on e2e-bmaas/e2e-vmaas/e2e-caas-full-install's own `/test` commands (their own workflows and the shared handler are both unmodified).

## Test plan

- [x] YAML parses, `yamllint --strict`, `actionlint`, full `pre-commit run` all pass
- [ ] This PR's own check list shows `E2E CaaS Netris Full Install` completing quickly as skipped on the initial push (not running, not red) -- proves the JIT-runner-403 path is no longer hit automatically
- [ ] Comment `/test e2e-caas-netris-full-install` on this PR and confirm the existing slash-command handler finds and reruns the same check (not a new one), and that attempt 2 actually starts the real job (reaches EC2 provisioning) -- expected to still hit the separate JIT-runner-403 issue at that point, which is out of scope here; this only proves it's on-demand now, not that the suite passes end-to-end

## Summary by CodeRabbit

* **Tests**
  * Added automated end-to-end coverage for the full-install suite.
  * Tests now run on pull requests, merge queues, scheduled daily runs, and on demand.
  * Added safeguards to report and fail the workflow when test execution is unsuccessful or cancelled.
  * Coverage includes validation of key application components during complete installation scenarios.

### DIFF
--- a/.github/workflows/e2e-caas-netris-full-install.yml
+++ b/.github/workflows/e2e-caas-netris-full-install.yml
@@ -73,6 +73,10 @@ jobs:
           {"repo":"${{ github.event.pull_request.head.repo.full_name || github.repository }}","ref":"${{ github.event.pull_request.head.ref || github.ref_name }}","containerfile":"osac-aap/execution-environment/execution-environment.yaml","imageKey":"aap.bootstrap.image","buildType":"ansible-builder"},
           {"repo":"${{ github.event.pull_request.head.repo.full_name || github.repository }}","ref":"${{ github.event.pull_request.head.ref || github.ref_name }}","containerfile":"osac-aap/execution-environment/execution-environment.yaml","imageKey":"aap.configAsCode.eeImage","buildType":"ansible-builder"}
         ]
+      # Pass author_association for fork PR authorization (empty for non-fork PRs)
+      fork-pr-author-association: ${{ github.event.pull_request.head.repo.fork && github.event.pull_request.author_association || '' }}
+      fork-pr-author: ${{ github.event.pull_request.head.repo.fork && github.event.pull_request.user.login || '' }}
+      pr-number: ${{ github.event.pull_request.number }}
 
   e2e:
     if: always()
--- a/.github/workflows/e2e-caas-netris-full-install.yml
+++ b/.github/workflows/e2e-caas-netris-full-install.yml
@@ -1,0 +1,89 @@
+---
+name: E2E CaaS Netris Full Install
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    branches: [main]
+  merge_group:
+  schedule:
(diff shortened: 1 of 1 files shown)


eliorerz:
Fixed in 7f058e7a — this workflow had zero fork-PR authorization (unlike every other e2e suite here), confirmed by checking the reusable workflow in osac-test-infra directly: no fork-pr-author-association/-author/pr-number inputs, no authorize-fork-pr call at all.

Real fix needed a companion change since the reusable workflow itself had nothing to hook into: opened osac-test-infra#324 to add the same inputs + the same Authorize fork PR gate (mirroring e2e-bmaas/vmaas/caas-full-install exactly, same position -- right after setup, before any Vault secret fetch). This PR's commit passes those inputs from the caller side.

Ordering note: this depends on osac-test-infra#324 merging first. No risk to automatic CI in the meantime since this workflow only runs on-demand (github.run_attempt != '1' gate) -- it would only matter if someone ran /test e2e-caas-netris-full-install before #324 lands, in which case it'd fail clearly on an invalid-input error rather than silently skipping the auth check.

PR #172, opened 2026-08-06
OSAC-3410: Add CaaS lifecycle metering with N+1 per-component billing

omer-vishlitzky:
## Summary

Extends the Metering Service to track CaaS (Kubernetes cluster) lifecycle events with per-component billing. Each cluster generates N+1 Kafka records per heartbeat or lifecycle transition: one for the control plane plus one per distinct worker node set.

### What this PR includes

- **Declarative transition tables**: Both VMaaS and CaaS state machines use `TransitionTable` maps instead of switch statements. The table IS the spec — missing entry = error (fail fast). Adding BMaaS = adding a new table.
- **CaaS state machine**: PROGRESSING/READY billable, FAILED/DELETING/DELETE_FAILED not. `ErrSkipTransition` sentinel for transitions with no billing effect.
- **Transient state handling** (from merged PR #176): STOPPING/STARTING return `ErrTransientState` — projection updates only FulfillmentVersion+TransitionTime, preserving CurrentState and billing context. DELETING emits `suspended.v1` (billing-ending, not transient).
- **Publish-first ordering**: `publishAndUpsert` enforces publish-before-commit by construction. Prevents permanent event loss on partial N+1 Kafka failure. Delete path was already correct.
- **N+1 decomposition at publish time**: 1 projection row per cluster, decomposed into per-component CloudEvents at publish. `created.v1` and `deleted.v1` are single records (audit only).
- **`DecomposeClusterEvents` shared helper**: Extracted from 4 call sites (Watch Consumer, Heartbeat Generator, Reconciler corrections, Reconciler synthetic heartbeats). Cluster-order only, no fallback — empty components = `ErrDataQuality`.
- **Scaling detection**: `ChangedComponents` emits `updated.v1` only for components whose `node_count` differs or were added/removed.
- **Heartbeat N+1**: Heartbeat Generator decomposes cluster_order into per-component heartbeats with flat billing dimensions.
- **Reconciliation**: `ClusterLister` interface, paginated `ListClusters`, `isBillableForType` dispatch, N+1 correction events, N+1 synthetic heartbeats.
- **Deterministic CloudEvent IDs**: `{eventID}/{nodeSet}` for adapter-level dedup on replay.

### Tickets covered

| Ticket | Summary | Coverage |
|--------|---------|----------|
| [OSAC-3421](https://redhat.atlassian.net/browse/OSAC-3421) | CaaS state machine + Watch Consumer | Fully met |
| [OSAC-3433](https://redhat.atlassian.net/browse/OSAC-3433) | CaaS heartbeat + reconciliation | Fully met |
| [OSAC-3438](https://redhat.atlassian.net/browse/OSAC-3438) | CaaS E2E tests | Separate PR (osac-test-infra) |

### Design reference

- **Epic:** [OSAC-3410](https://redhat.atlassian.net/browse/OSAC-3410)
- **Feature:** [OSAC-985](https://redhat.atlassian.net/browse/OSAC-985)
- **Design:** [enhancement-proposals#131](https://github.com/osac-project/enhancement-proposals/pull/131)
- **CaaS events spec:** [masayag/caas-metering-events gist](https://gist.github.com/masayag/7b679cc3c0c4071706a20506f8603b1f)

### Key design decisions

| Decision | Choice | Why |
|----------|--------|-----|
| State machines | Declarative `TransitionTable` maps | Table IS the spec, missing entry = error, no switch defaults to get wrong |
| Publish ordering | Publish-first via `publishAndUpsert` | Prevents permanent event loss on partial N+1 Kafka failure |
| N+1 decomposition | `DecomposeClusterEvents` shared helper | Single implementation, cluster-only, no fallback, empty = error |
| Projection model | 1 row per cluster | Simpler reconciliation; N+1 at publish time |
| Billing dims storage | `components` array in JSONB | Enables DimensionsEqual for scaling detection |
| created/deleted decomposition | Single record | Audit only, no billing effect (per spec) |
| CloudEvent IDs | `{eventID}/{nodeSet}` | Deterministic for adapter dedup on replay |
| Node sets source | `spec.node_sets` | Billing follows desired state, not status |
| Transient states (VMaaS) | STOPPING/STARTING preserve CurrentState | `previous_state=RUNNING` on `suspended.v1`, matching design exactly |
| DELETING (VMaaS) | Emits `suspended.v1` | Billing-ending, not transient — closes interval for direct deletes |
| DELETING (CaaS) | `suspended.v1` only from billable states | Non-billable→DELETING has no interval to close |
| ErrSkipTransition | Sentinel error | Forgotten check = reconnect, not silent data loss |
| Initial non-billable (CaaS) | `""→FAILED/DELETING/etc` = Skip | Prevents crash on bootstrap/reconnect with existing clusters |

### Review fixes included

Fixes from 5-agent critical review + amito's inline comments:

1. **Publish-first ordering** — `publishAndUpsert` enforces publish before commit, prevents event loss on partial N+1 Kafka failure (affected both VMaaS and CaaS)
2. **Missing initial CaaS transitions** — `""→FAILED/DELETING/DELETE_FAILED/UNSPECIFIED` added as Skip (prevents crash on first-observed non-billable cluster)
3. **`ErrSkipNonBillingTransition` renamed to `ErrSkipTransition`** — old name misleading for billable→billable transitions
4. **`DecomposeClusterEvents` shared helper** — extracted from 4 call sites per amito's review; cluster-only, no fallback, empty components = `ErrDataQuality`
5. **Proto enumeration completeness test** — iterates every `(from, to)` state pair, catches removed table entries
6. **DimensionsEqual component ordering test** — documents that order matters (sorted keys for determinism)
7. **Strengthened multi-removal test** — verifies exact NodeSet→HostType preservation

## Test plan

- [x] `go build ./...` passes
- [x] `SKIP_DB_TESTS=1 ginkgo run -r --timeout=1m internal/` — 240 specs pass (7 suites, ~4s)
- [x] Pre-commit (golangci-lint) passes
- [x] VMaaS transition table: 10 entries + wildcard expansions, all tested + unknown state error
- [x] CaaS transition table: 42 entries, 42 test entries (1:1) + unknown state error
- [x] Proto enumeration completeness test: every `(from, to)` state pair covered
- [x] `DecomposeClusterEvents`: happy path, empty components error, buildFn error propagation
- [x] `publishAndUpsert`: stale version test updated for publish-first semantics
- [ ] Deploy to dev cluster with CaaS cluster, verify N+1 events on lifecycle/heartbeat topics

## Summary by CodeRabbit

* **New Features**
  * Added metering support for CaaS clusters, including lifecycle events, billability, metadata, and state transitions.
  * Added component-level billing for control-plane and worker node-set changes, including scaling, additions, and removals.
  * Added cluster reconciliation, correction events, and synthetic heartbeats for all billable components.
* **Bug Fixes**
  * Improved event ordering, projection updates, stale-version handling, and checkpointing when publishing multiple events.
  * Added clearer validation for missing timestamps and unsupported transitions.
* **Tests**
  * Expanded coverage for cluster lifecycle, scaling, reconciliation, corrections, heartbeats, and transition handling.


osac-metering/metering-service/internal/heartbeat/generator.go
```diff
@@ -96,20 +96,22 @@ func (g *Generator) tick(ctx context.Context) error {
...
 		}
-		if err := g.publisher.Publish(ctx, ce); err != nil {
-			if len(publishedIDs) > 0 {
-				if cpErr := g.store.UpdateLastHeartbeat(ctx, publishedIDs, now); cpErr != nil {
-					g.logger.Error(cpErr, "failed to checkpoint partial heartbeat progress",
-						"published", len(publishedIDs))
+		for j := range hbEvents {
+			if err := g.publisher.Publish(ctx, hbEvents[j]); err != nil {
+				if len(publishedIDs) > 0 {
+					if cpErr := g.store.UpdateLastHeartbeat(ctx, publishedIDs, now); cpErr != nil {
```
amito: If we fail mid-publish (_e.g._, 1 of 3 events published), the last ID is not added to `pulishedIDs` and the two events will not be checkpointed. In the next tick all 3 events will be re-published.
This might be OK (some duplication which is deduped in the adapter).

osac-metering/metering-service/internal/watch/consumer.go
```diff
@@ -221,6 +233,125 @@ func (c *Consumer) handleTransientState(
...
+		return c.publishWithRetry(ctx, baseCE)
+	}
+
+	for _, comp := range components {
+		compCE, ceErr := c.buildComponentEvent(baseCE, eventID, comp)
+		if ceErr != nil {
+			return ceErr
+		}
+		if err := c.publishWithRetry(ctx, &compCE); err != nil {
+			return err
```
amito, line 285: Similar event duplication behavior I referred to in the heartbeat generator.

osac-metering/metering-service/internal/watch/consumer.go
```diff
@@ -221,6 +233,125 @@ func (c *Consumer) handleTransientState(
...
+		return nil
+	}
+
+	stateCtx := c.buildStateContext(existing, isBillable, transitionTime, dims)
+	for _, comp := range changed {
+		ce, ceErr := c.buildScalingEvent(event.GetId(), mapper, comp, stateCtx, transitionTime)
+		if ceErr != nil {
+			return ceErr
+		}
+		if err := c.publishWithRetry(ctx, &ce); err != nil {
```
amito: Similar event duplication behavior I referred to in the heartbeat generator.

osac-metering/metering-service/internal/watch/consumer.go
```diff
@@ -221,6 +233,125 @@ func (c *Consumer) handleTransientState(
 	return nil
 }
 
+func (c *Consumer) publishLifecycleEvents(ctx context.Context, baseCE *cloudevents.Event, mapper events.ResourceMapper, eventID string) error {
```
amito, line 272: The pattern in this function (`if ... cluster_order ; DecomposeCluster ... ; build - publish`) repeats here, in `buildCorrectionEvents`, in `buildSyntheticHeartbeats` and in `buildHeartbeatEvents`.
Can use a helper function to reduce duplication and share a common logic.

amito (approved):

masayag (changes requested):
## Code Review: Design Compliance Check

Reviewed against the linked design ([enhancement-proposals#131](https://github.com/osac-project/enhancement-proposals/pull/131)) and the "CaaS events spec" gist referenced in the PR description, at head commit `5244106f`. Cross-checked findings against CodeRabbit's existing review comments on this PR to confirm current status.

### Summary

This PR faithfully implements most of the design: declarative `TransitionTable` maps (with fail-fast on unmapped transitions), publish-before-commit ordering (`publishAndUpsert`), and the shared `DecomposeClusterEvents`/`BuildResourceEvents` abstraction for N+1 per-component billing, wired correctly through the Watch Consumer, Heartbeat Generator, and Reconciler (including `ClustersClient`/`isBillableForType` dispatch in `main.go`). Test coverage is extensive (956-line `cluster_test.go`, transition-table completeness tests, JSONB round-trip edge cases).

However, I found one **critical** correctness gap and one **high** design-compliance gap that I'd recommend blocking merge on, both independently corroborated by CodeRabbit's own still-open comments on this PR.

### Finding 1 — CRITICAL: Silent, permanent loss of billing interval boundaries

`ChangedComponents` in `internal/events/cluster.go` only compares `NodeCount` between old/new component records for a given `NodeSet`:

```go
for _, r := range newRecords {
    newByKey[r.NodeSet] = true
    old, exists := oldByKey[r.NodeSet]
    if !exists || old.NodeCount != r.NodeCount {
        changed = append(changed, r)
    }
}
```

It never fires when only `HostType`, `ClusterTemplate`, or `VersionName` changes for the same node set. `handleScalingEvent` in `internal/watch/consumer.go` then takes the "nothing changed" branch and updates the projection **without publishing any event**.

Two compounding problems:

- This code path is also reachable for VMaaS, where `ChangedComponents` always requires a `"components"` key that VMaaS dims never have — so **every** VMaaS dimension change while `RUNNING` (i.e. a resize) silently takes the no-op branch. The design's VMaaS diagram explicitly annotates `resize → osac.resource.updated.v1` under `RUNNING`; this is never emitted. `consumer_test.go` (around line 653) documents this as intentional ("no CloudEvent published for VMaaS" dimension changes), which contradicts the design.
- `buildProjectionState` still resets `BillableSince` to `now` whenever dimensions differ, even in this dropped-event branch — corrupting subsequent `duration_seconds` billing calculations with no adapter ever notified.
- The hourly Reconciliation Loop **cannot** self-heal this: by the time it runs, the projection already matches fulfillment's current dimensions (the watch consumer updated it in the no-op branch), so no drift is ever detected. This is a permanent gap, not a delayed correction.

CodeRabbit flagged the CaaS half of this independently (`cluster.go:340`) and it's still open. That fix alone won't close the VMaaS half, since `ChangedComponents`/`DecomposeClusterComponents` don't apply to non-decomposed resource types — VMaaS needs its own billing-dimension-changed check independent of the CaaS component model.

### Finding 2 — HIGH: `release_image` vs `version_name` field mismatch

Both the design and the referenced CaaS events spec gist define the per-component field as `release_image`, resolved from the cluster version. The implementation instead stores the raw `spec.GetVersionName()` under the key `version_name` with no resolution step:

```go
if vn := spec.GetVersionName(); vn != "" {
    dims["version_name"] = vn
}
```

The PR's own tests use release-image-shaped values (e.g. `"quay.io/openshift-release-dev/ocp-release:4.17.0-x86_64"`) for this field, so this isn't a typo — the naming was simply never aligned with the documented adapter contract. Any billing adapter built against the published schema would fail to find this dimension.

### Finding 3 — MEDIUM: `created.v1`/`deleted.v1` payload shape diverges

`publishLifecycleEvents` skips decomposition for `created`/`deleted` and publishes the raw nested `billing_dimensions.components: [...]` shape. Both the design and gist show these events as a single record with a **flat** `billing_dimensions: {cluster_template, release_image}` — no nested component list. Still open per CodeRabbit.

### Finding 4 — MEDIUM: `started.v1`/`resumed.v1` inconsistency across resource types

VMaaS sends `FAILED/DELETING/UNSPECIFIED → RUNNING` through `EventStarted`, while CaaS's equivalent transitions (`FAILED/DELETING/DELETE_FAILED/UNSPECIFIED → PROGRESSING/READY`) all use `EventResumed`. The design's VMaaS diagram doesn't explicitly annotate the `FAILED→RUNNING` event type, so this isn't an unambiguous violation, but it's an inconsistency an adapter would need to special-case. Still open per CodeRabbit.

### Finding 5 — LOW: potential nil-map panic

`buildComponentEvent` in `consumer.go` calls `baseCE.DataAs(&baseData)` then immediately writes `baseData["billing_dimensions"] = dims` without a nil check. `DataAs` can return a nil map (not an error) for zero-length data. Not reachable from current call sites, but worth a defensive check. Still open per CodeRabbit.

### Recommendation

Requesting changes on Finding 1 (untested billing-accuracy bug that reconciliation cannot self-heal) and Finding 2 (breaks the documented adapter field contract). Findings 3–5 are reasonable follow-ups but not blockers.

---
*Review performed by an AI coding agent, cross-referencing the PR against enhancement-proposals#131 and validated against CodeRabbit's existing automated review comments.*

PR #174, opened 2026-08-06
OSAC-3428: Implement Provider Adapter framework with Kafka consumer lifecycle

amito:
## Description

Adds a new Go framework package at `osac-metering/adapters/` that handles the full Kafka consumer lifecycle for provider metering adapters. Concrete adapter binaries (M360, Cost Management, etc.) import this package, implement the `ProviderAdapter` interface, and let the framework's `Runner` manage consumption, retries, dedup, and flushing.

**Architecture:**

```
Kafka topics
    │
    ▼
Runner (Sarama ConsumerGroup)
    │
    ├── Dedup cache (CloudEvent ID, 10m TTL)
    ├── Out-of-order tracker (resource_id → last transition_time)
    │
    ▼
ProviderAdapter.Submit(event)
    │
    ├── Success → track offset
    ├── RetryableError → exponential backoff (1s–5m, max 10 attempts)
    └── NonRetryableError → log + skip
    │
    ▼
Flush ticker (default 10s)
    │
    ▼
ProviderAdapter.Flush()
    │
    ├── Success → commit Kafka offsets
    └── Failure → do not commit (redelivery on restart)
```

**Key components:**

| File | Responsibility |
|------|---------------|
| `adapter.go` | `ProviderAdapter` interface, `MeteringEvent`, `SubmitResult`, error types |
| `runner.go` | `Runner` with Kafka consumer lifecycle, flush scheduling, graceful shutdown |
| `retry.go` | Exponential backoff (1s–5m, ±25% jitter, max 10 attempts), error classification |
| `dedup.go` | In-memory CloudEvent ID dedup cache with TTL eviction |
| `order.go` | Out-of-order detection via `transition_time` per `resource_id` |
| `metrics.go` | 8 Prometheus metrics (`osac_adapter_*`), handler for adapter HTTP servers |
| `kafka.go` | Sarama consumer configuration, TLS/SASL/SCRAM helpers |
| `cmd/echo-adapter/` | Echo adapter binary for E2E testing (see below) |

**Echo adapter** (`adapters/cmd/echo-adapter/`):

Replaces the previous `metering-service/cmd/test-adapter/` (which used raw Sarama and did not exercise the framework). The echo adapter implements the full `ProviderAdapter` interface and exercises the complete `Runner` lifecycle. It exposes an HTTP query API for E2E test assertions:

| Endpoint | Description |
|----------|-------------|
| `GET /events` | List events with filters (`type`, `resource_id`, `since`, `limit`) |
| `GET /events/count` | Count matching events |
| `GET /events/{id}` | Lookup single event by CloudEvent ID |
| `DELETE /events` | Clear buffer for test isolation |
| `GET /metrics` | Prometheus metrics |
| `GET /healthz` | Health check |

Configuration: `ECHO_BUFFER_SIZE` (default 1000), `METRICS_ADDR` (default `:2112`), plus standard `KAFKA_*` env vars.

## JIRA Tasks

- [OSAC-3428](https://issues.redhat.com/browse/OSAC-3428) — Implement Provider Adapter framework with Kafka consumer lifecycle
- [OSAC-3763](https://redhat.atlassian.net/browse/OSAC-3763) — Implement Echo Provider Adapter for integration testing
- Parent epic: [OSAC-3411](https://issues.redhat.com/browse/OSAC-3411) — Provider Adapter
- Parent feature: [OSAC-985](https://issues.redhat.com/browse/OSAC-985) — Metering and Usage Tracking

## Not Included / Out of Scope

- **DLQ handling** — Kafka DLQ topic, publisher, error envelope (OSAC-3436, moved to separate epic [OSAC-3666](https://redhat.atlassian.net/browse/OSAC-3666))
- **Concrete adapter implementations** — M360 adapter ([OSAC-3413](https://redhat.atlassian.net/browse/OSAC-3413)), Cost Management adapter
- **E2E / integration tests with real Kafka** — [OSAC-3440](https://redhat.atlassian.net/browse/OSAC-3440)
- **Replay tooling** and `osac_adapter_replay_lag_events` metric

### Unit-testing

68 unit tests using Ginkgo v2 / Gomega, all passing with race detector enabled:

```
$ cd osac-metering && go test ./adapters/... -race -v
ok  github.com/osac-project/osac-metering/adapters  2.137s
```

**Test coverage by component:**

| Test file | Coverage |
|-----------|----------|
| `adapter_test.go` | Error type wrapping/unwrapping, interface contracts |
| `dedup_test.go` | Duplicate suppression, TTL expiry, concurrent safety |
| `order_test.go` | Out-of-order detection, independent resources, TTL eviction |
| `retry_test.go` | Backoff intervals, max attempts, non-retryable skip, context cancellation |
| `metrics_test.go` | Counter increments, histogram observations, HTTP handler serving |
| `runner_test.go` | Full lifecycle: consume → dedup → order check → submit with retry → flush → offset commit; graceful shutdown; error handling; rebalance safety (context cancellation does not lose events) |
| `kafka_test.go` | TLS toggle (enabled/disabled), consumer defaults, broker address parsing |

No real Kafka broker required — tests use mock Sarama consumer group sessions and claims.

### E2E testing

The echo adapter binary (`adapters/cmd/echo-adapter/`) is included in this PR with Helm chart templates and CI image build workflow. E2E integration tests against a development cluster are tracked in [OSAC-3440](https://redhat.atlassian.net/browse/OSAC-3440).

## Review follow-up changes

Changes made in response to review feedback (masayag, CodeRabbit):

**Amended into existing commits:**
- Dedup-suppressed messages now track offsets (prevents unnecessary redelivery on restart)
- `adapter.Close()` is bounded by a 10s deadline in the Runner's shutdown path (addresses unbounded Close concern while preserving `io.Closer` compatibility)
- Context cancellation during retry backoff no longer tracks the offset — the event is redelivered after rebalance instead of being silently lost
- Echo adapter `/healthz` now calls `adapter.HealthCheck()` instead of returning hardcoded 200
- Echo adapter startup log redacts broker addresses (logs count only)
- Echo adapter deployment sets `KAFKA_CONSUMER_GROUP` to match KafkaUser ACL
- Echo adapter values removed from chart defaults (test-only, provided via CI values)
- Adapters Makefile `build-echo-adapter` creates `bin/` directory before build
- Pre-commit hook for adapters triggers on config file changes (`.golangci.yml`, `Makefile`, `go.mod`, `go.sum`)

**New commits:**
- **TLS toggle** — `TLSEnabled` field on `KafkaConfig`; TLS is only configured when explicitly requested. `kafka_test.go` added with unit tests for TLS paths, consumer defaults, and broker address parsing.
- **CI scaffolding** — `Makefile`, `.golangci.yml`, GitHub Actions unit test job (`run-osac-metering-adapters-tests`), and pre-commit lint hook (`osac-metering-adapters-golangci-lint`), matching the sibling `metering-service` module convention.
- **Echo adapter ([OSAC-3763](https://redhat.atlassian.net/browse/OSAC-3763))** — Replaces `metering-service/cmd/test-adapter/` with `adapters/cmd/echo-adapter/` that exercises the full Runner framework. Adds HTTP query API (`/events`, `/events/count`, `/events/{id}`, `DELETE /events`), Containerfile, Helm chart templates, and CI image build workflow (`build-metering-echo-adapter-image.yaml`).

## References

- Design spec: `docs/superpowers/specs/2026-08-05-provider-adapter-framework-design.md`
- Metering EP: [OSAC-985 design](https://github.com/osac-project/enhancement-proposals/tree/main/enhancements/OSAC-985-metering-and-usage-tracking/design.md)
- Implementation plan: `docs/superpowers/plans/2026-08-05-provider-adapter-framework.md`

## Merge criteria

- [ ] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

## Summary by CodeRabbit

* **New Features**
  * Added shared Kafka-based metering event processing with duplicate suppression and out-of-order tracking.
  * Added configurable retries, TLS/SASL authentication, provider health checks, flushing, and offset tracking.
  * Added Prometheus metrics for submissions, retries, duplicates, ordering, lag, and flushing.
  * Added an Echo Adapter with event inspection, filtering, counting, deletion, health, and metrics endpoints.
  * Added optional deployment and OpenShift routing for the Echo Adapter.

* **Tests**
  * Added comprehensive coverage for processing, retries, metrics, deduplication, ordering, and Kafka configuration.


bare-metal-fulfillment-operator/Containerfile
```diff
@@ -15,6 +15,7 @@ COPY --chown=1001:1001 osac-operator/api/go.mod osac-operator/api/go.sum osac-op
 COPY --chown=1001:1001 fulfillment-service/go.mod fulfillment-service/go.sum fulfillment-service/
 COPY --chown=1001:1001 osac-csi-driver/go.mod osac-csi-driver/go.sum osac-csi-driver/
 COPY --chown=1001:1001 osac-metering/metering-service/go.mod osac-metering/metering-service/go.sum osac-metering/metering-service/
+COPY --chown=1001:1001 osac-metering/adapters/go.mod osac-metering/adapters/go.sum osac-metering/adapters/
```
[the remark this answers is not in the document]
amito: Seems like a false-positive. The `COPY` command is same as the lines above it, _go.sum_ is copied for _osac-metering/adapeters/_`.

masayag (changes requested):
## Summary

Solid first cut of the provider-adapter framework — clean interface boundary, 55 passing Ginkgo specs with the race detector on, and sensible error classification (`RetryableError`/`NonRetryableError`). Found a few correctness gaps in the "skip"/rebalance flow, a CI coverage gap for the new module, and some structural suggestions for the M360/Cost Management adapters this is meant to host. Grouped by severity below.

## Correctness (blocking)

1. **Offsets aren't recorded when a message is dedup-suppressed, non-retryable, or exhausts retries** (`runner.go`, `processMessage`). Only the success path writes into `r.offsets`. Since `AutoCommit` is disabled, none of these three "skip" outcomes ever advances the commit watermark unless a later message on the same partition happens to succeed. If the skipped message is the last one on its partition, the consumer will redeliver (and reprocess) it on every future restart or rebalance — dedup and non-retryable "skip" don't actually skip once the process restarts. Recommend recording the offset in all three cases (tagged appropriately in metrics), not just on success.

2. **`Cleanup` doesn't flush before the session is torn down, and `flush()` can drop already-submitted work on a mid-flight rebalance** (`runner.go`, `flush()` / `Cleanup()`). `flush()` snapshots and clears `r.offsets` before checking whether `r.session` is still non-nil; if a rebalance runs `Cleanup` between those two steps, the successfully-submitted events' offsets are wiped without ever being committed. Kafka will redeliver them post-rebalance, and if the adapter's `Flush` isn't strictly idempotent, this risks double-billing. Recommend a best-effort flush+commit inside `Cleanup` before nilling the session, and/or not clearing `r.offsets` until commit is confirmed.

3. **TLS is unconditionally enabled in `newConsumerConfig`/`configureConsumerTLS`** (`kafka.go`), even when `KafkaConfig.TLSCACert` is empty and there's no toggle to run against a plaintext or SASL_PLAINTEXT broker. Any environment using a non-TLS internal Kafka listener will fail to connect at consumer startup. Needs an explicit enable/disable rather than inferring "always on."

## Testing / CI gap (blocking)

4. `osac-metering/adapters/` isn't wired into any CI job. `.github/workflows/unit-tests.yml` only runs `ginkgo` for `fulfillment-service/internal` and `osac-metering/metering-service/internal` — the 55 tests this PR adds won't run in CI. The `osac-metering-golangci-lint` pre-commit hook is also scoped to `^osac-metering/metering-service/.*\.go$` only, so this package isn't linted either. Please add a CI job (and a `Makefile` + `.golangci.yml`, matching the sibling `metering-service/` module's convention) before merge, so regressions here are actually caught going forward.

5. `kafka.go` has no corresponding `kafka_test.go` — the only file in the package without one. This is very likely how #3 slipped through; a basic test asserting `sc.Net.TLS.Enable` reflects the config instead of always being `true` would have caught it.

## Interface members declared but never wired up

- `ProviderAdapter.HealthCheck(ctx) error` is never called by `Runner` — there's a `MetricsHandler()` but no equivalent `/healthz`-style handler. Needed for k8s liveness/readiness probes once adapters are deployed.
- `SubmitResult{ProviderEventID, Idempotent}` returned by `Flush` is discarded (`_, err := r.adapter.Flush(ctx)`). Either consume it (log the provider event ID, metric the idempotent-replay case — which would also make finding #2 observable) or drop it from the interface for now.
- `osac_adapter_consumer_lag_events` is defined in `metrics.go` but never `Set()`.
- `newAdapterMetrics(provider string)` never uses `provider` — every call site re-passes `r.adapter.Name()` via `WithLabelValues` instead.

## Structural suggestions for the M360 / Cost Management adapters this framework will host

Since concrete adapters (OSAC-3413 for M360, Cost Management TBD) and the echo-adapter binary are explicitly out of scope here, worth settling these conventions now so the first real adapter isn't inventing them:

- **Directory layout** — nest concrete adapters under this package rather than adding new top-level `osac-metering/` siblings per provider, so the list doesn't sprawl at the top level as more providers are added:
  ```
  osac-metering/adapters/
  ├── *.go                 (this PR — framework)
  ├── cmd/echo-adapter/     (already named this in the PR description)
  └── providers/
      ├── m360/             (own go.mod, own Containerfile)
      └── costmanagement/   (own go.mod, own Containerfile)
  ```
- **Separate binaries and images per adapter**, not one combined binary running both. Each is its own consumer group with different latency/scaling/failure characteristics, and an unrecovered panic in one provider's client code shouldn't take the other's consumer group down with it. A shared Containerfile template (parameterized by a build arg for which provider subdirectory to build) can avoid duplicating boilerplate across providers without merging them into one image.
- **Extract a shared wire-contract package** (e.g. `osac-metering/schema`, its own `go.work` member) exporting the CloudEvent extension attribute names (`osacresourceid`, `osactenant`, ...), the JSON payload shape (currently the unexported `meteringData` struct in `metering-service/internal/events/mapper.go`, unreachable from here by design), and the topic name constants (currently unexported in `metering-service/internal/kafka/publisher.go`). Today `runner.go`'s `checkOutOfOrder` re-derives part of this contract by hand via an untyped `map[string]interface{}` and a hardcoded `"transition_time"` key — a shared, typed package would give M360/Cost Management adapters compile-time-checked access to every field instead of reverse-engineering JSON from the producer's source. This is the single highest-leverage change for easing adapter contributors' implementation.
- **Add a `README.md`/`doc.go`** to `osac-metering/adapters/` covering: the `ProviderAdapter` contract semantics (at-least-once delivery, what "idempotent" means for `Flush`, when to return `RetryableError` vs `NonRetryableError` vs a bare error), the event schema (once the shared package above exists), required wiring (`MetricsHandler` on `/metrics`, health handler once added), and the module/Containerfile/`go.work` conventions a new adapter needs to follow.

## Minor

- `dedupCache` (`dedup.go`) and `orderTracker` (`order.go`) are two independently hand-rolled copies of the same structure (mutex + `map[string]time.Time` + 30s eviction ticker). Worth collapsing into one generic `ttlCache[V any]` now that the module is on Go 1.26 with generics available — avoids maintaining/fixing the same concurrency-sensitive logic twice.
- No per-call timeout wraps `adapter.Submit`/`Flush`/`HealthCheck` — a slow/hung downstream call blocks `ConsumeClaim`'s message loop indefinitely. Consider a framework-level `context.WithTimeout` rather than relying on every adapter author to self-bound latency.
- The per-`Runner` Prometheus registry (`metrics.go`) only registers the 8 custom collectors — no `NewGoCollector()`/`NewProcessCollector()`, so `/metrics` will be missing standard goroutine/GC/memory metrics for every future adapter binary.

---

Great overall direction — the interface boundary and happy-path test coverage are strong. The two correctness items and the CI gap are what I'd want resolved before merge; the structural suggestions are for alignment before OSAC-3413 (M360) starts, not blockers on this PR itself.

osac-metering/adapters/adapter.go
```diff
@@ -0,0 +1,65 @@
...
+	Name() string
+	// Submit processes a single metering event.
+	Submit(ctx context.Context, event MeteringEvent) error
+	// Flush uploads any buffered events. Called on the flush ticker
+	// (default 10s) and on graceful shutdown.
+	Flush(ctx context.Context) (SubmitResult, error)
+	// HealthCheck verifies connectivity to the provider.
+	HealthCheck(ctx context.Context) error
+	// Close releases resources after the final Flush.
+	Close() error
```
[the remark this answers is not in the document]
amito: The current `Close()` error signature follows `io.Closer`'s interface pattern, which is the standard for resource cleanup. Changing to `Close(ctx context.Context)` error would prevent implementors from satisfying `io.Closer`. The adapter's `Close()` is called after the final `Flush(ctx)` (which already accepts context), so the window for cleanup work is bounded by the shutdown timeout on the flush. If a concrete adapter's `Close()` needs a timeout, it can create its own internal context.

osac-metering/adapters/runner.go
```diff
@@ -0,0 +1,335 @@
...
+		} else if result.Exhausted {
+			r.metrics.eventsFailed.WithLabelValues(provider, "retries_exhausted").Inc()
+			r.logger.Error(result.Err, "dropping event after retries exhausted",
+				"event_id", eventID, "topic", msg.Topic,
+				"partition", msg.Partition, "offset", msg.Offset,
+			)
+		}
+		// Track the offset even for failed events so the consumer does not
+		// redeliver the same poison message on every restart.
+		r.trackOffset(msg)
```
omer-vishlitzky, line 303: **Critical: silent event loss when a rebalance interrupts a retry backoff sleep.**

This `trackOffset` call fires for *any* `result.Err != nil`, but `submitWithRetry` (retry.go:108-114) has a third path besides `NonRetryable`/`Exhausted`: if the backoff sleep between attempts is interrupted, it returns an error with **both flags false**:

```go
if err := sleepFunc(ctx, backoff); err != nil {
    return retryResult{
        Attempts:      attempt + 1,
        TotalDuration: time.Since(start),
        Err:           ctx.Err(),
    }
}
```

`ctx` here is `session.Context()` (threaded through `ConsumeClaim` → `processMessage`), which Sarama cancels on **rebalance** — not just process shutdown. So: the adapter returns a retryable error, we start backing off, a routine rebalance happens (another consumer joins/leaves, a rolling deploy — not a rare event), the sleep returns `ctx.Err()`, neither `NonRetryable` nor `Exhausted` gets set (no metric increments, no error log), and this line still runs and tracks/eventually commits the offset anyway. The event was never successfully delivered to the adapter and is now unrecoverable, with zero trace that it happened.

Suggest only calling `r.trackOffset(msg)` here when `result.NonRetryable || result.Exhausted`, letting the ctx-cancelled case fall through untracked so normal Kafka redelivery (offset never advanced) picks it up on the next session. Might be worth a `runner_test.go` case that cancels `session.Context()` mid-backoff and asserts the offset is *not* tracked — I couldn't find one, and `retry_test.go`'s only cancellation test checks `submitWithRetry` in isolation without checking `NonRetryable`/`Exhausted`.

osac-metering/charts/osac-metering/templates/echo-adapter-deployment.yaml
```diff
@@ -89,18 +89,22 @@ spec:
...
+              value: "true"
             - name: KAFKA_TLS_CA_CERT
               value: /etc/kafka-secrets/ca.crt
             - name: KAFKA_SASL_USERNAME
-              value: osac-metering-test-adapter
+              value: osac-metering-echo-adapter
             - name: KAFKA_SASL_PASSWORD_FILE
               value: /etc/kafka-secrets/password
+            - name: METRICS_ADDR
+              value: ":8080"
```
omer-vishlitzky, line 109: **Critical: this Deployment can't authenticate its own consumer group.**

`cmd/echo-adapter/main.go:103-106` defaults `KAFKA_CONSUMER_GROUP` to `"echo-adapter-smoke-test"` when the env var is unset — and it's not set anywhere in this env list. But `echo-adapter-kafka-user.yaml`'s ACL only authorizes the literal group `osac-metering-echo-adapter`. Deployed as-is (and `vmaas-ci/values.yaml` now sets `echoAdapter.enabled: true`), the pod will fail to join its consumer group and never consume a message — and since `/healthz` is hardcoded to 200 (see other comment), nothing will surface that failure.

The old `test-adapter` this replaces hardcoded its group name directly in source (`sarama.NewConsumerGroup(brokers, "osac-metering-test-adapter", sc)`), matching its ACL by construction, so this specific failure mode is new — making the group configurable via env var without also wiring the chart to set it dropped that invariant.

Suggest adding:
```yaml
            - name: KAFKA_CONSUMER_GROUP
              value: osac-metering-echo-adapter
```
to this env list.

osac-metering/adapters/cmd/echo-adapter/main.go
```diff
@@ -0,0 +1,177 @@
...
+			SASLPassFile: os.Getenv("KAFKA_SASL_PASSWORD_FILE"),
+		},
+	}, logger)
+
+	// Serve metrics, health, and event query endpoints.
+	go func() {
+		mux := http.NewServeMux()
+		mux.Handle("/metrics", runner.MetricsHandler())
+		mux.HandleFunc("/healthz", func(w http.ResponseWriter, _ *http.Request) {
+			w.WriteHeader(http.StatusOK)
```
omer-vishlitzky, line 157: **`/healthz` is hardcoded to 200, never checks the adapter.**

```go
mux.HandleFunc("/healthz", func(w http.ResponseWriter, _ *http.Request) {
    w.WriteHeader(http.StatusOK)
})
```
`ProviderAdapter.HealthCheck(ctx)` is declared in the interface but never called anywhere in the framework. Combined with the consumer-group mismatch flagged on the Deployment: if this pod can never join its consumer group, k8s's liveness/readiness probes (wired to this endpoint) will report it healthy forever. Worth wiring this to `adapter.HealthCheck(ctx)` before other adapters copy the pattern.

omer-vishlitzky (commented):
**PR description is stale relative to actual scope.** The "Not Included / Out of Scope" section still says *"Echo adapter binary (`adapters/cmd/echo-adapter/`) — local smoke-test tool, not included in this PR"*, but it's clearly included in this diff (main.go, store.go, Containerfile, Helm Deployment/Service/Route/KafkaUser, and a dedicated CI image-build workflow all ship here). Per Jira this is tracked separately as OSAC-3763 (created the same day as this PR's final commits), with an explicit "OSAC-3428 must complete first" dependency on this ticket — so bundling the two is reasonable, but the description should be corrected so it doesn't contradict the actual diff for future readers.

osac-metering/adapters/cmd/echo-adapter/main.go
```diff
@@ -0,0 +1,177 @@
...
+func (a *echoAdapter) Name() string { return "echo" }
+
+func (a *echoAdapter) Submit(_ context.Context, event adapters.MeteringEvent) error {
+	fmt.Printf("[SUBMIT] id=%-36s type=%-30s topic=%-30s partition=%d offset=%d\n",
+		event.CloudEvent.ID(),
+		event.CloudEvent.Type(),
+		event.Topic,
+		event.Partition,
+		event.Offset,
+	)
```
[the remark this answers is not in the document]
amito: The echo-adapter uses fmt.Printf for human-readable output. If stdout is broken, there's nothing meaningful to do (`log.Fatal` would also fail). This is an intentional simplicity tradeoff for a test/development tool.

osac-metering/adapters/cmd/echo-adapter/store.go
```diff
@@ -0,0 +1,237 @@
...
+		limit = n
+	}
+
+	events := s.query(eventType, resourceID, since, limit)
+	if events == nil {
+		events = []storedEvent{}
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	json.NewEncoder(w).Encode(events) //nolint:errcheck
```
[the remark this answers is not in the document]
amito: The event store's HTTP handlers use `json.NewEncoder().Encode()`. Encoding errors here would indicate a fundamental serialization bug, not a runtime error worth recovering from. The HTTP connection would be broken anyway.

osac-metering/charts/osac-metering/templates/echo-adapter-deployment.yaml
amito: Thanks. Added in follow-up commit (a70699bb).

osac-metering/adapters/cmd/echo-adapter/main.go
amito: Thanks, fixed in follow-up commit (a70699bb).

osac-metering/adapters/runner.go
amito: Thanks, addressed in follow-up commit (a04c1da6).

masayag (changes requested):
## Re-review — v2 (after TLS toggle, CI wiring, echo-adapter commits)

### Addressed from the first review

Items #3 (TLS always forced on), #4 (no CI coverage), and #5 (no `kafka_test.go`) from my earlier review are fully resolved — good job on wiring the CI job, pre-commit hook, Makefile, `.golangci.yml`, the `KafkaConfig.TLSEnabled` toggle with tests, and the consume-loop error backoff. The echo-adapter lives at `adapters/cmd/echo-adapter/` with a useful HTTP query API for E2E assertions.

The `flush()` nil-session handling (my earlier #2) was also partially fixed — offsets are now retained instead of dropped when the session is nil, so the pure "rebalance wipes offsets" data-loss scenario is no longer possible. Good.

### Critical: new data-loss regression in the offset-tracking fix

The fix for my original #1 ("skip doesn't advance offset") introduced a worse bug. `processMessage` now calls `trackOffset(msg)` unconditionally whenever `result.Err != nil` (L291-293), which correctly covers the `NonRetryable` and `Exhausted` outcomes — but also covers a third path from `submitWithRetry` that is **neither** of those:

```go
// retry.go L108-113
if err := sleepFunc(ctx, backoff); err != nil {
    return retryResult{
        Attempts:      attempt + 1,
        TotalDuration: time.Since(start),
        Err:           ctx.Err(),
    }
}
```

This fires when a rebalance (or shutdown) cancels `session.Context()` while an event is mid-retry backoff sleep. `Submit` never got a definitive answer — the event was never actually delivered to the provider. Yet `trackOffset` commits its offset, silently and permanently dropping the event.

**Impact**: on any ordinary consumer-group rebalance that interrupts a retry, an in-flight event is lost forever with no log line (the `if result.NonRetryable` / `else if result.Exhausted` branches don't fire, and there's no `else` log). The old bug caused harmless redelivery loops; this causes silent data loss.

**Fix**: only call `trackOffset` for the intentional-give-up outcomes. For the ctx-cancelled case, skip `trackOffset` entirely (so the event's offset is never committed and it gets redelivered on the next session — the safe default). Add a V(1)-level log line for observability:

```go
if result.Err != nil {
    if result.NonRetryable {
        r.metrics.eventsFailed.WithLabelValues(provider, "non_retryable").Inc()
        r.logger.Error(result.Err, "dropping non-retryable event", ...)
        r.trackOffset(msg)
    } else if result.Exhausted {
        r.metrics.eventsFailed.WithLabelValues(provider, "retries_exhausted").Inc()
        r.logger.Error(result.Err, "dropping event after retries exhausted", ...)
        r.trackOffset(msg)
    } else {
        // Context cancelled mid-retry — do NOT advance offset;
        // event will be redelivered on the next session.
        r.logger.V(1).Info("submit interrupted by context cancellation, will redeliver",
            "event_id", eventID, "error", result.Err)
    }
    return
}
```

Also needs a test: cancel the context mid-retry and assert `r.offsets` does NOT contain the interrupted message's offset.

### Claimed fixes not present in the code

Two findings from `omer-vishlitzky` received replies saying "fixed in follow-up commit (a70699bb)" — but that commit does not exist in this PR's branch (`gh pr view --json headRefOid` reports HEAD is `97d7fc44`, and `git cat-file` cannot resolve `a70699bb`). Both bugs remain live:

1. **Consumer group ACL mismatch** — `echo-adapter-deployment.yaml` doesn't set `KAFKA_CONSUMER_GROUP`, so `main.go` defaults to `"echo-adapter-smoke-test"`. But `echo-adapter-kafka-user.yaml` only authorizes group `osac-metering-echo-adapter`. The deployed echo-adapter will get a Kafka `GroupAuthorizationException` on startup. Fix: add `- name: KAFKA_CONSUMER_GROUP` / `value: osac-metering-echo-adapter` to the Deployment env list.

2. **`/healthz` never calls `HealthCheck`** — still `w.WriteHeader(http.StatusOK)` unconditionally. For a dev/smoke-test tool this isn't a hard blocker, but liveness/readiness probes are defined in the Deployment referencing this endpoint, so a broken Kafka connection will never surface through health checks. At minimum call `adapter.HealthCheck(ctx)` and return 503 on failure.

### Still open from first review (non-blocking, for follow-up)

- `SubmitResult{ProviderEventID, Idempotent}` returned by `Flush` is still discarded — no logging, no metric.
- `osac_adapter_consumer_lag_events` gauge is still defined but never `Set()`.
- `newAdapterMetrics(provider string)` still never uses `provider`.
- No `README.md`/`doc.go` for adapter contributors yet (noted as follow-up for pre-OSAC-3413 readiness, not a blocker on this PR).

---

The ctx-cancellation data-loss regression is the one item I'd want resolved before merge — the ACL mismatch and missing commit should also be pushed before the echo-adapter can be tested in a cluster. The rest is fine as follow-up.

masayag (changes requested):
## Re-review — v2 (after TLS toggle, CI wiring, echo-adapter commits)

### Addressed from the first review

Items 3 (TLS always forced on), 4 (no CI coverage), and 5 (no `kafka_test.go`) from my earlier review are fully resolved — good job on wiring the CI job, pre-commit hook, Makefile, `.golangci.yml`, the `KafkaConfig.TLSEnabled` toggle with tests, and the consume-loop error backoff. The echo-adapter lives at `adapters/cmd/echo-adapter/` with a useful HTTP query API for E2E assertions.

The `flush()` nil-session handling (my earlier item 2) was also partially fixed — offsets are now retained instead of dropped when the session is nil, so the pure "rebalance wipes offsets" data-loss scenario is no longer possible. Good.

### Critical: new data-loss regression in the offset-tracking fix

The fix for my original item 1 ("skip doesn't advance offset") introduced a worse bug. `processMessage` now calls `trackOffset(msg)` unconditionally whenever `result.Err != nil` (L291-293), which correctly covers the `NonRetryable` and `Exhausted` outcomes — but also covers a third path from `submitWithRetry` that is **neither** of those:

```go
// retry.go L108-113
if err := sleepFunc(ctx, backoff); err != nil {
    return retryResult{
        Attempts:      attempt + 1,
        TotalDuration: time.Since(start),
        Err:           ctx.Err(),
    }
}
```

This fires when a rebalance (or shutdown) cancels `session.Context()` while an event is mid-retry backoff sleep. `Submit` never got a definitive answer — the event was never actually delivered to the provider. Yet `trackOffset` commits its offset, silently and permanently dropping the event.

**Impact**: on any ordinary consumer-group rebalance that interrupts a retry, an in-flight event is lost forever with no log line (the `if result.NonRetryable` / `else if result.Exhausted` branches don't fire, and there's no `else` log). The old bug caused harmless redelivery loops; this causes silent data loss.

**Fix**: only call `trackOffset` for the intentional-give-up outcomes. For the ctx-cancelled case, skip `trackOffset` entirely (so the event's offset is never committed and it gets redelivered on the next session — the safe default). Add a V(1)-level log line for observability:

```go
if result.Err != nil {
    if result.NonRetryable {
        r.metrics.eventsFailed.WithLabelValues(provider, "non_retryable").Inc()
        r.logger.Error(result.Err, "dropping non-retryable event", ...)
        r.trackOffset(msg)
    } else if result.Exhausted {
        r.metrics.eventsFailed.WithLabelValues(provider, "retries_exhausted").Inc()
        r.logger.Error(result.Err, "dropping event after retries exhausted", ...)
        r.trackOffset(msg)
    } else {
        // Context cancelled mid-retry — do NOT advance offset;
        // event will be redelivered on the next session.
        r.logger.V(1).Info("submit interrupted by context cancellation, will redeliver",
            "event_id", eventID, "error", result.Err)
    }
    return
}
```

Also needs a test: cancel the context mid-retry and assert `r.offsets` does NOT contain the interrupted message's offset.

### Claimed fixes not present in the code

Two findings from `omer-vishlitzky` received replies saying "fixed in follow-up commit (a70699bb)" — but that commit does not exist in this PR's branch (`gh pr view --json headRefOid` reports HEAD is `97d7fc44`, and `git cat-file` cannot resolve `a70699bb`). Both bugs remain live:

1. **Consumer group ACL mismatch** — `echo-adapter-deployment.yaml` doesn't set `KAFKA_CONSUMER_GROUP`, so `main.go` defaults to `"echo-adapter-smoke-test"`. But `echo-adapter-kafka-user.yaml` only authorizes group `osac-metering-echo-adapter`. The deployed echo-adapter will get a Kafka `GroupAuthorizationException` on startup. Fix: add `- name: KAFKA_CONSUMER_GROUP` / `value: osac-metering-echo-adapter` to the Deployment env list.

2. **`/healthz` never calls `HealthCheck`** — still `w.WriteHeader(http.StatusOK)` unconditionally. For a dev/smoke-test tool this isn't a hard blocker, but liveness/readiness probes are defined in the Deployment referencing this endpoint, so a broken Kafka connection will never surface through health checks. At minimum call `adapter.HealthCheck(ctx)` and return 503 on failure.

### Still open from first review (non-blocking, for follow-up)

- `SubmitResult{ProviderEventID, Idempotent}` returned by `Flush` is still discarded — no logging, no metric.
- `osac_adapter_consumer_lag_events` gauge is still defined but never `Set()`.
- `newAdapterMetrics(provider string)` still never uses `provider`.
- No `README.md`/`doc.go` for adapter contributors yet (noted as follow-up for pre-OSAC-3413 readiness, not a blocker on this PR).

---

The ctx-cancellation data-loss regression is the one item I'd want resolved before merge — the ACL mismatch and missing commit should also be pushed before the echo-adapter can be tested in a cluster. The rest is fine as follow-up.

masayag (approved):
## Re-review — v3 (current HEAD d554f6d0)

All critical and major findings from my previous reviews and from omer-vishlitzky's comments are resolved in this version. The branch was rebased so the referenced commit SHAs in earlier replies no longer exist, but the fixes are confirmed present in the current HEAD.

### Resolved

| Finding | Fix |
|---------|-----|
| ctx-cancellation data-loss regression | `runner.go` L290-300: `else` branch skips `trackOffset` + V(1) log; test added in `runner_test.go` |
| Consumer group ACL mismatch | `echo-adapter-deployment.yaml` now sets `KAFKA_CONSUMER_GROUP=osac-metering-echo-adapter` |
| `/healthz` hardcoded to 200 | `main.go` now calls `adapter.HealthCheck(r.Context())`, returns 503 on failure |
| Broker addresses logged in clear text | Now logs `broker_count=%d` instead of actual addresses |
| `mkdir -p bin` missing in Makefile | Fixed |
| pre-commit hook file pattern too narrow | Now includes `.golangci.yml`, `Makefile`, `go.{mod,sum}` |
| `values.yaml` echo adapter defaults | Commented out in base values; environment-specific `vmaas-ci/values.yaml` provides the real config |

### Minor suggestions for follow-up (non-blocking)

1. **`newAdapterMetrics(provider string)` — unused parameter** (`metrics.go` L31). The `provider` arg is never referenced in the function body. Either remove it or use it in metric Help strings.

2. **`consumerLag` gauge registered but never Set()** (`metrics.go` L56-59). It appears as a permanent zero in `/metrics` output — either wire it from `claim.HighWaterMarkOffset() - msg.Offset` in `ConsumeClaim`, or drop it until a follow-up implements it.

3. **`SubmitResult` from `Flush()` discarded** (`runner.go` L174). The echo-adapter returns `{Idempotent: true}` but the Runner ignores it. Worth at least a debug-level log for future adapters that return a `ProviderEventID`.

4. **`echoAdapter.HealthCheck` is a no-op** (`main.go` L80) — always returns nil, so probes will always report healthy even if Kafka is unreachable. Fine for a smoke-test binary, but add a comment so real adapters don't copy the pattern.

5. **`RetryableError` exported but functionally unused** (`adapter.go` L48-52). The framework retries ALL non-`NonRetryableError` errors, so wrapping in `RetryableError` is optional. Document this in a `doc.go` or inline comment to avoid confusion for future adapter contributors.

6. **No contributor README yet.** A `doc.go` or `README.md` describing the Submit/Flush contract, error classification, and how to wire a new adapter binary would lower the bar for M360 and Cost Management implementations.

---

The framework is solid — offset tracking, dedup, retry, graceful shutdown, and the echo-adapter reference are all working correctly. Approving with the above as follow-up suggestions.

PR #177, opened 2026-08-06
OSAC-2784: add get and delete CLI subcommands for Hub resource

alosadagrande:
## Summary

- Add dedicated `osac get hub` and `osac delete hub` CLI subcommands
- Hub only exists in the private API (`osac.private.v1.Hubs`), so the generic reflection-based get/delete require `--private`. These dedicated subcommands use `privatev1.NewHubsClient` directly, matching the existing `osac create hub` pattern
- `get hub` supports listing all hubs or showing detail for a specific hub by ID/name
- `delete hub` resolves by ID or name via `lookup.Find` before deleting

## Result

```console
$ osac get hub
ID    NAME  NAMESPACE  REGISTERED
hub0  -     osac       2026-07-22T07:18:05Z
$ osac get hub hub0
ID:         hub0
Name:       -
Namespace:  osac
Creator:    system
Created:    2026-07-22T07:18:05Z

$ osac delete hub hub0
Deleted hub 'hub0'.

$ osac get hub
No hubs found.

$ osac create hub --id hub0 --kubeconfig /home/alosadag/Documents/CNF/hub0/kubeconfig --namespace osac
Created hub `hub0`.

$ osac get hub hub0
ID:         hub0
Name:       -
Namespace:  osac
Creator:    system
Created:    2026-08-06T09:58:25Z
```

## Jira

https://issues.redhat.com/browse/OSAC-2784

## Test plan

- [x] `osac get hub` lists all registered hubs
- [x] `osac get hub hub0` shows hub detail
- [x] `osac delete hub hub0` deletes the hub
- [x] `osac get hub` confirms deletion (no hubs found)
- [x] `osac create hub --id hub0 ...` re-registers the hub
- [x] `osac get hub hub0` confirms re-registration
- [x] `go build ./...` passes
- [ ] Unit tests pass (`ginkgo run -r internal`)

> **Note:** The server-side `Delete` has no protection against deleting a hub with active workloads (clusters, compute instances). This is a pre-existing gap — a follow-up task should add a `referenceChecker` guard similar to catalog items.

## Summary by CodeRabbit

* **New Features**
  * Added CLI commands to list and view fulfillment hubs by ID or name.
  * Hub details now display metadata, namespace, creator, and timestamps in a table.
  * Added support for both singular and plural hub retrieval commands.
  * Added a CLI command to delete hubs by ID or name.
  * Added clear handling for configuration, lookup, API, and empty-result errors.

* **Tests**
  * Added coverage for hub command registration, aliases, and expected CLI subcommands.


fulfillment-service/internal/cmd/cli/delete/hub/delete_hub_cmd.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+		return resp.GetItems(), nil
+	})
+	if err != nil {
+		return err
+	}
+
+	_, err = client.Delete(ctx, privatev1.HubsDeleteRequest_builder{
+		Id: hub.GetId(),
+	}.Build())
```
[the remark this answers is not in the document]
alosadagrande: Agreed — the server-side `PrivateHubsServer.Delete` delegates directly to `GenericServer.Delete` without a `referenceChecker` guard (unlike e.g. catalog items). This is a real gap, but it's a server-side fix, not a CLI concern. This PR only adds the CLI subcommands to match the existing `create hub` UX.

Filed as a follow-up task to add a `referenceChecker` to the hub server that blocks deletion when active ClusterOrders or ComputeInstances reference the hub.

fulfillment-service/internal/cmd/cli/get/hub/get_hub_cmd.go
```diff
@@ -0,0 +1,139 @@
...
+		Long:    "List all registered hubs, or display details for a specific hub by ID or name.",
+		Example: `  # List all hubs
+  osac get hub
+
+  # Get a specific hub by ID
+  osac get hub hub0`,
+		Args: cobra.MaximumNArgs(1),
+		RunE: runner.run,
+	}
+	return result
```
tzvatot, line 69: 🟡 **Missing `--output` / `-o` flag for structured output**

Registering this subcommand shadows the generic reflection-based `get` path for "hub", so users lose access to JSON/YAML output via `osac get hub -o json`. This matters for scripting and automation.

Consider adding an `--output` flag (matching the generic `get` command's `-o table|json|yaml`) or document that users should use the full proto name with `--private` to get structured output.

fulfillment-service/internal/cmd/cli/get/hub/get_hub_cmd.go
```diff
@@ -0,0 +1,139 @@
...
+	conn, err := cfg.Connect(ctx, cmd.Flags())
+	if err != nil {
+		return fmt.Errorf("failed to create gRPC connection: %w", err)
+	}
+	defer conn.Close()
+
+	client := privatev1.NewHubsClient(conn)
+
+	if len(args) == 0 {
+		resp, err := client.List(ctx, privatev1.HubsListRequest_builder{}.Build())
```
tzvatot, line 106: 💡 **No pagination on hub list**

`client.List()` is called without `Limit`/`PageToken`. While hub counts are typically low (infrastructure-level resource), `get externalippool` sets the precedent with proper page-size and page-token iteration. Minor, but worth aligning for consistency.

fulfillment-service/internal/cmd/cli/get/hub/get_hub_cmd.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/osac-project/osac/fulfillment-service/internal/cmd/cli/lookup"
+	"github.com/osac-project/osac/fulfillment-service/internal/config"
+	"github.com/osac-project/osac/fulfillment-service/internal/terminal"
+)
+
+func Cmd() *cobra.Command {
+	runner := &runnerContext{}
+	result := &cobra.Command{
+		Use:     "hub [ID|NAME]",
+		Aliases: []string{"hubs"},
```
tzvatot, line 47: 💡 **Missing proto message name alias**

`create hub` uses `proto.MessageName((*privatev1.Hub)(nil))` as an alias, enabling `osac create osac.private.v1.Hub`. This command only has `"hubs"`. Minor inconsistency - `get externalippool` also only uses the plural, so both patterns exist in the codebase.

fulfillment-service/internal/cmd/cli/get/hub/get_hub_cmd.go
```diff
@@ -0,0 +1,139 @@
...
+		creator = v
+	}
+
+	fmt.Fprintf(writer, "ID:\t%s\n", h.GetId())
+	fmt.Fprintf(writer, "Name:\t%s\n", name)
+	fmt.Fprintf(writer, "Namespace:\t%s\n", h.GetSpec().GetNamespace())
+	fmt.Fprintf(writer, "Creator:\t%s\n", creator)
+	fmt.Fprintf(writer, "Created:\t%s\n", created)
+	return writer.Flush()
+}
```
tzvatot: 💡 **Detail view could show more hub fields**

Currently shows ID, Name, Namespace, Creator, Created. If the Hub proto has status or health fields (registration state, connected status, managed cluster count), including them here would improve operational visibility without requiring API introspection. Other describe commands show status/conditions when available.

tzvatot (changes requested):
Clean implementation that follows established CLI patterns from `get externalippool` and `create hub`. Good use of `lookup.Find` for name/ID resolution. Well-scoped PR.

| Category | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟡 Important | 2 |
| 💡 Suggestion | 3 |

### 🟡 No unit tests

The PR adds ~225 lines of new code with zero tests, and the PR checklist confirms "Unit tests pass" is unchecked. Other subcommand packages have registration tests (`create_cmd_test.go` verifies aliases and subcommand names, `describe_cmd_test.go` does the same). At minimum, add registration tests to verify `hub` is registered under `get` and `delete`, and that `"hubs"` is a valid alias.

jhernand:
@alosadagrande the hub is only in the private AI on purpose. Why do you think that it should be available directly with a custom command? If we do that it will fail anyhow for non-admin users, as they won't have permission to use the private API.

alosadagrande:
> @alosadagrande the hub is only in the private AI on purpose. Why do you think that it should be available directly with a custom command? If we do that it will fail anyhow for non-admin users, as they won't have permission to use the private API.

osac create hub already exists as a dedicated command using the private API directly, this PR just completes the CRUD surface (get + delete) for consistency.

fulfillment-service/internal/cmd/cli/get/hub/get_hub_cmd.go
alosadagrande: Good point. Notice that `get externalippool` — the only other dedicated subcommand under `get` — also does not have `--output` support. So both are consistent today.

That being said, I agree it is a valid improvement. I will track it as a follow-up to add `-o json/yaml` to all dedicated `get` subcommands (hub + externalippool) rather than doing it piecemeal here.

fulfillment-service/internal/cmd/cli/get/hub/get_hub_cmd.go
alosadagrande: Hubs are infrastructure-level resources — in practice there are 1-5 per deployment, never hundreds. The `lookup.Find` path (line 83-86) does use `Limit`/`Filter` for resolution, so only the list-all path skips pagination.

Notice that `get externalippool` follows the same pattern: no pagination on list-all. For the expected cardinality of hubs this is fine as-is.

fulfillment-service/internal/cmd/cli/get/hub/get_hub_cmd.go
alosadagrande: The proto message name alias in \`create hub\` exists because the generic \`create\` command resolves resource types by proto name. The \`get\` path does not use that mechanism — \`get externalippool\` also has no proto alias, just the plural form.

No user is going to type \`osac get osac.private.v1.Hub\`, so I do not see value in adding it here.

fulfillment-service/internal/cmd/cli/get/hub/get_hub_cmd.go
alosadagrande: I checked the proto definition — \`HubStatus\` is currently an empty message with no fields. There is no status, health, or connected state to display today.

Once the proto gets status fields we can extend the detail view, but for now we are showing everything that exists: ID, Name, Namespace, Creator, Created.

alosadagrande:
@tzvatot Thanks for the thorough review. Regarding the unit tests — agreed, I will add registration tests for `hub` under both `get` and `delete`, following the same pattern as `create_cmd_test.go` (verify subcommand names and aliases). Will push that in the next iteration.

For the other suggestions: replied inline with details on each one. In short, the implementation is consistent with `get externalippool` (the closest existing precedent for dedicated `get` subcommands). The `--output` flag is a valid improvement that I will track as a follow-up for all dedicated `get` subcommands.

jhernand:
@alosadagrande the `osac create hub` command should only be usable/visible for admnistrators. Currently it is visible (in `osac` help) because we didn't find a way (yet) to hide it only for users that aren't administrators. If you add custom commands for get and delete then we will have the same problem for those new commands, therefore we will be making the problem bigger. Instead of doing this can you explore if there is a way to hide the `create hub` command only for non-admin users?

alosadagrande:
@jhernand I agree the visibility concern is valid, but I think it's a separate problem from this PR.

**create hub** already exists as a dedicated command using the private API, same as _create instancetype_, _create clusterversion, create storagebackend _and_ create storagetier_. This PR just completes the CRUD surface for consistency on hubs. Security-wise, nothing changes; the server returns PermissionDenied for non-admins regardless.

Hiding admin commands from non-admin users would need to cover all of them. I can create a follow-up ticket to explore the best approach (e.g. cobra.Command.Hidden with an admin flag, or grouping under osac admin).

jhernand:
@alosadagrande what I am trying to convey is that there are already `osac get hub` and `osac delete hub` commands: they are part of the generic `get` and `delete` commands, and only visible for admin users together with the `--private` flag. What is missing is hiding the non-generic `osac create hub` command when the `--private` flag is not specified.

tzvatot:
All my previous findings are addressed - unit tests added, `-o` flag implemented, and the suggestions were reasonably explained. Thanks for the updates.

Waiting on the discussion with @jhernand about whether dedicated hub subcommands are the right approach vs. the existing generic `--private` path before approving.

alosadagrande:
@jhernand You are right. After looking into this more carefully, I realize that `osac get` and `osac delete` already support Hub through the generic reflection path when the user has logged in with `--private`. The dedicated subcommands are not needed for get/delete since they don't require any special handling like `create hub` does (reading the kubeconfig file from disk).

I will close this PR. The only remaining problem is the visibility of `create hub` for non-admin users — I can open a follow-up ticket to explore how to hide it when `--private` is not enabled.

PR #181, opened 2026-08-06
OSAC-2164: ClusterVersion CLI and ResolveFieldPath reuse

sk-ilya:
- Add ClusterVersion support to the CLI: VERSION column in cluster list tables with lookup_field-based resolution, and a nested Version section in describe cluster output
- Reuse generic ResolveFieldPath[T] across four hand-rolled proto field walkers in auth, references, and reflection packages

## Summary by CodeRabbit

* **New Features**
  * Added a **VERSION** column to cluster tables, displaying the associated version name.
  * Cluster details now show version, status, deprecation, and obsolescence information when available.
* **Bug Fixes**
  * Cluster details continue rendering when version information cannot be retrieved, with a warning instead of failing.
  * Version display falls back gracefully when data or names are unavailable.
* **Improvements**
  * Enhanced table lookups for displaying related resource values more reliably.

### DIFF
--- a/fulfillment-service/internal/cmd/cli/describe/cluster/describe_cluster.go
+++ b/fulfillment-service/internal/cmd/cli/describe/cluster/describe_cluster.go
@@ -90,6 +90,9 @@ func (c *runnerContext) run(cmd *cobra.Command, args []string) error {
 		versionsClient := publicv1.NewClusterVersionsClient(conn)
 		version, err = lookup.Find(versionName, "cluster version",
 			func(filter string, limit int32) ([]*publicv1.ClusterVersion, error) {
+				// Override the public-API default filter that hides obsolete versions.
+				// A cluster may reference any version regardless of lifecycle state.
+				filter = fmt.Sprintf("(%s) && this.spec.state > 0", filter)
 				resp, err := versionsClient.List(ctx, publicv1.ClusterVersionsListRequest_builder{
 					Filter: proto.String(filter),
 					Limit:  proto.Int32(limit),
--- a/fulfillment-service/internal/cmd/cli/describe/cluster/describe_cluster.go
+++ b/fulfillment-service/internal/cmd/cli/describe/cluster/describe_cluster.go
@@ -18,6 +18,7 @@ import (
 	"io"
 	"strings"
 	"text/tabwriter"
+	"time"
 
 	"github.com/spf13/cobra"
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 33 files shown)


fulfillment-service/internal/cmd/cli/describe/cluster/describe_cluster.go
```diff
@@ -81,12 +82,36 @@ func (c *runnerContext) run(cmd *cobra.Command, args []string) error {
...
+		version, err = lookup.Find(versionName, "cluster version",
+			func(filter string, limit int32) ([]*publicv1.ClusterVersion, error) {
+				resp, err := versionsClient.List(ctx, publicv1.ClusterVersionsListRequest_builder{
+					Filter: proto.String(filter),
+					Limit:  proto.Int32(limit),
+				}.Build())
+				if err != nil {
+					return nil, fmt.Errorf("failed to fetch cluster version: %w", err)
+				}
+				return resp.GetItems(), nil
```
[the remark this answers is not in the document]
sk-ilya: addressed

fulfillment-service/internal/rendering/tables/osac.public.v1.Cluster.yaml
```diff
@@ -27,6 +27,12 @@ columns:
   type: osac.public.v1.ClusterTemplate
   lookup: true
 
+- header: VERSION
+  value: "has(this.spec.version)? this.spec.version.name: ''"
+  type: osac.public.v1.ClusterVersion
+  lookup: true
+  lookup_field: spec.version
```
jhernand: I'd prefer to not introduce new fields in table definitions. Can we instead add a "lookup" function to the CEL environment that takes as parameter the reference ("this.spec.version" should be a reference) and returns the corresponding object? If we do that then this can be `lookup(this.spec.version).spec.version`.

sk-ilya: Good idea. But the `lookup` bool isn't a new field and is already widely used. Wouldn't adding the function make it redundant?

jhernand: The `lookup` bool can be removed if add the function, I think.

sk-ilya: Updated - PTAL

jhernand (approved):

sk-ilya:
@jhernand I had to rebase, could you please approve once again?

jhernand (approved):

PR #182, opened 2026-08-06
OSAC-2075: Add --network-attachment flag to BareMetalInstance CLI create command

danmanor:
## [OSAC-2075](https://redhat.atlassian.net/browse/OSAC-2075): Add --network-attachment flag to BareMetalInstance CLI create command

**Jira:** https://redhat.atlassian.net/browse/OSAC-2075
**Epic:** [OSAC-1448](https://redhat.atlassian.net/browse/OSAC-1448) — BareMetalInstance Network Attachments

### Summary

Adds a `--network-attachment` flag to `osac create baremetalinstance`, allowing tenants to specify per-NIC network attachments with subnet, interface, security-groups, and primary designation. This follows the same pattern as the existing ComputeInstance `--network-attachment` flag, extended with the `interface` and `primary` fields specific to `BareMetalNetworkAttachment`.

### Changes

- **Flag registration:** Added `--network-attachment` as a `StringArrayVar` (commas inside values are not treated as delimiters)
- **Parser:** `parseBareMetalNetworkAttachmentFlag` parses one flag value into a `BareMetalNetworkAttachment` proto message, supporting:
  - Bare subnet ID: `sub-123`
  - Key=value form: `subnet=sub-123,interface=data-0,primary,security-groups=sg1,sg2`
  - Both `security-groups=` and `security_groups=` aliases
  - Order-independent key parsing
- **Wiring:** `applyNetworkingFlags` collects parsed attachments and sets `spec.NetworkAttachments` before `fieldutil.ApplyFields`

### Usage

```bash
# Single NIC
osac create baremetalinstance --catalog-item gpu-node \
  --network-attachment interface=data-0,subnet=my-subnet,security-groups=my-sg

# Multi-NIC (exactly one must be primary)
osac create baremetalinstance --catalog-item gpu-node \
  --network-attachment interface=data-0,subnet=data-subnet,security-groups=my-sg,primary \
  --network-attachment interface=data-1,subnet=storage-subnet
```

### Testing

- **Unit tests:** 20 Ginkgo specs — table-driven tests for `parseBareMetalNetworkAttachmentFlag` (9 valid, 7 invalid inputs), `applyNetworkingFlags` (3 specs), flag registration (1 spec)
- **Integration tests:** N/A — CLI-side parsing only; server-side validation covered by [OSAC-1509](https://redhat.atlassian.net/browse/OSAC-1509)
- **Coverage:** 98-100% on new code (`parseBareMetalNetworkAttachmentFlag` 98%, `applyNetworkingFlags` 100%, `extractSecurityGroupListSuffix` 100%)

### Acceptance Criteria

- [x] AC-1: `osac create baremetalinstance` accepts `--network-attachment` flag
- [x] AC-2: Flag parses `interface`, `subnet`, `security-groups`, `primary` fields
- [x] AC-3: Multiple `--network-attachment` flags for multi-homed BM instances
- [x] AC-4: Parsed attachments set on `BareMetalInstanceSpec.network_attachments`
- [x] AC-5: Format matches EP examples

### DIFF
--- a/fulfillment-service/internal/cmd/cli/create/baremetalinstance/create_baremetal_instance_cmd.go
+++ b/fulfillment-service/internal/cmd/cli/create/baremetalinstance/create_baremetal_instance_cmd.go
@@ -91,6 +91,12 @@ func Cmd() *cobra.Command {
 		externalIPAttachmentFlagHelp,
 	)
 	flags.StringArrayVar(
+		&runner.args.networkAttachments,
+		"network-attachment",
+		nil,
+		networkAttachmentFlagHelp,
+	)
+	flags.StringArrayVar(
 		&runner.args.setFields,
 		"set",
 		nil,
@@ -108,6 +114,7 @@ type runnerContext struct {
 		name                 string
 		catalogItem          string
 		setFields            []string
+		networkAttachments   []string
 		sshKey               string
 		userData             string
 		runStrategy          string
@@ -180,6 +187,10 @@ func (c *runnerContext) run(cmd *cobra.Command, _ []string) error {
 	}
 	spec.AutoExternalIpAttachment = c.args.externalIPAttachment
 
+	if err := c.applyNetworkingFlags(&spec); err != nil {
+		return err
+	}
+
 	builtSpec := spec.Build()
 	if err := fieldutil.ApplyFields(builtSpec, c.args.setFields); err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


fulfillment-service/internal/cmd/cli/create/baremetalinstance/create_baremetal_instance_cmd.go
```diff
@@ -248,9 +259,137 @@ creates an ExternalIP with an ExternalIPAttachment for this instance
...
+				id = strings.TrimSpace(id)
+				if id != "" {
+					groups = append(groups, id)
+				}
+			}
+			return prefix, groups, true
+		}
+	}
+	return s, nil, false
+}
```
ori-amizur: **Code duplication: `extractSecurityGroupListSuffix` is copy-pasted from ComputeInstance**

This function is identical to `computeinstance/create_compute_instance_cmd.go:789`. Having two copies means bug fixes or behavior changes need to be applied in both places. Extract it to a shared package.

fulfillment-service/internal/cmd/cli/create/baremetalinstance/create_baremetal_instance_cmd.go
```diff
@@ -248,9 +259,137 @@ creates an ExternalIP with an ExternalIPAttachment for this instance
...
+	spec.NetworkAttachments = attachments
+	return nil
+}
+
+func extractSecurityGroupListSuffix(s string) (prefix string, groups []string, ok bool) {
+	lower := strings.ToLower(s)
+	for _, marker := range []string{"security-groups=", "security_groups="} {
+		if i := strings.Index(lower, marker); i >= 0 {
+			prefix = strings.TrimSpace(strings.TrimSuffix(lower[:i], ","))
+			rest := strings.TrimSpace(lower[i+len(marker):])
```
ori-amizur: **Case inconsistency: values get lowercased only when security groups are present**

`extractSecurityGroupListSuffix` works on a lowered copy, so when security groups are found the returned `prefix` is all lowercase. But when there are none, it returns the original string unchanged (line 310: `return s, nil, false`).

This means:
- `subnet=S1,interface=Eth0` → preserves case ("S1", "Eth0")
- `subnet=S1,interface=Eth0,security-groups=g1` → lowercases everything ("s1", "eth0")

For ComputeInstance this only affected subnet IDs (likely case-insensitive UUIDs). But here it also affects the `interface` field, where case could matter for physical NIC names.

The fix would be to use the original string's byte offsets rather than the lowered copy's when extracting the prefix and group values. This is a pre-existing bug in `extractSecurityGroupListSuffix` — worth fixing in both places when extracting the shared function.

ori-amizur (changes requested):
Two issues to address.

fulfillment-service/internal/cmd/cli/create/baremetalinstance/create_baremetal_instance_cmd.go
danmanor: Fixed in 7ce4d379 — extracted to `netutil.ExtractSecurityGroupListSuffix`. Both `baremetalinstance` and `computeinstance` now import from the shared package.

fulfillment-service/internal/cmd/cli/create/baremetalinstance/create_baremetal_instance_cmd.go
danmanor: Fixed in 7ce4d379 — the shared `netutil.ExtractSecurityGroupListSuffix` now uses original-string byte offsets (`s[:i]` instead of `lower[:i]`) so case is always preserved regardless of whether security groups are present. This also fixes the pre-existing bug in the ComputeInstance version.

fulfillment-service/internal/cmd/cli/create/baremetalinstance/create_baremetal_instance_cmd.go
```diff
@@ -248,9 +260,119 @@ creates an ExternalIP with an ExternalIPAttachment for this instance
...
+		attachments = append(attachments, na)
+	}
+	spec.NetworkAttachments = attachments
+	return nil
+}
+
+func parseBareMetalNetworkAttachmentFlag(s string) (*publicv1.BareMetalNetworkAttachment, error) {
+	s = strings.TrimSpace(s)
+	if s == "" {
+		return nil, fmt.Errorf("empty --network-attachment value")
```
ybettan, line 299: The bare subnet path goes through `ExtractSecurityGroupListSuffix` which lowercases the prefix, so a bare value like `Sub-123` is stored as `sub-123`. But in the key=value path below, the subnet value is taken from the raw input and preserves case. This means `Sub-123` and `subnet=Sub-123` produce different results.

fulfillment-service/internal/cmd/cli/create/netutil/parse.go
```diff
@@ -0,0 +1,41 @@
...
+			prefix = strings.TrimSpace(strings.TrimSuffix(s[:i], ","))
+			rest := strings.TrimSpace(s[i+len(marker):])
+			for _, id := range strings.Split(rest, ",") {
+				id = strings.TrimSpace(id)
+				if id != "" {
+					groups = append(groups, id)
+				}
+			}
+			return prefix, groups, true
+		}
```
ybettan, line 38: The old `extractSecurityGroupListSuffix` in `computeinstance` lowercased the prefix and groups. This new shared version preserves original case — which changes behavior for the VmaaS (ComputeInstance) flow too since it now imports this. Is this intentional?

fulfillment-service/internal/cmd/cli/create/baremetalinstance/create_baremetal_instance_cmd.go
danmanor: The shared `ExtractSecurityGroupListSuffix` (7ce4d379) now uses `s[:i]` instead of `lower[:i]`, so the prefix preserves original case in all paths. A bare `Sub-123` returns `prefix="Sub-123"` (line 40: `return s, nil, false`), and `subnet=Sub-123` also preserves the value through the key=value path. Both produce the same result now.

fulfillment-service/internal/cmd/cli/create/netutil/parse.go
danmanor: Yes, intentional. The old version had a bug where `subnet=Sub-1,security-groups=g1` would lowercase the subnet to `sub-1`, but `subnet=Sub-1` (no groups) would keep it as `Sub-1`. The fix makes both paths consistent — always preserve original case. In practice, OSAC resource IDs are UUIDs (case-insensitive), so this shouldn't cause any behavioral change for existing users. The server handles ID lookups case-insensitively.

ybettan:
/lgtm
/hold
@ori-amizur Can unhold when he is happy with the changes too

danmanor:
@ybettan @ori-amizur  mentioned he is waiting for you therefore unholding

PR #190, opened 2026-08-06
OSAC-3060: single-chart Kind deployment

omer-vishlitzky:
## Summary

Single deployment path for Kind environments using the same osac-installer charts as production. No kustomize, no setup.sh, no local IT charts. One way to deploy.

### What changed

**osac-prereqs chart** — conditional upstream installs for non-OLM (Kind) environments. Correction from an earlier draft of this description: cert-manager, trust-manager, and Envoy Gateway are **not** Chart.yaml dependencies on `osac-prereqs` — that chart has none. They're separate `helm install` calls orchestrated by the root Makefile's `install-prereqs-kind`, with `osac-prereqs` itself only rendering the CRs/resources that assume them already present:
- cert-manager (upstream jetstack, installed by the Makefile; `certManager.upstream.enabled` was leftover from an earlier design and has been removed — see "Fixes from review" below)
- trust-manager (upstream jetstack; `trustManager.upstream.enabled` disables osac-prereqs' own bundled manifests when trust-manager is installed separately)
- Envoy Gateway (`envoygateway.enabled`/`.resources`, installed by the Makefile)
- New templates: EnvoyProxy, GatewayClass, Gateway, TLSRoute, fake CRDs
- Keycloak Route gated behind OpenShift API capability detection

**Kind values files** — `osac-installer/values/kind/{prereqs,osac}.yaml`

**Root Makefile** — developer workflow:
```
make dev-env                      # Kind + prereqs + OSAC (~8 min)
make dev-env-full                 # + KubeVirt + AWX + catalog (~25 min, see note below)
make integration-test             # all component IT
make integration-test-fulfillment # fulfillment-service only
make integration-test-operator    # osac-operator only
make integration-test-bmf         # BMF operator only
make teardown                     # clean up
```

**Component test suites** — simplified to assume pre-existing cluster:
- osac-operator: removed cert-manager/prometheus install, kustomize deployment, fixed crash-loop detection + finalizer handling
- BMF: renamed test/e2e to test/integration, removed infra setup, fixed SA name + crash-loop detection. See "Fixes from review" — this now has real Metal3-backed provisioning coverage and a CI job, neither of which existed when this bullet was originally written.
- fulfillment-service: removed 871 lines of infra deployment from it_tool.go, replaced internal/testing dependency with direct exec

**Deleted** (all duplicate deployment paths):
- `kind-dev/setup.sh` and `teardown.sh`
- `fulfillment-service/it/charts/{postgres,keycloak,prometheus}`
- `fulfillment-service/internal/testing/{kind.go,command.go,examples/}`
- All kustomize overlays in osac-operator and BMF (`config/{default,manager,testing,console-proxy,...}`)
- `bare-metal-fulfillment-operator/hack/sync-helm-operator.py`
- All kustomize references in CI workflows and Makefiles

### Known limitations (not implemented, despite Makefile targets suggesting otherwise)

- **AWX operator install is a no-op.** `make install-awx` sets `--set awx.enabled=true`, but nothing consumes that flag anywhere in either chart — confirmed via `helm template`/grep. There is no working path (Helm dependency or otherwise) that installs the AWX operator on Kind. `dev-env-full`'s "+ AWX" is currently aspirational.
- **`make seed-catalog` is an unimplemented stub** (`echo "TODO: implement catalog seeding..."`). `dev-env-full`'s "+ catalog" does nothing yet.
- Neither of these is exercised by any CI job (`dev-env-full` isn't called anywhere in `.github/workflows/`), so this has no test-coverage impact today, but `make dev-env-full` doesn't currently deliver what its own help text promises.

### How it works

**Production (OpenShift)**:
```
Phase 1:  helm install osac-operators   → OLM Subscriptions
Phase 2:  helm install osac-prereqs     → trust-manager, CA, Keycloak
Phase 3:  helm install osac             → all OSAC components
```

**Kind (dev/IT)** — skip Phase 1, enable upstream deps in Phase 2:
```
Phase 2:  helm install osac-prereqs -f values/kind/prereqs.yaml
          → upstream cert-manager, trust-manager, Envoy Gateway, CA, Keycloak
Phase 3:  helm install osac -f values/kind/osac.yaml
          → same umbrella chart, bundledPostgres enabled
```

Same charts. Different values. Zero drift.

## Test plan

All of the following were run against real Kind clusters this round (not inferred):

- [x] `make dev-env` creates working Kind environment
- [x] `make integration-test-operator` passes (16/16 specs, fresh cluster)
- [x] `make integration-test-bmf` passes (6/6 specs, including new Metal3 provisioning tests; run twice in a row without tearing down to confirm idempotency)
- [x] `make integration-test-fulfillment` passes (258/258 specs, fresh cluster)
- [ ] `helm install osac-prereqs` with OpenShift values still works
- [ ] `helm install osac` with OpenShift values still works
- [x] Zero kustomize references remain (outside controller-gen labels)
- [x] `pre-commit run --all-files` clean (16 hooks, including gitleaks)
- [x] `go build`/`vet`/`golangci-lint` clean across all three Go modules; full unit suites green (fulfillment-service 85 suites, osac-operator + BMF envtest)


fulfillment-service/it/it_suite_test.go
```diff
@@ -33,35 +32,15 @@ import (
...
-	KeepService bool `json:"keep_service" envconfig:"keep_service" default:"false"`
-
-	// Debug indicates if the debug mode should be enabled. This means that the debugger binary will be added to
-	// the container image, and that the services will be started under the control of the debugger. Access to the
-	// debugger will be done via the following ports:
-	//
-	// - gRPC server: 30001
-	// - REST gateway: 30002
-	// - Controller: 30003
-	Debug bool `json:"debug" envconfig:"debug" default:"false"`
```
jhernand, line 51: How will the remote testing functionality work now? If it is lost then remove the configuration from the deployments.

fulfillment-service/AGENTS.md
```diff
@@ -70,18 +70,24 @@ ginkgo run -r
...
 
-# Preserve cluster for debugging
-IT_KEEP_KIND=true ginkgo run it
+# Run integration tests against the running cluster
+ginkgo run it
 
-# Run only setup (create cluster without tests)
-IT_KEEP_KIND=true ginkgo run --label-filter=setup it
+# Run only setup (connect + verify the pre-existing deployment, no tests)
+ginkgo run --label-filter=setup it
```
jhernand: This is useless if the setup of the cluster isn't done by this code. It would make more sense to remove this command completely and the "Setup" label.

fulfillment-service/it/it_suite_test.go
```diff
@@ -33,35 +32,15 @@ import (
 
 // Config contains configuration options for the integration tests.
 type Config struct {
-	// KeepKind indicates whether to preserve the kind cluster after tests complete.
-	// By default, the kind cluster is deleted after running the tests.
+	// KeepKind skips Cleanup()'s `kind export logs` dump -- see SetKeepCluster's doc comment.
+	// It does not affect cluster lifetime; Setup()/Cleanup() never create or destroy the
+	// cluster, only connect to and disconnect from a pre-existing one.
 	KeepKind bool `json:"keep_kind" envconfig:"keep_kind" default:"false"`
```
jhernand: This is useless now, I'd suggest you remove it.

fulfillment-service/AGENTS.md
```diff
@@ -70,18 +70,24 @@ ginkgo run -r
...
 
-# Clean up preserved cluster
-kind delete cluster --name fulfillment-service-it
+# Clean up the cluster when done
+kind delete cluster --name osac-dev  # or: make teardown (from repo root)
```
jhernand, line 87: Isn't there another place explaining cluster teardown? If so this should either be removed or replaced by a reference to that place.

fulfillment-service/AGENTS.md
```diff
@@ -265,7 +271,7 @@ As with any proto change, run `buf lint && buf generate` afterward (see [Linting
 - `SERVICE_SUFFIX` lint rule is intentionally excluded in `buf.yaml`
 - Unit tests: run `ginkgo run -r internal` (not `ginkgo run -r`) to avoid triggering integration tests
 - CI timeout: 1 hour for unit and integration test runs
-- Integration test logs uploaded as `logs-helm` and `logs-kustomize` artifacts (always, even on failure)
+- Integration test logs uploaded as the `logs` artifact (always, even on failure)
```
jhernand: This is no longer part of these tests, so it doesn't belong here.

fulfillment-service/README.md
```diff
@@ -206,42 +211,28 @@ To run the integration tests:
...
+whether cluster diagnostic logs are dumped on failure; it no longer affects cluster lifetime.
 
-By default, the kind cluster is deleted after the tests complete. If you want to preserve the cluster
-for debugging or manual inspection, you can set the `IT_KEEP_KIND` environment variable:
+The `setup` label gets you a deployed environment without running any actual test, useful for manual
+poking:
 
 ```bash
-$ IT_KEEP_KIND=true ginkgo run it
+$ ginkgo run --label-filter setup it
```
jhernand: This and the "setup" label should be removed.

fulfillment-service/README.md
```diff
@@ -206,42 +211,28 @@ To run the integration tests:
...
-That will create the Kind cluster, install the dependencies and deploy the application, but will not
-run any actual test.
-
-To clean up a preserved cluster manually:
+To tear down the cluster when you're done, from the `osac-workspace/osac` repo root:
 
 ```bash
-$ kind delete cluster --name fulfillment-service-it
+$ make teardown
 ```
```
jhernand, line 228: Doesn't belong here.

rccrdpccl (dismissed):
I think this change is too big and tries to tackle too many issues at once. For the next time, please split it into different logic units (i.e. kind-dev migration, it tests migration, bugfixes)

I think we should try to have everything defined as one or 2 charts, as the OLM charts dependency is not going to be easily solvable short term.

IMO we should focus on 2 aspects:
* clear API (value files structure)
* clear ownership (each service should define its dependencies)

WDYT about the following:
* osac-prereqs will output parsable output. this will allow us to parse it and decide wether we want to override some values for the next chart
* osac chart will run with a value file, but override some values based on the previous step output


```
  # Stage 1
  helm install osac-prereqs charts/osac-prereqs -f values/prod.yaml

  # Notes print something like:
  # OSAC_PREREQS_VALUES:
  # certManager:
  #   source: external
  # envoyGateway:
  #   source: external
  #   gatewayClass: openshift-default

  # Stage 2 — just pipe the notes
  helm install osac charts/osac -f values/prod.yaml \
    -f <(helm get notes osac-prereqs | yq '...')
```

This approach would achieve:
* Simple way to run it. No need to maintain makefiles, different environments and ways of deploying stuff. Just 2 charts well integrated amongst them
* Easily toggle between prod and dev (i.e. activating a OLM chart would automatically disable its counterpart)
* Clear ownership, as we have each component having their dependencies defined contextually, but with OLM (prod ready version) version available maintained in a different location (also indicating that the component itself is not maintaining the dependency deployment, just the minimum for dev/ci/compatibility)

PR #205, opened 2026-08-07
OSAC-3266: Add database migration to enforce name uniqueness and immutability

CrystalChun:
Name uniqueness for every resource is enforced through the tuple(tenant, project, name). This means a resource name must be unique within the tenant and project it occupies.

E.g. compute instance name is `dev-instance`
`dev-instance` can exist globally (no tenant, no project)
`dev-instance` can exist in tenant-A, no project
`dev-instance` can also be created in tenant-A, project-A
`dev-instance` can also be created in tenant-B
Another resource named `dev-instance` cannot be created in
tenant-A.

Names of resources cannot be changed after the resource is created.

## Summary by CodeRabbit

- **New Features**
  - Resource names are now unique within their applicable scope, including tenant, project, or global scopes.
  - Resource names cannot be changed after creation.
  - Duplicate names return a clear already-exists error, while attempted renames return an immutable-field error.
  - Soft-deleted resources can be recreated with the same name where supported.
- **Documentation**
  - Updated API documentation to describe name uniqueness and immutability rules.


fulfillment-service/internal/database/migrations/100_add_name_uniqueness_and_immutability.up.sql
```diff
@@ -0,0 +1,153 @@
...
+      join
+        pg_attribute a on a.attrelid = t.oid and a.attnum = any(ix.indkey)
+      where
+        t.relname = tbl and
+        ix.indisunique and
+        a.attname = 'name' and
+        not ix.indisprimary
+    loop
+      execute format('drop index if exists %I', idx_name);
+    end loop;
```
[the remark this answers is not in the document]
CrystalChun: Migration 40 changed `tenants` to `tenant` for users

fulfillment-service/internal/database/migrations/100_add_name_uniqueness_and_immutability_test.go
```diff
@@ -0,0 +1,392 @@
...
+		Expect(err).ToNot(HaveOccurred())
+
+		_, err = conn.Exec(ctx,
+			`insert into roles (id, name, tenant, data)
+			 values ($1, $2, $3, '{}')`,
+			"r2", "custom-role", "other-tenant",
+		)
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("unique"))
+	})
```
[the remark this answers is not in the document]
CrystalChun: already resolved

fulfillment-service/internal/servers/generic_server.go
```diff
@@ -705,6 +706,30 @@ func (s *GenericServer[O]) Update(ctx context.Context, request any, response any
 		tmpObject = requestObject
 	}
 
+	// Reject explicit attempts to change the name, which is immutable.
```
jhernand: Why do we need that? Won't the trigger catch this and result in a reasonable error message?

fulfillment-service/internal/servers/generic_server.go
```diff
@@ -705,6 +706,30 @@ func (s *GenericServer[O]) Update(ctx context.Context, request any, response any
...
+			if maskCoversField(requestMask, "metadata.name") || mergedName != "" {
+				return grpcstatus.Errorf(grpccodes.InvalidArgument, "%s", (&dao.ErrImmutable{
+					Fields: []string{"metadata.name"},
+				}).Error())
+			}
+		}
+	}
+
+	// Preserve the immutable name from the stored object so that requests omitting
+	// metadata don't accidentally clear it and trip the database immutability trigger.
```
jhernand: Is this something that happens today, before this change? If it isn't, why do we need to add this?

fulfillment-service/docs/API.md
```diff
@@ -75,7 +75,7 @@ The `Metadata` message is shared by all object types and contains the following
 | `creation_timestamp`   | `google.protobuf.Timestamp`  | Time the object was created.                     |
 | `deletion_timestamp`   | `google.protobuf.Timestamp`  | Time the object was marked for deletion.         |
 | `creator`              | `string`                     | Identity that created the object.                |
-| `name`                 | `string`                     | Human-friendly name (DNS label rules, optional). |
+| `name`                 | `string`                     | Human-friendly name (DNS label rules, optional). Unique within its scope and immutable after creation. |
```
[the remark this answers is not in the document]
CrystalChun: Doc will be added for this in a later change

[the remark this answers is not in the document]
CrystalChun: Doc will be added for this in a later change

fulfillment-service/internal/servers/generic_server.go
CrystalChun: If the name or metadata is omitted from the update request, it does trip the database immutability trigger because it thinks the name is "cleared"

It didn't happen before this change because name was not immutable, but this change adds the immutability aspect

fulfillment-service/internal/servers/generic_server.go
CrystalChun: Ah yes sorry I got caught up in some unit tests that were supposed to be triggering the name immutability but didn't. This is the wrong solution. 

Removed it, thank you for catching this!

fulfillment-service/docs/API.md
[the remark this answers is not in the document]
CrystalChun: No there's already a tracking issue

fulfillment-service/internal/servers/generic_server.go
jhernand: But isn't that the correct behaviour? If the user sends a request without an update mask that doesn't include the `metadata.name`  it means that they want to clear the metadata, and the server should reject that request.

CrystalChun: Gotcha! Thanks for the explanation. I've reverted it back to how it was.

DakCrowder (approved):

PR #207, opened 2026-08-08
OSAC-3735: force file-based secret storage in it harness

wgordon17:
## Summary
- `SecretStoreBuilder.Build()`'s keyring probe can't tell "keyring available" from "no keychain exists" on macOS, so the IT harness's per-test sandboxed `$HOME` (no keychain) always selects the keyring backend, which then fails or pops a real authorization dialog on save
- Adds an `OSAC_SECRET_STORE=file` override checked before the probe, wires it into the IT harness's `cliEnv()`, and documents it in the CLI's `--help` text
- Verified against the real `security` binary (dialog reproduced pre-fix) and the full IT suite's `cli`-labeled specs end-to-end (8/8 passing, no dialog, post-fix)

## Summary by CodeRabbit

* **New Features**
  * Added support for forcing file-based secret storage with `OSAC_SECRET_STORE=file`.
  * Updated configuration help text to explain file-based storage and keyring fallback behavior.
  * Secret storage remains available when the system keyring is unavailable or file storage is explicitly selected.

* **Bug Fixes**
  * Preserved normal keyring selection when the environment variable is unset or set to another value.

### DIFF
--- a/fulfillment-service/internal/config/config_secret_store.go
+++ b/fulfillment-service/internal/config/config_secret_store.go
@@ -81,7 +81,7 @@ func (b *SecretStoreBuilder) Build() (result SecretStore, err error) {
 	// $HOME still reports 'not found' rather than 'unavailable', so the probe alone would incorrectly select the
 	// keyring backend. Callers that know the keyring won't be usable, such as the integration test harness, use
 	// this override to force file-based storage without depending on the probe's outcome.
-	if os.Getenv(secretStoreEnvVar) == secretStoreFileValue {
+	if os.Getenv(secretStoreEnvVar) == secretStoreFile {
 		result, err = NewFileSecretStore().
 			SetLogger(b.logger).
 			SetDir(b.dir).
@@ -322,7 +322,7 @@ func (s *fileSecretStore) Save(ctx context.Context, object any) error {
 		return fmt.Errorf("failed to marshal secrets: %w", err)
 	}
 	dir := filepath.Dir(s.file)
-	err = os.MkdirAll(dir, os.FileMode(0755))
+	err = os.MkdirAll(dir, os.FileMode(0700))
 	if err != nil {
 		return fmt.Errorf("failed to create directory '%s': %w", dir, err)
 	}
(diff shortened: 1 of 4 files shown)


fulfillment-service/internal/config/config_secret_store.go
```diff
@@ -64,16 +64,31 @@ func (b *SecretStoreBuilder) SetDir(value string) *SecretStoreBuilder {
...
 func (b *SecretStoreBuilder) Build() (result SecretStore, err error) {
 	// Check parameters:
 	if b.logger == nil {
 		err = errors.New("logger is mandatory")
 		return
 	}
 
+	// Honor the explicit override, if set. This exists because the keyring probe below can't distinguish a
+	// keyring that is genuinely usable from one that doesn't exist at all: on macOS, probing a keychain-less
+	// $HOME still reports 'not found' rather than 'unavailable', so the probe alone would incorrectly select the
```
jhernand, line 81: Can't we fix this issue so that it works correctly in macOS?

fulfillment-service/internal/config/config_secret_store.go
```diff
@@ -307,7 +322,7 @@ func (s *fileSecretStore) Save(ctx context.Context, object any) error {
 		return fmt.Errorf("failed to marshal secrets: %w", err)
 	}
 	dir := filepath.Dir(s.file)
-	err = os.MkdirAll(dir, os.FileMode(0755))
+	err = os.MkdirAll(dir, os.FileMode(0700))
```
jhernand, line 325: This seems like an unrelated security improvement, should it be in a separate pull request?

wgordon17:
Cross-linking for comparison: opened #407 as an alternate approach to the same underlying bug. This PR (#207) bypasses the keyring probe via `OSAC_SECRET_STORE=file`; #407 instead provisions a real, throwaway per-test macOS keychain so the probe/save logic in `config_secret_store.go` is actually exercised on the platform where the bug lives. Submitted side-by-side for review comparison, not as a replacement.

jhernand:
I thought that the issue was that in macOS the CLI wasn't detecting correctly the fact that there isn't a keyring available. If it did, then it would fallback to use the file automatically. Can't we fix that?

PR #212, opened 2026-08-09
OSAC-2252: Populate default network_attachments when omitted on BareMetalInstance create

danmanor:
## [OSAC-2252](https://redhat.atlassian.net/browse/OSAC-2252): Populate default network_attachments when omitted on BareMetalInstance create

**Jira:** https://redhat.atlassian.net/browse/OSAC-2252
**Story type:** Task
**EP:** [BMaaS Networking](/enhancements/OSAC-1437-bmaas-networking) §Workflow

### Summary

When `network_attachments` is omitted on BareMetalInstance creation, the fulfillment-service now auto-populates with the tenant's default IPv4 Subnet and default SecurityGroup (labeled `osac.openshift.io/default`), using the first `fabric`-role interface from the HostType. If no defaults exist (tenant networking not yet provisioned), the field remains empty — existing behavior is preserved.

### Changes

- **`fulfillment-service/internal/servers/private_baremetal_instances_server.go`:**
  - Added `securityGroupsDao` and `tenancyLogic` to server struct for default resource lookup and tenant resolution
  - Added `applyDefaultNetworkAttachments()` in the `Create()` flow between `validateSpec` and `validateNetworkAttachments`
  - Added `findDefaultSubnet()` — lists subnets by `osac.openshift.io/default` label, filtered to tenant, returns first non-deleted IPv4 subnet
  - Added `findDefaultSecurityGroup()` — same pattern for security groups
  - Added `resolveDefaultInterface()` — walks catalog_item → template → host_type chain, returns first `fabric`-role interface name; returns error if HostType exists but has no fabric interface

### Testing

- **Unit tests:** 6 new tests in `private_baremetal_instances_server_test.go` covering:
  - Happy path: defaults populated with correct subnet, SG, and interface
  - Explicit attachments: no-op (unchanged)
  - No HostType on template: interface omitted
  - No default subnet: graceful skip
  - No default security group: graceful skip
  - No fabric interface on HostType: FailedPrecondition error
- **Regression:** All 1545 existing tests pass, 0 failures
- **Coverage:** All behavioral paths through public `Create` API are tested

### Acceptance Criteria

- [x] Auto-populate `network_attachments` when omitted on Create
- [x] Default Subnet from tenant's default IPv4 Subnet (labeled `osac.openshift.io/default`)
- [x] Default SecurityGroup from tenant's default SG (same label)
- [x] Default interface from first `fabric`-role HostType interface
- [x] Resolved attachments stored in spec (self-describing after creation)
- [x] Explicit `network_attachments` unchanged (no default population)

## Summary by CodeRabbit

* **New Features**
  * Bare-metal instance creation now automatically selects default subnet, security group, and fabric network attachments when applicable.
  * Explicit network attachments remain unchanged.
  * Interfaces without a host type are handled without unnecessary defaults.

* **Bug Fixes**
  * Improved fallback behavior when defaults are unavailable.
  * Added validation and clearer failure handling for host types missing required fabric interfaces.

### DIFF
--- a/fulfillment-service/internal/servers/private_baremetal_instances_server.go
+++ b/fulfillment-service/internal/servers/private_baremetal_instances_server.go
@@ -16,6 +16,7 @@ package servers
 import (
 	"context"
 	"errors"
+	"fmt"
 	"log/slog"
 	"maps"
 	"strings"
@@ -49,6 +50,7 @@ var _ privatev1.BareMetalInstancesServer = (*PrivateBareMetalInstancesServer)(ni
 type PrivateBareMetalInstancesServer struct {
 	privatev1.UnimplementedBareMetalInstancesServer
 	logger             *slog.Logger
+	tenancyLogic       auth.TenancyLogic
 	generic            *GenericServer[*privatev1.BareMetalInstance]
 	catalogItemsDao    *dao.GenericDAO[*privatev1.BareMetalInstanceCatalogItem]
 	templatesDao       *dao.GenericDAO[*privatev1.BareMetalInstanceTemplate]
@@ -56,6 +58,7 @@ type PrivateBareMetalInstancesServer struct {
 	subnetsDao         *dao.GenericDAO[*privatev1.Subnet]
 	virtualNetworksDao *dao.GenericDAO[*privatev1.VirtualNetwork]
 	networkClassesDao  *dao.GenericDAO[*privatev1.NetworkClass]
+	securityGroupsDao  *dao.GenericDAO[*privatev1.SecurityGroup]
 }
 
 func NewPrivateBareMetalInstancesServer() *PrivateBareMetalInstancesServerBuilder {
@@ -155,6 +158,15 @@ func (b *PrivateBareMetalInstancesServerBuilder) Build() (result *PrivateBareMet
 		return
 	}
 
+	securityGroupsDao, err := dao.NewGenericDAO[*privatev1.SecurityGroup]().
+		SetLogger(b.logger).
+		SetTenancyLogic(b.tenancyLogic).
+		SetMetricsRegisterer(b.metricsRegisterer).
+		Build()
+	if err != nil {
+		return
+	}
+
 	generic, err := NewGenericServer[*privatev1.BareMetalInstance]().
 		SetLogger(b.logger).
 		SetService(privatev1.BareMetalInstances_ServiceDesc.ServiceName).
@@ -169,6 +181,7 @@ func (b *PrivateBareMetalInstancesServerBuilder) Build() (result *PrivateBareMet
 
 	result = &PrivateBareMetalInstancesServer{
 		logger:             b.logger,
+		tenancyLogic:       b.tenancyLogic,
 		generic:            generic,
 		catalogItemsDao:    catalogItemsDao,
 		templatesDao:       templatesDao,
@@ -176,6 +189,7 @@ func (b *PrivateBareMetalInstancesServerBuilder) Build() (result *PrivateBareMet
 		subnetsDao:         subnetsDao,
 		virtualNetworksDao: virtualNetworksDao,
 		networkClassesDao:  networkClassesDao,
+		securityGroupsDao:  securityGroupsDao,
 	}
 	return
 }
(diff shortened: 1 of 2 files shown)


fulfillment-service/internal/servers/private_baremetal_instances_server.go
```diff
@@ -268,6 +285,147 @@ func (s *PrivateBareMetalInstancesServer) validateSpec(bmi *privatev1.BareMetalI
...
+	if err != nil {
+		return "", nil
+	}
+	templateID := refKey(catResp.GetObject().GetTemplate())
+	if templateID == "" {
+		return "", nil
+	}
+	tmplResp, err := s.templatesDao.Get().SetId(templateID).Do(ctx)
+	if err != nil {
+		return "", nil
```
ybettan: The catalog → template → host_type lookup chain silently swallows all DAO errors (`return "", nil`). If a call fails for a transient reason (e.g. DB timeout), the instance gets created with no interface — silently degraded. This is inconsistent with `findDefaultSubnet`/`findDefaultSecurityGroup` which return `Internal` errors on DAO failures. Should these propagate the error instead?

fulfillment-service/internal/servers/private_baremetal_instances_server.go
```diff
@@ -268,6 +285,147 @@ func (s *PrivateBareMetalInstancesServer) validateSpec(bmi *privatev1.BareMetalI
...
+		if err != nil {
+			s.logger.ErrorContext(ctx, "Failed to determine default tenant for network attachment defaults",
+				slog.Any("error", err))
+			return grpcstatus.Errorf(grpccodes.Internal, "failed to determine tenant")
+		}
+	}
+
+	subnet, err := s.findDefaultSubnet(ctx, tenantName)
+	if err != nil {
+		return err
```
ybettan, line 310: If the tenant has a default SecurityGroup but no default Subnet (or vice versa), this silently skips populating any defaults. Is that intentional? A partial setup (one default exists but the other doesn't) might indicate a misconfiguration worth surfacing to the user rather than silently leaving `network_attachments` empty.

fulfillment-service/internal/servers/private_baremetal_instances_server.go
```diff
@@ -268,6 +285,147 @@ func (s *PrivateBareMetalInstancesServer) validateSpec(bmi *privatev1.BareMetalI
...
+	ctx context.Context, tenantName string) (*privatev1.Subnet, error) {
+	filter := fmt.Sprintf(
+		"this.metadata.labels['%s'] == 'true' && this.metadata.tenant == %q",
+		defaultLabel, tenantName,
+	)
+	listResp, err := s.subnetsDao.List().SetFilter(filter).Do(ctx)
+	if err != nil {
+		s.logger.ErrorContext(ctx, "Failed to list default subnets",
+			slog.String("tenant", tenantName), slog.Any("error", err))
+		return nil, grpcstatus.Errorf(grpccodes.Internal, "failed to find default subnet")
```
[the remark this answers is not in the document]
danmanor: False positive — tenant names in OSAC are system identifiers (e.g., `system`), not customer PII. The existing `default_networking_provisioner.go` in the same package logs `slog.String("tenant", tenantName)` in 4 places. This follows the established codebase pattern.

fulfillment-service/internal/servers/private_baremetal_instances_server.go
danmanor: Good catch — fixed in a7367280. `resolveDefaultInterface` now distinguishes `NotFound` errors (chain unresolvable → return empty, no error) from other DAO errors (e.g., DB timeout → propagate as `Internal`), consistent with `findDefaultSubnet`/`findDefaultSecurityGroup`.

fulfillment-service/internal/servers/private_baremetal_instances_server.go
danmanor: Intentional. Default networking is provisioned as a unit during tenant onboarding (VN + Subnet + SG + NATGateway together, gated by `DefaultNetworkingReady` condition on Tenant). A partial state means onboarding is still in progress. Silently skipping is the correct behavior — the tenant can retry once defaults are ready. An error here would block BMI creation during the (brief) tenant onboarding window.

PR #227, opened 2026-08-10
OSAC-3848: fix started.v1/resumed.v1 mapping on first activation

omer-vishlitzky:
Watch Consumer mapped every non-empty previous state crossing into billable to `resumed.v1`, including a resource's genuine first activation — CREATE always seeds a concrete state before the billable transition fires, so the `started.v1` table row was unreachable. Resolves started vs resumed from a persisted `EverBillable` flag instead.

Jira: https://redhat.atlassian.net/browse/OSAC-3848

## Summary by CodeRabbit

* **New Features**
  * Added persistent tracking to distinguish a resource’s first billable activation from later resumptions.
  * Updated metering events for compute instances and clusters to report accurate `started` or `resumed` states.
  * Improved resource-state persistence and reconciliation for billability history.

* **Bug Fixes**
  * Corrected event reporting when resources return to billable states.
  * Fixed first-boot scenarios that could be incorrectly reported as resumptions.
  * Ensured billability indicators remain consistent with resource history.


osac-metering/metering-service/internal/database/migrations/0_create_metering_resource_state.up.sql
```diff
@@ -13,6 +13,7 @@ CREATE TABLE metering_resource_state (
     current_state     TEXT        NOT NULL,
     previous_state    TEXT,
     is_billable       BOOLEAN     NOT NULL DEFAULT FALSE,
+    ever_billable     BOOLEAN     NOT NULL DEFAULT FALSE,
```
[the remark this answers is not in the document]
masayag: we don't have this released anywhere yet - it is ok to skip the upgrade flow

masayag:
Following up on the discussion about reconciling `ever_billable` and `is_billable` — here's a concrete suggestion for relying on `ever_billable` + `billable_since` together, rather than three independently-writable booleans/timestamps.

**The invariant already holds**: every write path in this PR sets `is_billable == true` iff `billable_since != nil`. Instead of trusting each of the four write sites (Watch Consumer, both Reconciler branches, and the base case) to keep `is_billable` in sync by convention, Postgres can enforce it structurally:

```sql
is_billable BOOLEAN GENERATED ALWAYS AS (billable_since IS NOT NULL) STORED
```

This removes `is_billable` from the `INSERT`/`UPDATE` column list entirely — it can never drift from `billable_since` again, by construction, regardless of what any current or future caller does.

**`ever_billable` can go one step further too.** It can't be a `GENERATED` column (its value depends on the row's *previous* state, which a same-row expression can't see), but the `UPSERT` can derive it from `billable_since` directly instead of trusting the caller's already-OR'd value:

```sql
ON CONFLICT (resource_id) DO UPDATE SET
    ...
    ever_billable = metering_resource_state.ever_billable OR (EXCLUDED.billable_since IS NOT NULL),
    ...
```

vs. the current:

```sql
ever_billable = metering_resource_state.ever_billable OR EXCLUDED.ever_billable,
```

The difference matters: the current formula still trusts whatever `ever_billable` the caller computed and just prevents it from *decreasing*. If a caller computes `ever_billable=false` when it should be `true` (exactly the class of bug fixed in `60dc25a2`'s reconciler `state_drift` path), `false OR false` doesn't self-correct. Deriving straight from `billable_since` would have — that field was already being set correctly at every call site (including the buggy reconciler path, before this PR's fix), so this formula is a strict robustness improvement, not just a style preference. It also collapses "how do we compute `ever_billable`" from 3+ call sites down to a single formula.

Happy to open this as a fast-follow PR if useful — didn't want to push a commit onto this branch without checking first.

osac-metering/metering-service/internal/database/migrations/0_create_metering_resource_state.up.sql
amito: This edit is on an existing migration (`0_create_...`). I assume that we can make this manipulation on an existing migration as we're pre-production? Is there a case for adding another migration for DB "version management"?

osac-metering/metering-service/internal/database/container.go
```diff
@@ -194,13 +194,6 @@ func (i *Instance) init(ctx context.Context) error {
 	return nil
 }
 
-func (i *Instance) URL(ctx context.Context) (string, error) {
```
amito, line 197: Is this related to the changes introduced in this PR?

osac-metering/metering-service/internal/database/migrations/0_create_metering_resource_state.up.sql
omer-vishlitzky: I would say no, we don't want to add more migration files if there is nothing to migrate.

osac-metering/metering-service/internal/database/container.go
omer-vishlitzky: no, just dead code

PR #233, opened 2026-08-10
OSAC-3864: extract shared types into common proto files

danielerez:
Part of [OSAC-3662](https://redhat.atlassian.net/browse/OSAC-3662).

## Summary
  - Extracts shared protobuf types that cause circular import chains into dedicated common proto
  files, unblocking [OSAC-3865](https://redhat.atlassian.net/browse/OSAC-3865) (move reference types to their target files)
  - Applied to both private and public proto trees (8 new files, 16 files updated)

  ### Types extracted
  
  | Type(s) | From | To |
  |---------|------|----|
  | `ComputeInstanceImage`, `ComputeInstanceDisk` | `compute_instance_type.proto` |
  `compute_instance_common_type.proto` |
  | `BareMetalInstanceImage` | `baremetal_instance_type.proto` |
  `baremetal_instance_common_type.proto` |
  | `ClusterNetwork` | `cluster_type.proto` | `cluster_common_type.proto` |
  | `SecurityRule`, `Protocol` | `security_group_type.proto` | `security_rule_type.proto` |

  ### Why
  
  Template and catalog-item files import the base resource file for these shared types. Moving a
  reference back to the template/catalog-item file would create a circular import (see [PR #85
  review](https://github.com/osac-project/osac/pull/85) for the full analysis). Extracting the shared
  types into leaf files breaks every cycle.

## Test plan
  
  - [x] `buf lint` passes with zero errors
  - [ ] `buf generate` produces valid Go code
  - [ ] `ginkgo run -r internal` passes (no runtime behavior change — proto messages are structurally
  identical)


CrystalChun (approved):
/lgtm

danielerez:
Commits already merged as part of: https://github.com/osac-project/osac/pull/234

PR #237, opened 2026-08-10
OSAC-3764: Add BMH lifecycle manager

mennyaboush:
## Summary

On-demand BareMetalHost CR management for inventory backends paired with Metal3 ([OSAC-1339](https://redhat.atlassian.net/browse/OSAC-1339)):

- **CreateBMH**: Creates BMH CR with BMC address, credentials, boot MAC, consumer reference. Idempotent — matches full ConsumerRef identity. BMC cert verification configurable.
- **DeleteBMH**: Deletes BMH CR by name. Idempotent (ignores NotFound). Does not delete BMC credentials Secrets.
- **IsBMHReady**: Returns true when provisioning state is Available and operational status is OK.

Reusable by any inventory backend — not BCM-specific. The BCM backend is the first consumer (Phase 2: OSAC-3767, [OSAC-3771](https://redhat.atlassian.net/browse/OSAC-3771)).

**Depends on:** [osac#228](https://github.com/osac-project/osac/pull/228) (BCM HTTP client)

## Test plan

- [ ] `make build` passes
- [ ] `make test` passes (7 new tests: create new/idempotent/conflict, delete existing/not-found, readiness available/registering/error/not-found)
- [ ] `make lint` passes (0 issues)

## Summary by CodeRabbit

- **New Features**
  - Added BareMetalHost lifecycle management, including creation, deletion, conflict handling, credential preservation, and readiness checks.
  - Added integration with the BCM inventory service, including secure connectivity, version validation, device retrieval and updates, inventory metadata management, metrics, and clearer error reporting.
  - Added configurable host settings such as BMC access, boot configuration, labels, consumer references, and certificates.

- **Tests**
  - Added comprehensive coverage for BareMetalHost lifecycle operations and BCM connectivity, validation, API behavior, and error handling.

### DIFF
--- a/bare-metal-fulfillment-operator/internal/baremetalhost/baremetalhost_suite_test.go
+++ b/bare-metal-fulfillment-operator/internal/baremetalhost/baremetalhost_suite_test.go
@@ -1,0 +1,29 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package baremetalhost
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2" //nolint:revive,staticcheck
+	. "github.com/onsi/gomega"    //nolint:revive,staticcheck
+)
+
+func TestBareMetalHost(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "BareMetalHost Suite")
+}
--- a/bare-metal-fulfillment-operator/internal/baremetalhost/manager.go
+++ b/bare-metal-fulfillment-operator/internal/baremetalhost/manager.go
@@ -1,0 +1,177 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package baremetalhost provides utilities for managing on-demand
+// BareMetalHost CRs. It is a shared utility — any inventory backend
+// that needs to create, delete, or check readiness of BMH CRs can
+// use this package.
+package baremetalhost
+
+import (
+	"context"
+	"fmt"
+
+	metal3api "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// CreateParams holds the parameters for creating a BareMetalHost CR.
+type CreateParams struct {
+	Name                           string
+	BMCAddress                     string
+	CredentialsSecret              string
+	BootMACAddress                 string
+	DisableCertificateVerification bool
+	ConsumerRef                    *corev1.ObjectReference
+	Labels                         map[string]string
+}
+
+// Manager handles on-demand BareMetalHost CR creation, deletion, and
+// readiness checking. Any inventory backend that needs BMH CRs can
+// use this utility.
+type Manager struct {
+	client    client.Client
+	namespace string
+}
+
+// NewManager creates a manager for the given Metal3 namespace.
+func NewManager(k8sClient client.Client, namespace string) *Manager {
+	return &Manager{
+		client:    k8sClient,
+		namespace: namespace,
+	}
+}
+
+// Namespace returns the Metal3 namespace used for BareMetalHost CRs.
+func (m *Manager) Namespace() string {
+	return m.namespace
+}
+
+// CreateBMH creates a BareMetalHost CR in the configured namespace.
+// Sets spec.online = false — the controller manages power via reconcileManagement.
+// Idempotent: if a BMH with the same name already exists and its consumerRef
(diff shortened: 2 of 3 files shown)


bare-metal-fulfillment-operator/internal/inventory/bmh_lifecycle.go
adriengentil, line 1: I know it was part of the design to create a new struct, but I think it's no different from the metal3 client. So I wonder if these functions should just be part of the existing metal3 client. In the current state, the BCM client will be tightly coupled with metal3 anyway

/cc @carbonin

carbonin: The metal3 client implements a specific interface to fulfill the generic inventory backend. I'm not sure this logic actually belongs there. Sure, they're both metal3 related, but this is going to be a totally separate runtime right? I think it actually might be more confusing to add it to the inventory client.

adriengentil: My concern is that will have metal3/bmh operations scattered through the codebase. Would it make sense to centralize the metal3-related code in its own package, and keep the concrete implementation of the inventory interface in this "inventory" package?

bare-metal-fulfillment-operator/internal/inventory/bmh_lifecycle.go
```diff
@@ -0,0 +1,162 @@
...
+	namespace string
+}
+
+// NewBMHLifecycleManager creates a manager for the given Metal3 namespace.
+func NewBMHLifecycleManager(k8sClient client.Client, namespace string) *BMHLifecycleManager {
+	return &BMHLifecycleManager{
+		client:    k8sClient,
+		namespace: namespace,
+	}
+}
```
ybettan: I don't see any interface that that struct is implementing.

This is a golang best practice, especially for mocking it for upper layer's testing - see dependency injection in golang https://medium.com/@raafvargas/dependency-injection-in-go-35293ef7b6.

This will make it much harder for consumer of this code to write test for their changes withough caring about the dependency packages they use.

bare-metal-fulfillment-operator/internal/inventory/bmh_lifecycle.go
ybettan: > I know it was part of the design to create a new struct, but I think it's no different from the metal3 client. So I wonder if these functions should just be part of the existing metal3 client. In the current state, the BCM client will be tightly coupled with metal3 anyway

I am not familiar with the code as you probably are but I tend to agree with @adriengentil.

Since `bare-metal-fulfillment-operator/internal/inventory/client.go` is defining the inventory behavior then I think it belong their as well. We can have a second interface for `CreateHost`, `DeleteHost`, `UpdateHost` that will only be implemented by the BCM backend and leave the BMH creation for the admin for metal3/openstack implementations.

mennyaboush: > > I know it was part of the design to create a new struct, but I think it's no different from the metal3 client. So I wonder if these functions should just be part of the existing metal3 client. In the current state, the BCM client will be tightly coupled with metal3 anyway
> 
> I am not familiar with the code as you probably are but I tend to agree with @adriengentil.
> 
> Since `bare-metal-fulfillment-operator/internal/inventory/client.go` is defining the inventory behavior then I think it belong their as well. We can have a second interface for `CreateHost`, `DeleteHost`, `UpdateHost` that will only be implemented by the BCM backend and leave the BMH creation for the admin for metal3/openstack implementations.

@ybettan  I thinking interface is the right choice  - as you mention also in your [second comment](https://github.com/osac-project/osac/pull/237#discussion_r3764825563)
but I didn't fully understand your approach... You think that the interface should be implement by the [Metal3 Inventory client](https://github.com/osac-project/osac/blob/main/bare-metal-fulfillment-operator/internal/inventory/metal3.go#L55) as Adrien suggest?  just declare there the new interface and implement it at the BCM client only? or there is different approach I missed here?

carbonin: >My concern is that will have metal3/bmh operations scattered through the codebase

I don't really see this as a problem. Actually the way this PR is written is that this should be a utility that is usable by anything that needs it. The fact that it shares technology with the metal3 inventory backend is kinda irrelevant, they're not the same thing. I'd rather group code by function than what libraries it happens to integrate with.

>Since bare-metal-fulfillment-operator/internal/inventory/client.go is defining the inventory behavior then I think it belong their as well.

I disagree, this logic (creating BMHs from an external source) is specific to BCM for now. The client interface in `bare-metal-fulfillment-operator/internal/inventory/client.go` dictates the behavior that _every_ backend must implement.

This logic for creating BMHs is something that has nothing to do directly with assigning hosts from the outside inventory from what I can see in this PR so I don't think it actually belongs in `inventory` at all. Maybe `internal/baremetalhost`?

Then if something else needs to create BMHs they can use that package as well. No need to expand the `inventory` package's scope.

> We can have a second interface for CreateHost, DeleteHost, UpdateHost that will only be implemented by the BCM backend and leave the BMH creation for the admin for metal3/openstack implementations.

I don't think we need this because the host CRUD is not happening in the backend system. The host will always be created the same way (create a BMH CR). We could create an interface for test mocks, but I don't see the need for one in this code.

TLDR; If we want this BMH CRUD to be a shared utility then put it some place external to this package entirely. The BCM inventory backend can use it when that backend comes along and implements the existing inventory client interface.

mennyaboush: @carbonin In general I agree with you but 

> We could create an interface for test mocks, but I don't see the need for one in this code.

enabling develop tests easily for any future inventory that will use the bmh management is not reason good enough for the interface?

adriengentil: @mennyaboush we don't really need an interface as we have only one way to create bmh. So, I think the interface should be created only for the scope of the mocks

ybettan: > I disagree, this logic (creating BMHs from an external source) is specific to BCM for now.

Agree

> The client interface in bare-metal-fulfillment-operator/internal/inventory/client.go dictates the behavior that every backend must implement.

In only dictate the behavior if the interface is used in the controller (then it has to be implemented by every backend). This is why I have suggested an additional interface (that will not be used directly in the controller to make sure we don't require metal3/openstack to implement it as well). 

But...

> Maybe internal/baremetalhost

I agree, this is a better suggestion. What is the direction we are going toward, admin create BMHs in advance or OSAC creates BMH for the admin on demand (based on some kind of limit/template defined by the admin)?

> Then if something else needs to create BMHs they can use that package as well. No need to expand the inventory package's scope.

If we are going toward a solution that creates BMH for the admin on demand (for all backend eventually) then I don't see it to be very different logically from the `inventory.Clien` interface TBH.

> We could create an interface for test mocks, but I don't see the need for one in this code.

I think that every package "contact point" should be an interface as a general rule of thumb. It create a clean interface between packages and allow writing real *unit* tests isolated from other.

I won't die on that hill but those are my 2 cents.

adriengentil (approved):

PR #239, opened 2026-08-10
OSAC-3610: parameterize hardcoded Keycloak credentials in osac-installer

amej:
## Summary

- `osac-prereqs` shipped Keycloak with hardcoded admin (`admin`/`admin`) and demo tenant user (`foobar`) credentials with no way to override them — a blocker for standing up Red Hat Demo Platform environments without publicly-known credentials.
- Adds `keycloak.adminUsername`, `keycloak.adminPassword`, and `keycloak.defaultUserPassword` Helm values, defaulting to today's literals. Per agreed direction with Wolfgang, [OSAC-3610](https://redhat.atlassian.net/browse/OSAC-3610) ticket reporter: values are overridable but **not required** — `osac-test-infra`'s e2e suite and existing CI values files are unaffected, and Demo Platform's own deploy automation is responsible for overriding with a generated value.
- Reuses the existing `resolve-realm-secrets.sh` placeholder-substitution hook (previously used only for OIDC client secrets) to set the realm-admin's password at Keycloak import time, replacing a static baked-in argon2 credential hash.
- **Sources credentials via a chart-managed `keycloak-admin-credentials` Secret (`secretKeyRef`), not plaintext in pod specs.** Added in response to a CodeRabbit security finding — the chart materializes `values.yaml`'s existing `adminUsername`/`adminPassword`/`defaultUserPassword` into a Secret and every consumer (`KEYCLOAK_ADMIN`, `KEYCLOAK_ADMIN_PASSWORD`, `REALM_ADMIN_USERNAME`, `REALM_ADMIN_PASSWORD`, `ADMIN_USERNAME`, `ADMIN_PASSWORD`, `DEFAULT_USER_PASSWORD`) reads from it via `secretKeyRef`. This is backward-compatible: the Helm values, defaults, and override mechanism are unchanged — `osac-test-infra`/CI values files are unaffected — only the in-pod delivery mechanism changed (no longer visible via `oc get deployment -o yaml`, only via `oc get secret`). Matches the pattern already used by the static `prerequisites/keycloak/service/` manifests below.
- Applies the equivalent change to the static `prerequisites/keycloak/service/` reference manifests via the same `keycloak-admin-credentials` Secret (operator-created prerequisite — see header comment in `deployment.yaml`).
- Fixes escaping bugs surfaced across two rounds of testing/review: the realm.json placeholder substitution didn't JSON-escape passwords containing backslashes or control characters (tabs/newlines), and the demo-user `reset-password` API call interpolated the password into a JSON body with no escaping at all (both previously safe only because the values were fixed literals) — replaced hand-rolled sed escaping with `python3 json.dumps` everywhere. Also fixed a `curl` call in the static manifest that silently swallowed HTTP errors (missing `-f`).
- Adds `scripts/validate-keycloak-credentials.sh` as a regression test (8 checks, covering both the chart and static-manifest paths), wired into `make helm-validate`.

## ✅ Live-Cluster Verification — Complete (re-verified after the Secret-based rework)

Originally flagged as a required pre-merge blocker (the realm-admin's plaintext `"value"` credential relies on Keycloak's `--import-realm` bootstrap accepting an unhashed password at import time — a new code path for this codebase, since every other user's password is set post-import via the Admin REST API instead). **Verified against a live kind cluster, then re-verified after adding the `keycloak-admin-credentials` Secret / `secretKeyRef` rework:**

- `resolve-realm-secrets.sh`'s init container ran cleanly; the imported `realm.json` showed the correctly-resolved plaintext admin credential.
- Keycloak logs: `Realm 'osac' imported` / `Import finished successfully` — no errors on the plaintext credential.
- **Positive control:** `admin`/`admin` login against the `osac` realm returned a valid access token — confirmed both before and after the Secret-based rework.
- **Negative control:** a wrong password correctly returned `invalid_grant`.
- The `keycloak-set-passwords` Job completed successfully for all 7 demo users (now also `secretKeyRef`-sourced); verified `tenant1_admin`/`foobar` also authenticates.
- Confirmed via `kubectl get deployment -o jsonpath` that the pod spec now shows `secretKeyRef` (not a plaintext value) for all credential env vars, and via `kubectl get secret -o jsonpath | base64 -d` that the Secret itself holds the expected values.

No fallback to the originally-considered Job-based Admin REST API approach is needed — the sed-substitution reuse works as designed. 

## Jira

https://redhat.atlassian.net/browse/OSAC-3610

## Test plan

- [x] `helm lint charts/osac-operators/` and `helm lint charts/osac-prereqs/` pass
- [x] `helm template` against all 5 real values files (`bmaas-ci`, `caas-ci`, `development`, `full-ci`, `vmaas-ci`) for `osac-operators`, `osac-prereqs`, and `osac` (umbrella, with the documented `--set service.externalHostname=...`/`internalHostname=...` workaround for the pre-existing schema requirement)
- [x] New regression suite (`scripts/validate-keycloak-credentials.sh`, wired into `make helm-validate`, 8 checks): defaults preserved (via the Secret), `--set` overrides propagate into the Secret correctly, pod specs never carry plaintext credentials, `realm.json`/resolve-realm-secrets coverage for both the chart and static-manifest paths, all JSON-escaping fixes verified against the actual production scripts (not reimplementations) — including deliberately reverting each fix and confirming the test catches the regression, then restoring it
- [x] `pre-commit run --all-files` passes (three pre-existing, unrelated failures: missing `buf`/`golangci-lint` binaries in this environment — no Go/proto files touched by this change)
- [x] Live-cluster verification — **complete, re-verified after the Secret rework, see above**

## Confidence

High — the core parameterization is thoroughly verified structurally and via live-cluster testing (twice, before and after the Secret-based rework), all escaping bugs found during testing/review now have genuine regression coverage, and the previously-unverified realm-import behavior is confirmed working end-to-end.

## Rollback

Revert this commit. Since all new Helm values default to today's hardcoded literals, no data migration or state cleanup is needed — a rollback returns exactly to the pre-fix hardcoded-credential behavior.

## Risk Assessment

Low — changes are contained to the `osac-prereqs` chart, its files, and the static `prerequisites/keycloak/service/` reference manifests. No controller code, CRDs, or other charts touched. The previously-open live-cluster risk is now closed, re-verified after the Secret-based credential delivery was added.

---

_This PR description was drafted with AI assistance ([create-pr](https://github.com/osac-project/osac-workspace/tree/main/skills/create-pr) v0.1.0). Review for accuracy_

## New Features
- Keycloak administrator and default user credentials can be configured through Helm values and deployment secrets.
- Realm initialization and demo-user password setup now use the configured credentials.
- Credentials containing special characters are handled safely.

## Bug Fixes
- Replaced fixed credential handling with Secret-based configuration for more reliable deployments.

## Tests
- Added automated validation for default, customized, Secret-based, and special-character credentials.
- Helm validation now checks credential configuration during chart validation.

### DIFF
--- a/osac-installer/charts/osac-prereqs/files/hooks/resolve-realm-secrets.sh
+++ b/osac-installer/charts/osac-prereqs/files/hooks/resolve-realm-secrets.sh
@@ -26,12 +26,14 @@ REALM_ADMIN_PASSWORD="${REALM_ADMIN_PASSWORD:?REALM_ADMIN_PASSWORD must be set}"
 
 # REALM_ADMIN_USERNAME/PASSWORD are user-supplied Helm values (unlike the
 # auto-generated base64 client secrets above) substituted into JSON string
-# values, so each needs two escaping passes: first JSON string-escaping
-# (backslash, then quote) since the raw value lands inside a JSON string,
-# then sed replacement-escaping (backslash, ampersand, and the s### delimiter)
-# so sed doesn't reinterpret characters introduced by the JSON escaping.
+# values, so each needs two escaping passes: first full JSON string escaping
+# (via python3's json.dumps, so control characters like tabs/newlines are
+# handled correctly, not just backslash/quote) since the raw value lands
+# inside a JSON string, then sed replacement-escaping (backslash, ampersand,
+# and the s### delimiter) so sed doesn't reinterpret characters introduced
+# by the JSON escaping.
 json_escape_string() {
-    printf '%s' "$1" | sed -e 's/\\/\\\\/g' -e 's/"/\\"/g'
+    python3 -c "import json,sys; print(json.dumps(sys.argv[1])[1:-1])" "$1"
 }
 
 escape_sed_replacement() {
--- a/osac-installer/charts/osac-prereqs/files/hooks/resolve-realm-secrets.sh
+++ b/osac-installer/charts/osac-prereqs/files/hooks/resolve-realm-secrets.sh
@@ -21,9 +21,28 @@ ADMIN_SECRET=$(oc get secret "${SECRET_NAME}" -n "${NAMESPACE}" -o jsonpath='{.d
 [[ -n "${CONTROLLER_SECRET}" ]] || { echo "ERROR: ${SECRET_NAME} missing osac-controller key" >&2; exit 1; }
 [[ -n "${ADMIN_SECRET}" ]] || { echo "ERROR: ${SECRET_NAME} missing osac-admin key" >&2; exit 1; }
 
+REALM_ADMIN_USERNAME="${REALM_ADMIN_USERNAME:-admin}"
+REALM_ADMIN_PASSWORD="${REALM_ADMIN_PASSWORD:?REALM_ADMIN_PASSWORD must be set}"
+
+# REALM_ADMIN_USERNAME/PASSWORD are user-supplied Helm values (unlike the
+# auto-generated base64 client secrets above) substituted into JSON string
+# values, so each needs two escaping passes: first JSON string-escaping
+# (backslash, then quote) since the raw value lands inside a JSON string,
+# then sed replacement-escaping (backslash, ampersand, and the s### delimiter)
+# so sed doesn't reinterpret characters introduced by the JSON escaping.
+json_escape_string() {
+    printf '%s' "$1" | sed -e 's/\\/\\\\/g' -e 's/"/\\"/g'
+}
+
+escape_sed_replacement() {
+    printf '%s' "$1" | sed -e 's/[\&#]/\\&/g'
+}
+
 sed \
     -e "s#__OSAC_CONTROLLER_CLIENT_SECRET__#${CONTROLLER_SECRET}#" \
     -e "s#__OSAC_ADMIN_CLIENT_SECRET__#${ADMIN_SECRET}#" \
+    -e "s#__OSAC_REALM_ADMIN_USERNAME__#$(escape_sed_replacement "$(json_escape_string "${REALM_ADMIN_USERNAME}")")#" \
+    -e "s#__OSAC_REALM_ADMIN_PASSWORD__#$(escape_sed_replacement "$(json_escape_string "${REALM_ADMIN_PASSWORD}")")#" \
     "${RAW_REALM}" > "${RESOLVED_REALM}"
 
 echo "Realm secrets resolved -> ${RESOLVED_REALM}"
--- a/osac-installer/charts/osac-prereqs/templates/keycloak/resources.yaml
+++ b/osac-installer/charts/osac-prereqs/templates/keycloak/resources.yaml
@@ -372,6 +372,10 @@ spec:
           value: /realm-raw/realm.json
         - name: REALM_OUTPUT_PATH
           value: /realm/realm.json
+        - name: REALM_ADMIN_USERNAME
+          value: {{ .Values.keycloak.adminUsername | quote }}
+        - name: REALM_ADMIN_PASSWORD
+          value: {{ .Values.keycloak.adminPassword | quote }}
         volumeMounts:
         - name: scripts
           mountPath: /scripts
@@ -397,9 +401,9 @@ spec:
           readOnly: true
         env:
         - name: KEYCLOAK_ADMIN
-          value: admin
+          value: {{ .Values.keycloak.adminUsername | quote }}
         - name: KEYCLOAK_ADMIN_PASSWORD
-          value: admin
+          value: {{ .Values.keycloak.adminPassword | quote }}
         - name: KC_HTTP_ENABLED
           value: "false"
         - name: KC_HTTPS_ENABLED
@@ -499,33 +503,38 @@ data:
         -H 'Content-Type: application/x-www-form-urlencoded' \
         -d 'grant_type=password' \
         -d 'client_id=admin-cli' \
-        -d 'username=admin' \
-        -d 'password=admin' | python3 -c "import sys,json; print(json.load(sys.stdin)['access_token'])")
+        --data-urlencode "username=${ADMIN_USERNAME}" \
+        --data-urlencode "password=${ADMIN_PASSWORD}" | python3 -c "import sys,json; print(json.load(sys.stdin)['access_token'])")
       [[ -n "${TOKEN}" ]] || { echo "ERROR: Could not get admin token"; return 1; }
 
       USER_ID=$(curl -k -sf -X GET "https://keycloak:443/admin/realms/osac/users?username=$username" \
         -H "Authorization: Bearer $TOKEN" | python3 -c "import sys,json; print(json.load(sys.stdin)[0]['id'])")
       [[ -n "${USER_ID}" ]] || { echo "ERROR: Could not find user: $username"; return 1; }
 
+      # Build the JSON body with python3 (already used above for parsing)
+      # rather than hand-rolled bash string interpolation, so $password can
+      # safely contain quotes, backslashes, or other JSON-significant chars.
+      PAYLOAD=$(python3 -c "import json,sys; print(json.dumps({'type': 'password', 'value': sys.argv[1], 'temporary': False}))" "$password")
+
       curl -k -sf -X PUT "https://keycloak:443/admin/realms/osac/users/$USER_ID/reset-password" \
         -H "Authorization: Bearer $TOKEN" \
         -H "Content-Type: application/json" \
-        -d "{\"type\":\"password\",\"value\":\"$password\",\"temporary\":false}"
+        -d "$PAYLOAD"
 
       echo "Set password for $username"
     }
 
     # Set all passwords
     echo "Setting passwords for all tenant users..."
-    set_password "tenant1_admin" "foobar"
-    set_password "tenant1_user" "foobar"
-    set_password "tenant2_admin" "foobar"
-    set_password "tenant2_user" "foobar"
-    set_password "twotenant" "foobar"
-    set_password "my_user" "foobar"
-    set_password "your_user" "foobar"
+    set_password "tenant1_admin" "${DEFAULT_USER_PASSWORD}"
+    set_password "tenant1_user" "${DEFAULT_USER_PASSWORD}"
+    set_password "tenant2_admin" "${DEFAULT_USER_PASSWORD}"
+    set_password "tenant2_user" "${DEFAULT_USER_PASSWORD}"
+    set_password "twotenant" "${DEFAULT_USER_PASSWORD}"
+    set_password "my_user" "${DEFAULT_USER_PASSWORD}"
+    set_password "your_user" "${DEFAULT_USER_PASSWORD}"
 
-    echo "All passwords set to 'foobar'"
+    echo "All demo-user passwords set"
 ---
 apiVersion: batch/v1
 kind: Job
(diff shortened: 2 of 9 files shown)


amej:
@coderabbitai fullreview

amej:
## Fixes Applied Successfully

Fixed 4 file(s) based on 3 CodeRabbit feedback item(s).

**Files modified:**
- `osac-installer/charts/osac-prereqs/files/hooks/resolve-realm-secrets.sh`
- `osac-installer/prerequisites/keycloak/service/deployment.yaml`
- `osac-installer/prerequisites/keycloak/service/password-setup-job.yaml`
- `osac-installer/scripts/validate-keycloak-credentials.sh`

**Applied:**
- Incomplete JSON escaping (control characters like tabs/newlines) — replaced hand-rolled sed escaping with `python3 json.dumps` across all 3 affected sites, extended regression test coverage accordingly.
- Decoded passwords appearing in test failure output — removed from all 3 failure messages.

**Deferred (with reason):** the suggestion to source credentials via `secretKeyRef` instead of Helm values conflicts with an explicit, already-negotiated decision on this ticket to keep them as overridable-but-not-required Helm values — see the PR description.

**Commit:** `bfd5ba0b`

The latest autofix changes are on the `bugfix/OSAC-3610-hardcoded-keycloak-passwords` branch.

amej:
## CodeRabbit Autofix Review Complete

Reviewed 2 unresolved CodeRabbit feedback item(s) and did not apply new code changes in this run — both were stale duplicates of items already handled in the prior autofix pass (GitHub hadn't auto-marked them resolved after that push):

- **Incomplete JSON escaping** — already fixed (commit `bfd5ba0b`/`e85fedcf`), verified the fix is present in the current pushed code, marked resolved.
- **Credentials as Helm values vs. `secretKeyRef`** — already deliberately deferred with an explained reason (conflicts with an explicit, already-negotiated product decision on this ticket), marked resolved.

A third thread (decoded passwords in test output) was already `isOutdated` per GitHub and excluded automatically.

amej:
## Fixes Applied Successfully

Fixed 3 file(s) based on 5 CodeRabbit feedback item(s) (second review round).

**Files modified:**
- `osac-installer/charts/osac-prereqs/templates/keycloak/resources.yaml`
- `osac-installer/prerequisites/keycloak/service/password-setup-job.yaml`
- `osac-installer/scripts/validate-keycloak-credentials.sh`

**Applied:**
- Test coverage gap for the static `prerequisites/keycloak/service/` manifest path (its own `realm.json` copy and inline resolve-realm-secrets script had no direct test coverage before) — 2 new test sections added, verified they catch regressions.
- **Credentials now sourced via a chart-managed `keycloak-admin-credentials` Secret (`secretKeyRef`) instead of plaintext Helm values in pod specs.** This reconsiders the previous round's deferral — implemented in a way that preserves backward compatibility: `values.yaml`'s existing overridable-but-not-required defaults are unchanged, so `osac-test-infra`/CI values files remain unaffected (the actual credential values and override mechanism don't change, only the in-pod delivery mechanism).
- `curl` call in the static `password-setup-job.yaml` that silently swallowed HTTP errors on the reset-password request (missing `-f`, unlike the chart's already-correct `-sf`) — now uses `-fsS`.
- Added an upfront PyYAML dependency check in the test script with a clear error message instead of a raw traceback if missing.

**Commit:** `9543fad6`

The latest autofix changes are on the `bugfix/OSAC-3610-hardcoded-keycloak-passwords` branch.

amej:
> ### ❌ Failed checks (2 errors, 1 warning)
> Check name 	Status 	Explanation 	Resolution
> No-Hardcoded-Secrets 	❌ Error 	values.yaml adds adminPassword: admin and defaultUserPassword: foobar; the installer renders them as plaintext env values, outside the developer/maintenance admin/admin exception. 	Remove password literals from chart defaults and source all Keycloak credentials from Kubernetes Secrets or required external overrides.

on _origin/main_, the credentials were hardcoded literals baked directly into _resources.yaml_'s templates (value: _admin_ at lines 400/402, "_foobar_" literals at lines 520–526) — not exposed as Helm values at all. _values.yaml_ had zero visibility into these credentials.

This means CodeRabbit's "No-Hardcoded-Secrets" finding is actually flagging something this PR introduces as an improvement, not a regression:

Before (_main_): credentials hardcoded in templates, invisible in _values.yaml_, zero override path — exactly the bug [OSAC-3610](https://redhat.atlassian.net/browse/OSAC-3610) reports.
After (this PR): credentials moved into _values.yaml_ as explicit, named, documented defaults (_adminUsername: admin, adminPassword: admin, defaultUserPassword: foobar_), fully overridable via _--set/values files_, and no longer rendered as plaintext in pod specs (sourced via the _keycloak-admin-credentials_ Secret instead).

About the CodeRabbit's behaviour
The scanner is pattern-matching "a password literal exists in values.yaml" without the context that this is strictly better than the pre-PR state it's comparing against — the literals didn't disappear from main, they were relocated from an untouchable template into a documented, overridable, Secret-backed value. 

The finding reflects a deliberate, already-negotiated trade-off (not a required-Secret model, to avoid breaking osac-test-infra/CI), not a new regression worth blocking on.

> Comment `@coderabbitai help` to get the list of available commands.

@coderabbitai help

amej:
## Fixes Applied Successfully

Fixed 1 file based on 1 CodeRabbit feedback item (third review round).

**Files modified:**
- `osac-installer/scripts/validate-keycloak-credentials.sh`

**Applied:**
- Replaced weak substring checks in Test 7 (does `keycloak-admin-credentials` appear *anywhere* in the file?) and added missing coverage for the static Job manifest, with structural `assert_secret_key_ref` checks (7 new assertions total) that verify each specific credential env var points at the correct Secret name **and** key. Verified the new checks catch a real regression: deliberately pointed `KEYCLOAK_ADMIN` at the wrong key (`admin-password` instead of `admin-username`), confirmed the test failed, then restored it — the old substring check would have missed this exact bug.

**Commit:** `aafb79f2`

The latest autofix changes are on the `bugfix/OSAC-3610-hardcoded-keycloak-passwords` branch.

amej:
@coderabbitai full review

osac-installer/charts/osac-prereqs/values.yaml
```diff
@@ -15,6 +15,9 @@ keycloak:
   enabled: true
   realmOverwrite: true
   hostname: "keycloak.keycloak.svc.cluster.local"
+  adminUsername: admin
+  adminPassword: admin
+  defaultUserPassword: foobar
```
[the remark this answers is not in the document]
amej: @coderabbitai https://github.com/osac-project/osac/pull/239#issuecomment-5243719499

amej:
## Fixes Applied Successfully

Fixed 1 file based on 1 CodeRabbit feedback item.

**Files modified:**
- `osac-installer/scripts/validate-keycloak-credentials.sh`

**Commit:** `bb604dca`

The latest autofix changes are on the `bugfix/OSAC-3610-hardcoded-keycloak-passwords` branch.

osac-installer/charts/osac-prereqs/templates/keycloak/resources.yaml
amej, line 1: Default Helm  chart values overridden at the time of installation.

amej (commented):
@tzvatot and @ybettan : Please review.

amej:
@coderabbitai approve

CrystalChun (approved):
/lgtm

CrystalChun:
Unrelated to your changes, but wondering if we still need Keycloak in the `prerequisites` folder if it's in the osac-prereqs already

osac-installer/charts/osac-prereqs/files/hooks/resolve-realm-secrets.sh
```diff
@@ -21,9 +21,30 @@ ADMIN_SECRET=$(oc get secret "${SECRET_NAME}" -n "${NAMESPACE}" -o jsonpath='{.d
...
+# handled correctly, not just backslash/quote) since the raw value lands
+# inside a JSON string, then sed replacement-escaping (backslash, ampersand,
+# and the s### delimiter) so sed doesn't reinterpret characters introduced
+# by the JSON escaping.
+json_escape_string() {
+    python3 -c "import json,sys; print(json.dumps(sys.argv[1])[1:-1])" "$1"
+}
+
+escape_sed_replacement() {
+    printf '%s' "$1" | sed -e 's/[\&#]/\\&/g'
```
tzvatot, line 40: 🟡 **`escape_sed_replacement` bracket expression portability**

The character class `[\&#]` relies on backslash being literal inside `[...]` in sed. POSIX says this is undefined; GNU sed (the target platform) treats it as literal, so it works correctly here.

Since `json_escape_string()` already uses `python3 json.dumps` upstream, the input to this function will only contain JSON-escaped sequences (like `\\`, `\"`, `\n`, `\t`) - not raw control characters. The sed escaping is therefore correct in practice.

No change needed for RHEL/UBI containers. If portability ever matters, the two escaping functions could be collapsed into a single Python call that outputs a sed-safe string directly.

Same pattern is duplicated inline in `prerequisites/keycloak/service/deployment.yaml:76`.

osac-installer/scripts/validate-keycloak-credentials.sh
```diff
@@ -0,0 +1,342 @@
...
+  *"/users?username="*) echo '[{"id":"fake-user-id"}]' ;;
+  *) exit 0 ;;  # bare readiness-check GET (https://keycloak:443/realms/osac, no distinguishing flag)
+esac
+EOF
+chmod +x "${TMP_DIR}/curlbin/curl"
+
+run_set_password_script() {
+    local script_content="$1"
+    rm -f "${CAPTURE_FILE}"
+    echo "${script_content}" > "${TMP_DIR}/extracted-set-passwords.sh"
```
tzvatot, line 196: 💡 **`echo` for writing script content is fragile**

`echo` can misinterpret content starting with `-n`, `-e`, or `-E` as flags, and in some configurations interprets escape sequences. While the YAML-extracted content won't trigger these in practice, `printf '%s\n' "${script_content}"` is the unconditionally safe alternative.

Minor - test code only.

osac-installer/scripts/validate-keycloak-credentials.sh
```diff
@@ -0,0 +1,342 @@
...
+# (not a plaintext `value:`)? Uses PyYAML for a real structural check rather
+# than brittle multi-line grep -F (which silently splits multi-line needles
+# into OR'd single-line alternatives -- see Test 7's comment below).
+assert_secret_key_ref() {
+    local render="$1" kind="$2" resource_name="$3" container_name="$4" env_name="$5" expected_secret="$6" expected_key="$7" description="$8"
+    python3 -c "
+import yaml, sys
+docs = list(yaml.safe_load_all(sys.stdin.read()))
+for d in docs:
+    if not d or d.get('kind') != '${kind}' or d.get('metadata', {}).get('name') != '${resource_name}':
```
tzvatot, line 54: 💡 **Shell variables interpolated directly into Python source**

The helper passes `${kind}`, `${resource_name}`, `${container_name}`, etc. into a Python string via shell expansion. If any parameter contained a single quote, the Python code would break or potentially execute unintended code.

All current callers use hardcoded string literals (`Deployment`, `keycloak-service`, etc.), so this is safe today. For future-proofing, consider passing values via environment variables instead of string interpolation (e.g., `KIND="${kind}" python3 -c "import os; kind = os.environ['KIND']; ..."`).

Minor - test code only, all callers are hardcoded.

tzvatot (commented):
Well-executed credential parameterization that solves a real security/usability blocker for Demo Platform deployments. The approach is sound: chart-managed `keycloak-admin-credentials` Secret with `secretKeyRef` delivery, proper JSON/sed escaping via `python3 json.dumps`, and comprehensive regression tests (342 lines, 7 test sections) covering both chart and static manifest paths. The author iterated through 4 rounds of CodeRabbit feedback - addressing escaping gaps, switching to Secret-based delivery, adding trailing-newline handling, and strengthening structural assertions. Live-cluster verification was completed twice. No blocking issues.

| Category | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟡 Important | 1 |
| 💡 Suggestion | 2 |

PR #250, opened 2026-08-10
NO-ISSUE: fix - echo-adapter returns full CloudEvent instead of summary

omer-vishlitzky:
## Summary

- Echo-adapter's `storedEvent` stripped CloudEvent data (extensions, payload) down to a summary — only `id`, `type`, `source`, `time`, `resource_id` were kept
- E2E tests need the full event to validate `billing_dimensions`, `transition_time`, `previous_state`, `osactenant`, `osacresourcetype`, etc.
- Store the raw CloudEvent JSON merged with Kafka metadata, matching the original test-adapter design from PR #135

## What changed

`storedEvent` now stores the full CloudEvent as `json.RawMessage` (serialized via CloudEvents SDK, then enriched with `topic`, `partition`, `offset`, `received_at`). Server-side filtering uses indexed fields extracted at insert time. HTTP responses return the full event, not a summary.

## Test plan

- [x] `go build ./cmd/echo-adapter/` passes
- [x] `make test` — 67 specs pass
- [x] `go vet ./cmd/echo-adapter/` clean

## Summary by CodeRabbit

- **New Features**
  - Event responses now return the original CloudEvent JSON directly.
  - Event listing and lookup provide consistent CloudEvent-formatted data.
  - Filtering and result counts continue to use event type, resource ID, and receipt time.

- **Bug Fixes**
  - Events that cannot be serialized as valid JSON are safely skipped instead of being stored or returned.

### DIFF
--- a/osac-metering/adapters/cmd/echo-adapter/store.go
+++ b/osac-metering/adapters/cmd/echo-adapter/store.go
@@ -12,6 +12,7 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"log"
 	"net/http"
 	"strconv"
 	"sync"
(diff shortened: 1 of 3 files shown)


masayag (approved):

osac-metering/adapters/cmd/echo-adapter/store.go
```diff
@@ -60,45 +61,48 @@ func (s *eventStore) clear() {
...
+// CloudEvent JSON as-is. Kafka metadata (topic, partition, offset) is
+// kept in indexed fields for filtering but not merged into the JSON
+// to avoid overwriting CloudEvent attributes or introducing invalid
+// extension names.
 func (s *eventStore) add(event adapters.MeteringEvent) {
 	ce := event.CloudEvent
 
+	raw, err := json.Marshal(ce)
+	if err != nil {
+		return
```
amito, line 88: We're losing an event here. Let's at least log the error to figure out what went wrong.

amito (commented):
Looks good, one really small comment. Thanks

PR #251, opened 2026-08-10
OSAC-2969: vault tenant onboarding

DakCrowder:
This PR sets up lifecycle operations around vault namespaces for tenants - namely creating a namespace within vault for each tenant, ensuring policies are created per tenant to allow for future secret access, and wiring things up e2e so the tenant controllers can do this fully against a configured vault compatible secret store.

There is a configurable top level namespace in Vault (e.g. /osac) that then contains tenants like so:
```bash
/osac/tenant-a
/osac/tenant-b
```

When a tenant is onboarded, the tenant onboarding controller now creates the namespace and various other required resources (e.g. ensures secrets can be created, set up auth etc.).  The lifecycle client is configured to use keycloak and a client credentials grant to access vault and manage the tenant resources.  All necessary vault flags for this have been added to the charts and controller but currently kept optional.

Please note that individual secret creation for a tenant or tenant user (e.g. creating or fetching a secret for a tenant) is NOT yet supported - this PR is a prerequisite that will allow for the full flow to be implemented next.

## Summary by CodeRabbit

* **New Features**
  * Added optional Vault integration for tenant secret-store namespaces.
  * Tenant synchronization provisions Vault namespaces, while tenant deletion cleans them up.
  * Added Keycloak-based authentication with cached tokens and automatic renewal.
  * Added Vault lifecycle configuration, including JWT, TLS certificate, and credential settings.
  * Tenant status now reports Vault namespace provisioning readiness.
  * Vault credentials are renewed automatically for secret-store operations.

* **Documentation**
  * Updated the documented default OpenBao image version.


DakCrowder:
@coderabbitai review

fulfillment-service/internal/controllers/tenant/tenant_reconciler_function.go
```diff
@@ -423,6 +451,68 @@ func (t *task) updateCondition(conditionType privatev1.TenantConditionType, stat
...
+
+	err := t.r.vaultLifecycle.EnsureTenantNamespace(ctx, tenantName)
+	if err != nil {
+		t.r.logger.ErrorContext(ctx, "Failed to provision vault namespace for tenant",
+			slog.String("tenant_id", t.tenant.GetId()),
+			slog.String("tenant_name", tenantName),
+			slog.Any("error", err),
+		)
+		return fmt.Errorf("failed to provision vault namespace: %w", err)
+	}
```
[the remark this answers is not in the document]
DakCrowder: The namespace write is actually persisted at a different layer (in the Run method) - the error call propagates up and the Update() call for the tenant is not persisted unless the oparation is successful.

fulfillment-service/internal/controllers/tenant/tenant_reconciler_function.go
```diff
@@ -229,6 +243,10 @@ func (t *task) syncToIDP(ctx context.Context) error {
 		slog.String("tenant_name", tenantName),
 	)
 
+	if err := t.ensureVaultNamespace(ctx); err != nil {
```
danielerez: I think it might be redundant here. State is already set at line 236, so if vault fails here, the next reconcile takes the SYNCED path and calls ensureVaultNamespace again anyway.

fulfillment-service/charts/service/templates/controller/deployment.yaml
```diff
@@ -134,6 +164,21 @@ spec:
         - --grpc-listener-tls-crt=/etc/fulfillment-controller/tls/tls.crt
         - --grpc-listener-tls-key=/etc/fulfillment-controller/tls/tls.key
         - --metrics-listener-address=0.0.0.0:8002
+        {{ if and .Values.vault.endpoint .Values.vault.credentials }}
```
danielerez: nit: should be with '-'
```
{{- if
```
to avoid blank lines in the rendered args list

fulfillment-service/internal/vault/vault_authenticator.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+	if err = validatePathComponent(b.vaultAuthMountPath, "vault auth mount path"); err != nil {
+		return
+	}
+
+	httpClient := &http.Client{
+		Timeout: 30 * time.Second,
+	}
+	if b.caPool != nil {
+		transport, ok := http.DefaultTransport.(*http.Transport)
```
danielerez, line 159: just for consistency, maybe use "vaultapi.DefaultConfig()..." everywhere

fulfillment-service/internal/controllers/tenant/tenant_reconciler_function.go
```diff
@@ -349,6 +373,10 @@ func (t *task) delete(ctx context.Context) error {
 		slog.String("idp_name", tenantName),
 	)
 
+	if err := t.deleteVaultNamespace(ctx); err != nil {
```
danielerez: hmm, just to be safe, maybe worth deleting the namespace before the tenant (line 366)?
I mean, if vault fails then IDP is still intact and no work is lost on retry (not critical if IDP delete is idempotent...)

DakCrowder: Ah nice catch, better to do in that order

fulfillment-service/internal/controllers/tenant/tenant_reconciler_function.go
DakCrowder: It does then require an additional sync to get the vault data created but it looks like it will happen almost immediately - will remove this to be more inline with how the networking status is checked.

fulfillment-service/internal/controllers/tenant/tenant_reconciler_function.go
```diff
@@ -326,15 +344,21 @@ func (t *task) delete(ctx context.Context) error {
 		return fmt.Errorf("tenant still has %d project(s) pending deletion", remaining)
 	}
 
-	// Skip if not synced to IDP yet
+	// Skip IDP cleanup if not synced to IDP yet, but still clean up vault.
 	if t.tenant.GetStatus().GetState() != privatev1.TenantState_TENANT_STATE_SYNCED {
+		if err := t.deleteVaultNamespace(ctx); err != nil {
```
CrystalChun, line 345: Should we check the vault condition here to see if we need to clean up vault?

fulfillment-service/internal/vault/vault_authenticator.go
DakCrowder: This is setting up something slightly different, the vaultapi.DefaultConfig is provided by that package to hit vault itself (so requests made against the vault api).  Here we're setting up a client to hit keycloak in preparation of a token exchange, so a slightly different need/flow than that other client.

fulfillment-service/internal/vault/vault_lifecycle.go
```diff
@@ -0,0 +1,368 @@
...
+		return fmt.Errorf("failed to create namespace %q: %w", tenantName, err)
+	}
+	return nil
+}
+
+func (c *VaultLifecycleClient) mountKV(ctx context.Context, client *vaultapi.Client,
+	tenantName string) error {
+	err := client.Sys().MountWithContext(ctx, c.kvMountPath, &vaultapi.MountInput{
+		Type:    "kv",
+		Options: map[string]string{"version": "2"},
```
CrystalChun, line 255: Maybe naïve question, but will this always be kv version 2?

fulfillment-service/internal/controllers/tenant/tenant_reconciler_function.go
DakCrowder: We can - but the deleteVaultNamespace should be idempotent and safe to call regardless.  Given that tenant deletion is (presumably) a pretty rare operation I think I'd rather lean on calling than trying to rely on derived state.  The state _should_ always be correct for triggering this op but might be slightly safer to lean on calling.

fulfillment-service/internal/vault/vault_lifecycle.go
DakCrowder: Not naïve at all :) - there are 2 KV versions vault provides: an older kv1 and a newer kv2.  The kv1 is simpler (straight up key value store) but kv2 gives history (through versioning) and provides an audit trail amongst some other features.  Also allows us to use those versioning/other features if needed in the future (secret rotation I know has been mentioned but not planned yet) and makes it easier to do.

https://developer.hashicorp.com/vault/docs/secrets/kv#kv-version-2

fulfillment-service/internal/controllers/tenant/tenant_reconciler_function.go
CrystalChun: Ah ok that makes sense! Thanks for the explanation

fulfillment-service/internal/vault/vault_lifecycle.go
CrystalChun: Good to know, thanks for the explanation!

CrystalChun (approved):
/lgtm

PR #261, opened 2026-08-11
OSAC-3766: Implement FindFreeHost for BCM inventory backend

mennyaboush:
## Summary

Implements `FindFreeHost` for the BCM inventory backend ([OSAC-3766](https://redhat.atlassian.net/browse/OSAC-3766), Phase 2 Task 3).

**Stacked on:** [osac#353](https://github.com/osac-project/osac/pull/353) ([OSAC-4067](https://redhat.atlassian.net/browse/OSAC-4067): BCM inventory adapter) — merge #353 first, then rebase this PR onto main.

Builds on the BCM inventory adapter introduced in [OSAC-4067](https://redhat.atlassian.net/browse/OSAC-4067):
- Queries BCM via `bcmclient.GetDevices`, filters client-side (BCM has no server-side filtering)
- **Filters**: LiteNode only, `extra_values` present with `resource_class`, not assigned (no `osac_instance_id`), valid MAC address, valid Kubernetes hostname (RFC 1123), no `/` in hostname
- Each exclusion produces a warning log identifying the hostname and reason
- Candidates shuffled randomly to reduce contention
- Returns `nil, nil` when no matching free host is found
- `InventoryHostID` constructed as `namespace/hostname` using `bmhManager.Namespace()`
- `osac_bcm_hosts_available` gauge metric updated per `host_type` on each call

**Depends on:**
- [osac#353](https://github.com/osac-project/osac/pull/353) ([OSAC-4067](https://redhat.atlassian.net/browse/OSAC-4067): BCM inventory adapter)
- [osac#228](https://github.com/osac-project/osac/pull/228) ([OSAC-3759](https://redhat.atlassian.net/browse/OSAC-3759): BCM HTTP client) ✅ merged
- [osac#237](https://github.com/osac-project/osac/pull/237) ([OSAC-3764](https://redhat.atlassian.net/browse/OSAC-3764): BMH lifecycle manager) ✅ merged

## Test plan

- [x] `go build ./...` passes
- [x] `make test` passes (12 new FindFreeHost tests)
- [x] `make lint` passes (0 issues)
- [x] Tests cover: matching free host, skip PhysicalNode, skip nil extra_values, skip missing resource_class, skip assigned host, skip `/` in hostname, skip invalid K8s hostname, skip missing/malformed MAC, hostType filter, empty result, empty hostType

### DIFF
--- a/bare-metal-fulfillment-operator/internal/inventory/bcm.go
+++ b/bare-metal-fulfillment-operator/internal/inventory/bcm.go
@@ -20,12 +20,19 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math/rand"
 	"path/filepath"
+	"regexp"
 	"strings"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"sigs.k8s.io/controller-runtime/pkg/certwatcher"
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
 
 	"github.com/osac-project/osac/bare-metal-fulfillment-operator/internal/baremetalhost"
+	"github.com/osac-project/osac/bare-metal-fulfillment-operator/internal/bcmclient"
+	"github.com/osac-project/osac/bare-metal-fulfillment-operator/internal/shared"
 )
 
 const certBaseDir = "/etc/osac/certs"
@@ -35,9 +42,27 @@ const certBaseDir = "/etc/osac/certs"
 // without depending on the bcmclient package.
 type BCMAPI interface {
 	CertWatcher() *certwatcher.CertWatcher
+	GetDevices(ctx context.Context) ([]bcmclient.Device, error)
 }
 
-var _ Client = (*BCMClient)(nil)
+var (
+	_ Client = (*BCMClient)(nil)
+
+	macPattern      = regexp.MustCompile(`^[0-9a-fA-F]{2}(:[0-9a-fA-F]{2}){5}$`)
+	hostnamePattern = regexp.MustCompile(`^[a-z0-9]([a-z0-9\-]*[a-z0-9])?$`)
+
+	bcmHostsAvailable = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "osac_bcm_hosts_available",
+			Help: "Number of unassigned BCM LiteNodes by host type",
+		},
+		[]string{"host_type"},
+	)
+)
+
+func init() {
+	metrics.Registry.MustRegister(bcmHostsAvailable)
+}
 
 // BCMClientConfig holds the BCM-specific options parsed from the inventory config.
 type BCMClientConfig struct {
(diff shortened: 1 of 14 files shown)


bare-metal-fulfillment-operator/internal/inventory/bcm.go
```diff
@@ -119,9 +137,103 @@ func NewBCMClientForTest(client *bcmclient.Client, bmhManager *baremetalhost.Man
...
+	hostType := matchExpressions["hostType"]
+
+	candidates := make([]bcmclient.Device, 0, len(devices))
+	for _, d := range devices {
+		if d.ChildType != "LiteNode" {
+			continue
+		}
+
+		if d.ExtraValues == nil {
+			log.V(1).Info("Skipping BCM device with nil extra_values — set resource_class and osac_bmc_credentials_secret",
```
adriengentil: maybe info log is too much?

bare-metal-fulfillment-operator/internal/inventory/bcm.go
```diff
@@ -119,9 +137,103 @@ func NewBCMClientForTest(client *bcmclient.Client, bmhManager *baremetalhost.Man
...
+
+	matchManagedBy := matchExpressions["managedBy"]
+	if matchManagedBy == "" {
+		matchManagedBy = shared.OsacDefaultManagedByValue
+	}
+	if matchManagedBy != shared.OsacDefaultManagedByValue {
+		return nil, nil
+	}
+
+	devices, err := c.client.GetDevices(ctx)
```
adriengentil, line 145: we always need to query all devices? There's no pagination?

bare-metal-fulfillment-operator/internal/inventory/bcm.go
```diff
@@ -119,9 +137,103 @@ func NewBCMClientForTest(client *bcmclient.Client, bmhManager *baremetalhost.Man
...
+
+		if hostType != "" && resourceClass != hostType {
+			continue
+		}
+
+		if _, assigned := d.ExtraValues[bcmclient.ExtraValueInstanceID]; assigned {
+			continue
+		}
+
+		if strings.Contains(d.Hostname, "/") {
```
adriengentil: why this check?

bare-metal-fulfillment-operator/internal/inventory/bcm.go
```diff
@@ -119,9 +137,103 @@ func NewBCMClientForTest(client *bcmclient.Client, bmhManager *baremetalhost.Man
...
+			log.V(1).Info("Skipping BCM device with '/' in hostname",
+				"hostname", d.Hostname)
+			continue
+		}
+
+		if !hostnamePattern.MatchString(d.Hostname) || len(d.Hostname) > 63 {
+			log.V(1).Info("Skipping BCM device with invalid hostname — must be a valid Kubernetes name",
```
adriengentil: doesn't bcm enforce hostname? this constraint is not unique to k8s

bare-metal-fulfillment-operator/internal/inventory/bcm.go
```diff
@@ -119,9 +137,103 @@ func NewBCMClientForTest(client *bcmclient.Client, bmhManager *baremetalhost.Man
...
+				"hostname", d.Hostname)
+			continue
+		}
+
+		if !macPattern.MatchString(d.MAC) {
```
adriengentil: doesn't bcm enforce already enforce mac address format?

bare-metal-fulfillment-operator/internal/inventory/bcm.go
mennyaboush: our flow depends on InventoryHostId = namespace/hostname

To avoid bugs we need to validate there is no / in the name. 
telling that and review that comment I understand that we filttered name with / with the regex so this block redundant

bare-metal-fulfillment-operator/internal/inventory/bcm.go
mennyaboush: unfortunately no... that why we limits the length response to avoid OOM

bare-metal-fulfillment-operator/internal/inventory/bcm.go
mennyaboush: right. will remove it.

bare-metal-fulfillment-operator/internal/inventory/bcm.go
mennyaboush: BCM likely enforces DNS-safe hostnames, but K8s object names are stricter than DNS (lowercase required). We validate here to catch it early instead of failing at BMH creation time.

bare-metal-fulfillment-operator/internal/inventory/bcm.go
mennyaboush: BCM does not enforce MAC format - it's optional and accepts all-zeros as a placeholder. We validate because Metal3 BMH needs a real MAC

bare-metal-fulfillment-operator/internal/inventory/bcm.go
adriengentil: ok we'll see how it behaves, but we might need a caching layer at some point

bare-metal-fulfillment-operator/internal/inventory/bcm.go
adriengentil: hostname shouldn't be case sensitive, can we just lower case the hostname? if that's the only limitation

adriengentil (approved):

PR #268, opened 2026-08-11
OSAC-3841: Vendor network-runner 0.3.8: serialize Cumulus NVUE operations

ybettan:
Fixes concurrent nv config apply race condition and idempotency on retries. See network-runner 0.3.8 for details.

---

/assign @danmanor


danmanor (approved):

ybettan:
The `pre-commit` job is a false positive. It doesn't seems to block the PR from merging so I will leave it as is.
```
  1. `pre-commit/pre-commit` — Gitleaks detected a false positive: the Codecov badge token (XvFwDpezAH) in the vendored file osac-aap/vendor/ansible_collections/ansible/eda/README.md:7. This is a read-only badge token belonging to the upstream ansible/event-driven-ansible project. Gitleaks exits with code 1 when it finds any match, which fails the step.
  2. `pre-commit/gitleaks-rotation-comment` — This is a downstream step that runs because gitleaks found something. It tries to post a warning comment on the PR, but gets a 403 "Resource not accessible by integration" because the workflow's
  GITHUB_TOKEN doesn't have pull_requests: write permission for this PR.
```

PR #269, opened 2026-08-11
OSAC-3872: Add hairpin MASQUERADE for ingress DNAT on tenant VLAN

ybettan:
When the DNAT target (MetalLB ingress VIP) is on the same VLAN subnet
as the source worker, the packet hairpins back out the VLAN sub-interface
without hitting the veth MASQUERADE. The response bypasses the namespace
conntrack and the DNAT is never reversed, breaking the console operator
health check and any pod-to-ingress traffic via the public IP.

Add a MASQUERADE rule on the VLAN sub-interface inside each per-cluster
namespace so the source is rewritten to the gateway IP, forcing the
response back through the namespace where conntrack can reverse the DNAT.

---

/assign @danmanor

### DIFF
--- a/osac-aap/collections/ansible_collections/agentless_net/steps/roles/external_access/playbooks/l3_create_dnat.yaml
+++ b/osac-aap/collections/ansible_collections/agentless_net/steps/roles/external_access/playbooks/l3_create_dnat.yaml
@@ -28,3 +28,31 @@
         -o {{ _dnat_veth_iface }} -j MASQUERADE
       when: _veth_masq_check.rc != 0
       changed_when: true
+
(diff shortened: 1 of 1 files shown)


ybettan:
/hold

Should wait for https://github.com/osac-project/osac/pull/266 to get merged and then rebase this PR on top of it (the first commit in this PR belongs to https://github.com/osac-project/osac/pull/266)

danmanor (approved):

ybettan:
/unhold

https://github.com/osac-project/osac/pull/266 has been merged.

danmanor (approved):

PR #272, opened 2026-08-11
OSAC-3468: Signal project after project membership deletion

CrystalChun:
Allows the project to be re-reconciled after a project membership is deleted to unblock the project's deletion.

## Summary by CodeRabbit

- **Bug Fixes**
  - Project membership deletion now notifies the associated project after cleanup.
  - Deletion completes successfully even when project lookup or notification encounters an error.
  - Notifications are handled for nested projects and memberships without authorization groups.
  - Notifications are skipped when no project name is available.
  - Membership finalizers are removed reliably in all supported deletion scenarios.

### DIFF
--- a/fulfillment-service/internal/controllers/projectmembership/project_membership_reconciler_function.go
+++ b/fulfillment-service/internal/controllers/projectmembership/project_membership_reconciler_function.go
@@ -603,6 +603,7 @@ func (t *task) delete(ctx context.Context) error {
 		if err := t.cleanupProjectMembership(ctx); err != nil {
 			return err
 		}
+		t.signalProject(ctx)
 	}
 
 	metadata.SetFinalizers(updated)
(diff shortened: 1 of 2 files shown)


fulfillment-service/internal/controllers/projectmembership/project_membership_reconciler_function.go
```diff
@@ -664,3 +665,31 @@ func (t *task) cleanupProjectMembership(ctx context.Context) error {
 
 	return nil
 }
+
+// signalProject looks up the project by name and signals it so that the
+// project reconciler re-runs. This is used when the project membership is deleted to
```
DakCrowder, line 670: nit/def not blocking: I don't know if we need to specify callers in a method docstring, just describing signalProject should be enough and then makes it more re-usable.

DakCrowder (approved):

PR #277, opened 2026-08-11
OSAC-3954: populate Name on resolved ClusterTemplateReference and HostTypeReference

omer-vishlitzky:
## Summary

- Set `Name` (from `metadata.name`) alongside `Id` on resolved `ClusterTemplateReference` and `HostTypeReference` in `validateAndTransformCluster`
- Previously only `Id` was set, leaving `Name` empty — metering reads `Name` for the `cluster_template` billing dimension, which was always `""`
- Fix covers all three resolution sites: create path (line 760), signal/update path (line 198), and update validation path (line 986)
- Added test assertions verifying `Name` is preserved after reference resolution

## Root Cause

`private_clusters_server.go` resolves user-provided template/host-type references to their canonical form (by ID). The resolved `ClusterTemplateReference` was constructed with only `SetId()`, dropping the `Name` field. Downstream, `osac-metering`'s `ClusterBillingDimensions()` reads `spec.GetTemplate().GetName()` for the `cluster_template` billing dimension — always empty.

## Test plan

- [x] Existing cluster tests pass (`ginkgo run --focus="Cluster" internal/servers` — 115 passed)
- [x] Added assertions in "Creates object with template specified by name" and "Creates object with template and host type specified by name" tests verifying `GetName()` returns expected value after resolution
- [ ] CaaS metering E2E tests (osac-test-infra PR #349) should pass once this lands in the image

## Summary by CodeRabbit

* **Bug Fixes**
  * Private cluster creation now consistently retains the names of selected templates and host types, whether chosen by name, ID, or through a catalog item.
  * Catalog-item-based creation now correctly records the resolved template details before applying defaults.

### DIFF
--- a/fulfillment-service/internal/servers/private_clusters_server.go
+++ b/fulfillment-service/internal/servers/private_clusters_server.go
@@ -944,6 +944,11 @@ func (s *PrivateClustersServer) validateAndTransformCatalogItem(ctx context.Cont
 		return grpcstatus.Errorf(grpccodes.InvalidArgument, "template '%s' has been deleted", templateRefStr)
 	}
 
+	resolvedTemplateRef := &privatev1.ClusterTemplateReference{}
+	resolvedTemplateRef.SetId(template.GetId())
+	resolvedTemplateRef.SetName(template.GetMetadata().GetName())
+	cluster.GetSpec().SetTemplate(resolvedTemplateRef)
+
 	// Apply spec defaults from the template (user and field_definition values take precedence):
 	utils.ApplyClusterSpecDefaults(cluster.GetSpec(), template.GetSpecDefaults())
 
--- a/fulfillment-service/internal/servers/private_clusters_server.go
+++ b/fulfillment-service/internal/servers/private_clusters_server.go
@@ -196,6 +196,7 @@ func (s *PrivateClustersServer) Create(ctx context.Context,
 		if hostType != nil {
 			hostTypeRef := &privatev1.HostTypeReference{}
 			hostTypeRef.SetId(hostType.GetId())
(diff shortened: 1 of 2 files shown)


masayag (approved):
Reviewed with a focus on whether `hostType.GetMetadata()`/`template.GetMetadata()` can ever be nil across the resolution paths touched here (`Create`, `validateAndTransformCluster`, `validateAndTransformCatalogItem`).

Confirmed safe in all flows:
- All four `GetMetadata()` call sites are already behind an explicit `hostType != nil` / `template != nil` check.
- Even without that, the generated protobuf accessors (`GetMetadata()`, `GetName()`) are nil-receiver safe (`if x != nil { ... }; return nil/""`), so the chain can't panic regardless.
- `generic_dao_get.go` / `generic_dao_list.go` always construct and attach a non-nil `Metadata` for any row successfully hydrated from the DB, so a non-nil `hostType`/`template` returned from `lookupHostType`/`lookupTemplate` always has real metadata, not just a nil-safe empty string.

One non-blocking nit:

`validateTemplateImmutability()`'s comment (around L598-600, unchanged by this PR) says:

> // Compare by refKey (Id) rather than proto.Equal because validateAndTransformCluster normalizes the stored reference to Id-only, while the reference validator interceptor may backfill Name on incoming requests.

That's now stale — after this change, `validateAndTransformCluster`/`validateAndTransformCatalogItem` set both `Id` *and* `Name` on the resolved reference, not Id-only. Behavior is unaffected (`refKey()` prefers `Id` when present), but worth a follow-up comment update so it doesn't mislead future readers.

LGTM otherwise.

PR #286, opened 2026-08-12
OSAC-3985: requeue instead of firing CaaS storage job when no tier definitions registered

zszabo-rh:
## Summary

- On fresh install, `osac-create-tenant-cluster-storage` fired before any StorageTiers were registered, producing a predictable AAP job failure (`storage_tier_definitions must be non-empty`). Self-healed on retry but generated noise.
- Add a 4-line guard in `handleUpdate()`: if `TiersClient` is configured but tier definitions are empty, requeue with `StatusPollInterval` instead of firing the job.
- New test case in Stage 2 context: verifies no job is fired and `RequeueAfter = pollInterval` when TiersClient returns empty tier list.

Fixes: [OSAC-3985](https://redhat.atlassian.net/browse/OSAC-3985)

## Test plan

- [x] `make test` — 633 passed, 0 failed
- [x] `make lint` — 0 issues
- [ ] Observed on fresh install: no failed AAP job before StorageTiers register

## Breaking changes

- [ ] This PR does not introduce breaking changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved storage provisioning when tenant-specific storage tier definitions are unavailable.
  * The system now periodically retries instead of starting cluster-storage provisioning prematurely.
  * Prevents cluster-storage jobs from being created until required tier information is available.
  * Handles missing tier definitions and tier-resolution failures without reporting a reconciliation error.
  * Improves recovery as tier information becomes available or fulfillment services resume.

### DIFF
--- a/osac-operator/internal/controller/storage_controller.go
+++ b/osac-operator/internal/controller/storage_controller.go
@@ -370,9 +370,9 @@ func (r *StorageReconciler) handleUpdate(ctx context.Context, instance *v1alpha1
 				{ClusterName: clusterName, Ready: false, Reason: v1alpha1.TenantReasonNotFound},
 			}
 
-			if r.TiersClient != nil && len(tierDefinitions) == 0 {
+			if r.TiersClient != nil && r.BackendsClient != nil && len(tierDefinitions) == 0 {
 				log.Info("no storage tier definitions registered yet, requeueing",
-					"tenant", tenantName, "requeueAfter", r.StatusPollInterval)
+					"tenant", tenantName)
 				return ctrl.Result{RequeueAfter: r.StatusPollInterval}, nil
 			}
 
--- a/osac-operator/internal/controller/storage_controller.go
+++ b/osac-operator/internal/controller/storage_controller.go
@@ -370,6 +370,12 @@ func (r *StorageReconciler) handleUpdate(ctx context.Context, instance *v1alpha1
 				{ClusterName: clusterName, Ready: false, Reason: v1alpha1.TenantReasonNotFound},
 			}
 
+			if r.TiersClient != nil && len(tierDefinitions) == 0 {
+				log.Info("no storage tier definitions registered yet, requeueing",
+					"tenant", tenantName, "requeueAfter", r.StatusPollInterval)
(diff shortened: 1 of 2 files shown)


osac-operator/internal/controller/storage_controller.go
```diff
@@ -356,6 +356,12 @@ func (r *StorageReconciler) handleUpdate(ctx context.Context, instance *v1alpha1
 				{ClusterName: clusterName, Ready: false, Reason: v1alpha1.TenantReasonNotFound},
 			}
 
+			if r.TiersClient != nil && len(tierDefinitions) == 0 {
```
akshaynadkarni: Nit: we only fetch tier definitions when both TiersClient and BackendsClient are set (see resolveAndInjectTierContext). This guard only checks TiersClient.

So if TiersClient is set but BackendsClient is nil, tierDefinitions is empty because we never called the API — not because no tiers exist. We'd still requeue here and never provision.

In production both clients are wired together in main.go, so this probably won't happen. But for consistency, worth checking both:

```suggestion
			if r.TiersClient != nil && r.BackendsClient != nil && len(tierDefinitions) == 0 {
```

osac-operator/internal/controller/storage_controller_test.go
```diff
@@ -415,6 +415,38 @@ var _ = Describe("Storage Controller", func() {
 	})
 
 	Context("Stage 2: Cluster storage provisioning", func() {
+		It("should requeue without firing a job when TiersClient is set but no tier definitions are registered", func() {
```
akshaynadkarni, line 453: Nit: this covers empty tier list. Same guard also fires when tier resolution errors — resolveAndInjectTierContext clears tierDefinitions on failure, so len == 0 there too. We test API failure when SCs exist, but not when we'd trigger provisioning (no SCs + clusterProvider).

akshaynadkarni (approved):
Left a couple of nits. 
Overall lgtm.

akshaynadkarni (approved):
/lgtm

zszabo-rh:
Confirmed that the E2E CaaS failure is related to the fix.

The original intention was just to add a small patch to improve CI stability, but at this point it seems like a high-cost, low-reward task and probably not worth fixing.

PR #295, opened 2026-08-12
OSAC-3116: update docs for typed resource references

danielerez:
Summary

  - Replace stale "Object references" section in API.md that described references as plain string
  fields (and mentioned typed references as a "future plan") with comprehensive documentation covering
  full vs local references, wire format, cross-project lookups, server-side validation/resolution,
  error format, and complete reference type inventory
  - Add "Filtering on reference fields" section to FILTER.md with examples for
  this.spec.virtual_network.name, this.spec.catalog_item.name, etc.
  - Fix stale YAML examples in CATALOG_ITEMS.md — template: "osac.templates.sandbox" → template: { 
  name: sandbox, shared: true }, host_type: fc430 → host_type: { name: fc430 }
  - Fix stale gRPC request/response example in OSAC-CLI-HOWTO.md to use nested reference objects and
  proper spec/status structure

  Test plan

  - [x] Verify API.md reference type tables match current proto definitions
  - [x] Verify FILTER.md examples match actual CEL filter paths used in server code (confirmed against
  private_subnets_server.go, catalog_item_reference_checker.go)
  - [x] Verify CATALOG_ITEMS.md YAML examples are valid ProtoJSON for the current proto schema
  (confirmed ClusterTemplateReference, HostTypeReference, ComputeInstanceTemplateReference field
  shapes)
  - [x] Verify OSAC-CLI-HOWTO.md request/response example reflects actual wire format

## Summary by CodeRabbit

* **Documentation**
  * Updated API examples to use typed references, including names, IDs, projects, and shared scope.
  * Added guidance and examples for filtering typed reference fields.
  * Revised catalog item examples to use structured template and host type references.
  * Updated cluster creation and response examples to reflect nested specification and status fields.

### DIFF
--- a/fulfillment-service/docs/API.md
+++ b/fulfillment-service/docs/API.md
@@ -431,7 +431,7 @@ The URL space must never be deeper than a collection and its members. For exampl
 belongs to a `VirtualNetwork`, do not define a nested URL like
 `/api/fulfillment/v1/virtual_networks/123/subnets`. Instead, `Subnet` has its own top-level
 collection at `/api/fulfillment/v1/subnets`, and users find the subnets of a virtual network using
-the filter capability with a CEL expression like `this.spec.virtual_network == "123"`.
+the filter capability with a CEL expression like `this.spec.virtual_network.name == "my-vnet"`.
 
 ### HTTP verb mapping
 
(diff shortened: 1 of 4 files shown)


fulfillment-service/docs/API.md
```diff
@@ -492,30 +492,170 @@ top-level `state` enum.
...
+  bool shared = 4;
+}
+```
+
+| Field     | Description |
+|-----------|-------------|
+| `id`      | Unique identifier of the referenced object. |
+| `name`    | Human-readable name of the referenced object. |
+| `project` | Project where the referenced object lives. Defaults to the caller's project when omitted. |
+| `shared`  | When `true`, the lookup targets the `shared` tenant (overrides the caller's tenant). |
```
jhernand, line 520: Not part of this commit, but I wonder why we added a boolean instead of `string tenant = 4`. Do you know?

fulfillment-service/docs/API.md
```diff
@@ -492,30 +492,170 @@ top-level `state` enum.
...
+   `project`/`shared` overrides for full references).
+2. Looks up the referenced object by `id`, `name`, or both.
+3. If both `id` and `name` are provided, verifies they refer to the same object.
+4. Auto-populates whichever of `id` or `name` was not provided by the caller.
+
+Invalid references produce an `InvalidArgument` error with `google.rpc.BadRequest` details
+containing one `FieldViolation` per invalid reference. The `field` value is the dot-separated
+path to the reference field (e.g., `object.spec.template`, `object.spec.network_attachments[0].subnet`).
+
+### Reference types in the API
```
jhernand: This is going to eventually outdated, and it doesn't provide much value: if someone is interested they can grep the source to find them. I'd suggest to remove this section.

fulfillment-service/docs/API.md
```diff
@@ -492,30 +492,170 @@ top-level `state` enum.
...
 
-When an object needs to reference another object, define a `string` field named after the
-relationship or the type of the referenced object. The value of the field is the unique identifier
-of the referenced object.
+When an object needs to reference another object, the API uses typed reference messages instead of
+plain string fields. Each referenced type has its own message, giving the schema full type safety
+and enabling automatic server-side validation and resolution.
 
-For example, when a `Cluster` references a template:
+There are two kinds of references:
```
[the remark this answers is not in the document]
jhernand: Agree, use "object" instead of "resource".

fulfillment-service/docs/API.md
```diff
@@ -492,30 +492,170 @@ top-level `state` enum.
...
+referencing its parent `VirtualNetwork`, or a `NetworkAttachment` referencing a `Subnet` and
+`SecurityGroup`.
+
+### Naming convention
+
+Each reference type is named `<TargetType>Reference` or `<TargetType>LocalReference` and is
+defined in the target type's `*_type.proto` file. The spec field that uses it is named after the
+relationship (e.g., `template`, `virtual_network`, `role`), not after the reference type itself.
+
+### Wire format (REST/JSON)
```
jhernand: The primary wire format is protobuf. I'd suggest you rename this to "JSON representation", or similar.

jhernand (dismissed):

fulfillment-service/docs/API.md
danielerez: hmm, iinm, it was to limit cross-tenant references to only the shared tenant.
i.e. the interceptor in reference_validator.go hardcodes tenant to "shared" when the flag is set: https://github.com/osac-project/fulfillment-service/blob/049b1d0b398013b5a171265fc6232eccaf2c1425/internal/references/reference_validator.go#L421
but I guess that if we need to reference objects in arbitrary tenants, we could add a string tenant field without breaking the existing shared boolean :/

fulfillment-service/docs/API.md
danielerez: Makes sense, the tables duplicate what's already in the proto files and will go stale anyway...
Will remove.

fulfillment-service/docs/API.md
danielerez: Sure, will rename to match the convention used elsewhere in the doc.

jhernand (approved):

PR #311, opened 2026-08-13
OSAC-3210: reject Cluster creation when NetworkClass has no k8s_manager

ori-amizur:
Validate during Cluster creation that the network attachment's NetworkClass (resolved via Subnet → VirtualNetwork → NetworkClass) has a k8s_manager configured. Without k8s_manager, the hosting cluster has no MetalLB IPAddressPool and CaaS VIP allocation cannot function.

Follows the BareMetalInstance pattern ([OSAC-1464](https://redhat.atlassian.net/browse/OSAC-1464)) which validates fabric_manager on the same chain. Dangling references are skipped — they fail independently during provisioning.

## Summary by CodeRabbit

* **New Features**
  * Cluster creation now validates network attachments and their associated network configuration.
  * Creation is rejected when the selected network class lacks the required Kubernetes manager configuration.
  * Clusters without network attachments continue to be created without this validation.

* **Bug Fixes**
  * Missing referenced network resources are handled safely.
  * Network lookup failures now return an appropriate internal error.

### DIFF
--- a/fulfillment-service/internal/servers/private_clusters_server.go
+++ b/fulfillment-service/internal/servers/private_clusters_server.go
@@ -56,6 +56,9 @@ type PrivateClustersServer struct {
 	catalogItemsDao    *dao.GenericDAO[*privatev1.ClusterCatalogItem]
 	hostTypesDao       *dao.GenericDAO[*privatev1.HostType]
 	clusterVersionsDao *dao.GenericDAO[*privatev1.ClusterVersion]
+	subnetsDao         *dao.GenericDAO[*privatev1.Subnet]
+	virtualNetworksDao *dao.GenericDAO[*privatev1.VirtualNetwork]
+	networkClassesDao  *dao.GenericDAO[*privatev1.NetworkClass]
 	generic            *GenericServer[*privatev1.Cluster]
 }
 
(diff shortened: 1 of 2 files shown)


danmanor:
@ori-amizur This PR should not be merged — it would block CaaS on BM-only deployments (no k8s manager), which is a valid deployment path.

**The original design assumption is outdated.** The design said "Cluster creation is also rejected — without a k8sManager, there is no MetalLB IPAddressPool for VIP allocation." This was true when the design placed MetalLB IPAddressPool creation under the k8s manager's Ansible role. But PR #258 (OSAC-2350, merged) moved IPAddressPool creation into the **Subnet controller** directly — it's gated on `NetworkClass.spec.vip_prefix_length` being set, with no dependency on k8s manager.

The current code path:
1. `resolveVIPCIDR` reads `vip_prefix_length` from NetworkClass (not k8s manager)
2. Stamps VIP CIDR annotation on Subnet CR
3. `ensureMetalLBIPAddressPool` creates the pool — gated on the annotation, not on k8s manager

So in a BM-only deployment (fabricManager only, no k8sManager), if the NetworkClass has `vip_prefix_length` set, CaaS VIP allocation works fine. The k8s manager is only needed for VM workloads (bridging VMs from OVN overlay to fabric), not for CaaS with BM node sets.

**Recommendation:** Close this PR and [OSAC-3210](https://redhat.atlassian.net/browse/OSAC-3210). The design's BM-Only section will be updated to remove the CaaS rejection — only ComputeInstance creation should require k8s manager ([OSAC-1464](https://redhat.atlassian.net/browse/OSAC-1464)).

PR #331, opened 2026-08-14
OSAC-3704: Add regression tests for VirtualNetwork delete protection with typed references

clobrano:
## Summary

Adds regression test coverage to migration 90's test suite for the VirtualNetwork delete-protection trigger fix.

- Verifies `check_virtual_network_not_in_use()` correctly blocks VirtualNetwork deletion when Subnets reference it using typed reference format
- Verifies the same for SecurityGroups and NATGateways
- Follows the pattern from migration 55's tests, adapted for typed references

## Context

Migration 90 fixed a JSON path mismatch in the `check_virtual_network_not_in_use()` trigger that allowed VirtualNetworks to be deleted while children still referenced them ([OSAC-3704](https://redhat.atlassian.net/browse/OSAC-3704)). The fix updated the path from `data->'spec'->>'virtual_network'` to `data->'spec'->'virtual_network'->>'id'` to handle typed reference objects. The migration's test suite verified backfill behavior but lacked tests for the trigger protection itself.

## Test plan

- [x] All three new tests pass locally
- [x] Existing migration 90 tests still pass
- [ ] CI passes

🤖 Generated with Claude Code

## Summary by CodeRabbit

* **Tests**
  * Added integration coverage for preventing the soft deletion of virtual networks still referenced by subnets, security groups, or NAT gateways.
  * Improved validation that deletion errors identify the affected network and referencing resource type, providing clearer feedback when deletion is blocked.

### DIFF
--- a/fulfillment-service/internal/database/migrations/90_update_triggers_for_typed_references_test.go
+++ b/fulfillment-service/internal/database/migrations/90_update_triggers_for_typed_references_test.go
@@ -16,7 +16,9 @@ package migrations
 import (
 	"context"
 	"encoding/json"
+	"errors"
 
+	"github.com/jackc/pgx/v5/pgconn"
 	. "github.com/onsi/ginkgo/v2/dsl/core"
 	. "github.com/onsi/gomega"
 )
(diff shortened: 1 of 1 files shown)


ygalblum (approved):
/approve
/lgtm

PR #337, opened 2026-08-14
OSAC-2971: Wire tenant scoped user authentication for Secret Management apis

DakCrowder:
Adds the ability for an individual tenant-user to make use of the Secrets api.  Uses token forwarding to vault under the hood to get tenant scoped vault tokens that then itneract with the kv store.

Additionally, a few adjustments were made to the way certs are passed in the helm charts for vault to be better aligned with how we do things for other cert bundles atm.

## Summary by CodeRabbit

* **New Features**
  * Added Vault-compatible secret store configuration for service deployments.
  * Added tenant-aware Vault authentication and secret retrieval.
  * Added support for trusted Vault CA bundles supplied through ConfigMaps.
  * Added installer settings for Vault endpoints, namespaces, mounts, authentication, credentials, and TLS configuration.
* **Bug Fixes**
  * Improved tenant isolation for secret access, including cross-tenant protection.
  * Improved Vault authentication error handling and validation.


fulfillment-service/internal/vault/vault_user_jwt_token_source.go
```diff
@@ -0,0 +1,128 @@
...
+	token := auth.TokenFromContext(ctx)
+	if token == nil || token.Raw == "" {
+		return "", fmt.Errorf("user JWT token not found in request context")
+	}
+
+	if tenant == "" {
+		return "", fmt.Errorf("tenant is required for user vault authentication")
+	}
+
+	namespace := path.Join(s.parentNamespace, tenant)
```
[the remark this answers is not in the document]
DakCrowder: The path should have already been validated via validatePathComponent - it exists but in a different level than this

fulfillment-service/internal/vault/vault_lifecycle_test.go
```diff
@@ -422,7 +422,7 @@ var _ = Describe("VaultLifecycleClient", func() {
 
 			boundClaims, ok := roleBody["bound_claims"].(map[string]any)
 			Expect(ok).To(BeTrue())
-			orgs, ok := boundClaims["organization"].([]any)
+			orgs, ok := boundClaims["groups"].([]any)
```
CrystalChun, line 425: nit: should the var name be `tenants`?

fulfillment-service/internal/vault/vault_user_jwt_token_source.go
```diff
@@ -0,0 +1,128 @@
...
+	if token == nil || token.Raw == "" {
+		return "", errors.New("user JWT token not found in request context")
+	}
+
+	if tenant == "" {
+		return "", errors.New("tenant is required for user vault authentication")
+	}
+
+	namespace := path.Join(s.parentNamespace, tenant)
+	vaultToken, _, err := loginToVault(
```
CrystalChun, line 106: Were you planning to use the lease duration in a later change?

fulfillment-service/internal/vault/vault_lifecycle.go
```diff
@@ -305,15 +311,16 @@ path "%s/metadata/*" {
...
-		},
-		"user_claim": "sub",
-		"policies":   []string{"tenant-kv-access"},
-	})
+	_, err := client.Logical().WriteWithContext(ctx,
+		fmt.Sprintf("auth/%s/role/%s", TenantAuthMountPath, TenantAuthRole), map[string]any{
+			"role_type":       "jwt",
+			"bound_audiences": []string{c.keycloakAudience},
+			"bound_claims": map[string]any{
+				"groups": []string{tenantName},
```
CrystalChun, line 319: Is the key `"groups"` from keycloak or vault? If it's keycloak, how come the switch to `"groups"`?

fulfillment-service/it/it_public_secrets_test.go
```diff
@@ -0,0 +1,278 @@
...
+			}
+			Expect(found).To(BeTrue(), "created secret not found in list")
+		})
+
+		It("updates the secret data", func(ctx context.Context) {
+			response, err := secretsClient.Update(ctx, publicv1.SecretsUpdateRequest_builder{
+				Object: publicv1.Secret_builder{
+					Id:   secretId,
+					Data: updatedData,
+				}.Build(),
```
CrystalChun, line 112: Mind adding metadata.name to this update since it'll be required after https://github.com/osac-project/osac/pull/205 merges
```
					Metadata: publicv1.Metadata_builder{
						Name: secretName,
					}.Build(),
```

fulfillment-service/internal/vault/vault_user_jwt_token_source.go
DakCrowder: Yes - there is a follow up ticket to implement a more complete caching where this will be used

fulfillment-service/it/it_public_secrets_test.go
DakCrowder: Ah good point - will update callers

CrystalChun: Thank you! I think it should just be this one 👀

fulfillment-service/internal/vault/vault_lifecycle.go
DakCrowder: The "groups" key is from the Keycloak JWT claims.  I was having some issues with vault accessing the organizations object and I think I got fooled looking at a JWT from the it environment which has groups populated with tenant names but digging deeper looks like that is maybe happenstance that populates these groups with tenants in the it tool.  I'm still looking (or at least will monday :) ) but I think the correct approach might be to try and see if i can get vault matching on the organizations object claim w/ some other mapping change.

DakCrowder:
This is being closed as the authentication methods will be changing.  The core of this work will remain in terms of allowing users access to the secrets store, but implementation will be meaningfully different so will be easier to review as a new PR.

fulfillment-service/internal/vault/vault_lifecycle.go
DakCrowder: Per https://redhat-internal.slack.com/archives/C0B6F4DNP3P/p1786972566748449 I'm planning on re-working this so we won't use these claims directly in Vault.  Closed this PR and will re-open a new one with a tweaked implementation.

PR #350, opened 2026-08-16
OSAC-1541: Improve SetLimit semantics in generic DAO

danielerez:
Summary

  - Reject negative limit values with an error instead of silently treating them as zero items
  - When limit is explicitly set to zero, return only the count (skip item-fetching query) - useful
  for clients that only need to know how many items match a filter
  - Distinguish "limit not set" (use default) from "limit set to zero" (count-only) via a limitSet
  field in ListRequest

  Details

  The SetLimit method on ListRequest previously had legacy semantics from before ListRequest existed
  as a type:

  - Negative value: was returning zero items silently → now returns an error
  - Zero (explicit via SetLimit(0)): was using default limit (100) → now runs only the SELECT COUNT(*)
  query and returns the total with an empty items slice
  - Zero (not set, SetLimit never called): was using default limit → unchanged
  - Positive value: unchanged

  A limitSet bool field was added to ListRequest to distinguish between the two zero cases.

  Changes are in generic_dao_list.go (DAO logic) and generic_dao_test.go (updated + new test cases).

  Test plan

  - [x] "Rejects negative limit with an error" — verifies error is returned
  - [x] "Returns only count when limit is explicitly set to zero" — verifies count-only response with
  empty items
  - [x] "Uses default limit when limit is not set" — verifies default behavior is preserved
  - [x] All 10 limit-related DAO tests pass against PostgreSQL

### DIFF
--- a/fulfillment-service/internal/database/dao/generic_dao_list.go
+++ b/fulfillment-service/internal/database/dao/generic_dao_list.go
@@ -27,8 +27,8 @@ type ListRequest[O Object] struct {
 	request[O]
 	filter   string
 	limit    int32
-	limitSet bool
 	offset   int32
+	limitSet bool
 }
 
 // SetFilter sets the CEL expression that defines which objects should be returned.
@@ -69,7 +69,9 @@ func (r *ListRequest[O]) Do(ctx context.Context) (response *ListResponse[O], err
 func (r *ListRequest[O]) do(ctx context.Context) (response *ListResponse[O], err error) {
 	// Reject negative limits early:
 	if r.limit < 0 {
-		err = fmt.Errorf("limit must be a non-negative integer, but it is %d", r.limit)
+		err = &ErrValidation{
+			Reason: fmt.Sprintf("limit must be a non-negative integer, but it is %d", r.limit),
+		}
 		return
 	}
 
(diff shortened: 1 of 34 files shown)


fulfillment-service/internal/database/dao/generic_dao_list.go
```diff
@@ -25,9 +25,10 @@ import (
 // ListRequest represents a request to list objects with pagination and filtering.
 type ListRequest[O Object] struct {
 	request[O]
-	filter string
-	limit  int32
-	offset int32
+	filter   string
+	limit    int32
+	limitSet bool
```
jhernand: If you put this boolean flag here it will introduce a 4 bytes hole in the struct, three of them unused, just padding. I'd suggest you put it at the end of the struct. Alternatively consider using a single integer field as bitmap for this and other potential boolean fields in the future.

danielerez: Good catch, then I'll move limitSet to the end so limit/offset stay packed.
hmm, about a bitmap, maybe worth keeping it a named bool for now since there's just one flag. we can switch if more flags show up later...

CrystalChun (approved):
/lgtm

jhernand (approved):

CrystalChun (approved):
/lgtm

PR #361, opened 2026-08-17
OSAC-3290: install OSAC CSI driver from tenant storage playbooks

wgordon17:
## Summary
- Adds an idempotent, provider-agnostic Helm install of the OSAC CSI driver (`csi-driver` + `csi-backends` charts) via a new `osac.service.csi_driver_install` role, wired into all three hub-targeted tenant storage dispatch points — installed alongside, not instead of, each provider's existing OLM-based CSI operator, since no StorageClass points at the new driver yet
- Wired at the playbook level rather than into `osac.service.storage_provider`'s own dispatcher, to preserve that role's existing zero-live-I/O unit-test guarantee; excludes the CaaS/guest-cluster call site since this role always targets the hub
- AC-4/AC-6 (StorageClass naming/provisioner cutover to the new driver) are explicitly out of scope for this PR — see Scope Decisions in the design doc; a follow-up story will do the cutover and remove the now-redundant OLM operator install

## Summary

- **Deployment:** Added the idempotent `osac.service.csi_driver_install` role for provider-agnostic CSI driver Helm installation on hub clusters.
- **Playbooks:** Wired installation into compute-instance JIT storage, Tenant storage, and Tenant cluster storage paths. `ClusterOrder` and guest-cluster paths remain excluded.
- **Runtime:** Added provider filtering, controller Service validation, endpoint merging, convergence checks, Lease locking, Helm install/upgrade, and failure handling.
- **API surface:** Added role defaults, the required `csi_driver_install_tiers` argument, and playbook override points.
- **CI:** Updated Helm to v3.21.4 with checksum verification. Added local OCI chart registry setup, trust-store handling, and cleanup.
- **Tests:** Added integration coverage for wiring, idempotency, upgrades, provider handling, missing Services, lock contention, failure paths, and cleanup. Test results are not supplied.
- **Documentation:** Documented the new role and its Lease mutex behavior.
- **Database:** No database changes.
- **Authentication:** No authentication model changes. Test setup creates per-run storage credentials.

### Backward compatibility

Existing provider-specific OLM CSI operators remain in place. StorageClass naming, provisioner migration, and operator removal are deferred. Existing override points remain available. The role writes controller endpoints to `vendorControllers` and reads legacy `vendorSockets` values as fallback. LVMS/topolvm support is not included.

## Risk classification

**risk:show** — Based on the supplied evidence, the change affects shared hub-cluster storage deployment but includes provider validation, Service checks, convergence checks, endpoint preservation, Lease locking, and a pinned Helm version. It does not change database schemas, authentication behavior, StorageClass names, provisioners, or existing OLM operators. Review severity counts and test results are unavailable.

The change is close to **risk:ask** because it changes CSI infrastructure deployment. It does not qualify based on the supplied evidence because migration and operator-removal work is deferred, and no persistent data changes are introduced. The repository’s formal labeling criteria were not supplied.


zszabo-rh:
this builds on the AAP-owned install model while [OSAC-4252](https://redhat.atlassian.net/browse/OSAC-4252) is deciding csi-backends/driver ownership, so worth keeping aligned with that outcome
I know you were in the discussion, so just leaving this here as a minor reminder :)

zszabo-rh:
consider including LVMS vendor

It is somewhat different from vast/pure/trident: the role currently requires each vendor to have a `<vendor>-csi-controller Service` and fails fast without one. LVMS/topolvm is node-local, so no controller Service or vendorSocket for the controller -> all we really need for LVMS is an entry in the OSAC meta-driver's `node.vendorSockets` config pointing at topolvm-node's existing socket (plan is we'd bypass topolvm controller in CreateVolume). 

To make LVMS fit you could add a node-local vendor class in this role now (node-socket-only, plus skip the controller-Service fail-fast).
..or just leave it out of scope here and I'd add the node-local wiring under OSAC-3702. I'm already specifying the topolvm-node socket target for that feature, so I'm totally fine with owning it.

osac-aap/collections/ansible_collections/osac/service/roles/csi_driver_install/tasks/main.yaml
```diff
@@ -0,0 +1,378 @@
...
+    # be a no-op. Errs toward false on any mismatch or missing data, so an
+    # uncertain pre-check simply falls back to running the install block.
+    - name: Determine whether this dispatch is already fully converged
+      ansible.builtin.set_fact:
+        _csi_driver_install_converged: >-
+          {% set _ns = namespace(ok=false) -%}
+          {% if _csi_driver_install_precheck_status is not none -%}
+          {% set _s = _csi_driver_install_precheck_status -%}
+          {% set _ctrl = {} -%}
+          {% for _pair in (_s['values'].controller.vendorSockets | default('', true) | trim).split(',') if _pair | trim | length > 0 -%}
```
akshaynadkarni, line 132: Accessing `_s['values'].controller.vendorSockets` and `_s['values'].node.vendorSockets` directly will raise a `jinja2.exceptions.UndefinedError` and fail the playbook if the previously deployed Helm release does not explicitly define the `.controller` or `.node` keys (or defines them as null). In Jinja, a trailing `default` filter cannot rescue a nested attribute access if an intermediate key in the chain is undefined.

To make the pre-check resilient against customized or corrupted Helm release states, guard the intermediate keys with the `default` filter:

```suggestion
          {% set _ctrl = {} -%}
          {% for _pair in ((_s['values'].controller | default({}, true)).vendorSockets | default('', true) | trim).split(',') if _pair | trim | length > 0 -%}
          {% set _kv = _pair.split('=', 1) -%}
          {% set _ = _ctrl.update({_kv[0]: _kv[1]}) -%}
          {% endfor -%}
          {% set _node = {} -%}
          {% for _pair in ((_s['values'].node | default({}, true)).vendorSockets | default('', true) | trim).split(',') if _pair | trim | length > 0 -%}
```

osac-aap/collections/ansible_collections/osac/service/roles/csi_driver_install/tasks/main.yaml
```diff
@@ -0,0 +1,378 @@
...
+      ansible.builtin.set_fact:
+        _csi_driver_install_prior_driver_status: "{{ _csi_driver_install_release_info.status | default(none, true) }}"
+
+    # Bracket notation (['values']) is required here, not dot notation: the
+    # release dict has a real key literally named "values", which collides
+    # with Python dict's own .values() method — attribute access would
+    # silently return the bound method instead of the chart values.
+    - name: Extract previously-deployed controller/node vendorSockets strings (empty string if unset)
+      ansible.builtin.set_fact:
+        _csi_driver_install_prior_controller_sockets_raw: >-
```
akshaynadkarni, line 257: Similar to the precheck task above, accessing nested attributes on `_csi_driver_install_prior_driver_status['values'].controller.vendorSockets` directly will raise an `UndefinedError` and fail the playbook if the previous release has an undefined or null `.controller` (or `.node`) key. 

To prevent onboarding failures on existing releases, guard the intermediate keys:

```suggestion
      ansible.builtin.set_fact:
        _csi_driver_install_prior_controller_sockets_raw: >-
          {{ ((_csi_driver_install_prior_driver_status['values'].controller | default({}, true)).vendorSockets | default('', true) | trim)
             if (_csi_driver_install_prior_driver_status is not none) else '' }}
        _csi_driver_install_prior_node_sockets_raw: >-
          {{ ((_csi_driver_install_prior_driver_status['values'].node | default({}, true)).vendorSockets | default('', true) | trim)
             if (_csi_driver_install_prior_driver_status is not none) else '' }}
```

akshaynadkarni (commented):
@wgordon17 

The concurrency controls (`osac.service.lease` locks, quick-convergence pre-checks) and the local TLS OCI registry (`localtest.me`) integration test setup you’ve introduced here seem very useful.

I want to flag a dual-ownership conflict regarding the `csi-backends` Helm release between the umbrella installer and AAP:

### The Problem: Dual Ownership
Currently, this PR defaults `storage_provider_csi_backends_enabled: true` in the role defaults. This means tenant onboarding will attempt to install/upgrade the `osac-csi-backends` Helm release on the hub. 

However, on standard hub clusters, the base `osac-installer` (umbrella chart) also deploys `csi-backends` as a sub-chart when `csiDriver.enabled` is `true`. When both installers attempt to manage the same cluster-wide resources (such as `vast-csi-controller`) under different release names, Helm throws ownership errors (`invalid ownership metadata`) and fails.

This collision was observed during testing under ticket **[OSAC-4252](https://redhat.atlassian.net/browse/OSAC-4252)** (and worked around in PR #551 by defaulting the AAP flag to `false`).

Longer term, **[OSAC-4525](https://redhat.atlassian.net/browse/OSAC-4525)** (Dynamic Vendor Controller Lifecycle) will remove `csi-backends` from the installer entirely and shift its deployment lifecycle to the `osac-operator` (tied to `StorageBackend` CR registration). At that point, AAP will no longer manage `csi-backends` at all.

### Questions & Next Steps
* Were you able to test this on a cluster provisioned using `osac-installer`? If so, did you run into Helm ownership collisions on the cluster-wide resources (e.g., `vast-csi-controller` ClusterRoles/DaemonSets)?
* To align with the current workaround in **OSAC-4252** and prevent onboarding failures in our standard CI/CD and lab environments, perhaps we should flip `storage_provider_csi_backends_enabled` to `false` by default in `storage_provider/defaults/main.yaml` while leaving it enabled inside your integration tests to verify the rollback/installation logic?

Let me know your thoughts.

osac-aap/collections/ansible_collections/osac/service/roles/csi_driver_install/tasks/main.yaml
```diff
@@ -0,0 +1,378 @@
...
+        reset_values: false
+        values: "{{ _csi_driver_install_driver_values }}"
+      register: _csi_driver_install_driver_result
+
+  rescue:
+    - name: Report OSAC CSI driver install failure
+      ansible.builtin.fail:
+        msg: >-
+          OSAC CSI driver install failed. Failed task: {{ ansible_failed_task.name | default('unknown') }}.
+          Error: {{ ansible_failed_result.msg | default('see logs for details') }}.
```
zszabo-rh: Caderabbit also flagged this that logging the full `ansible_failed_result.msg` is not the best practice (could theoretically include internal hostnames or auth details from Helm module failures).  Not critical but I agree this could be revised, e.g. logging only ansible_failed_task.name without the full error message, or gating the detail behind a debug verbosity check.

zszabo-rh (commented):
Consider fixing below finding (originally from coderabbit), otherwise LGTM!

osac-aap/collections/ansible_collections/osac/service/roles/csi_driver_install/tasks/main.yaml
```diff
@@ -0,0 +1,379 @@
...
+          {% set _ns = namespace(ok=false) -%}
+          {% if _csi_driver_install_precheck_status is not none -%}
+          {% set _s = _csi_driver_install_precheck_status -%}
+          {% set _ctrl = {} -%}
+          {% for _pair in (_s['values'].controller.vendorControllers | default(_s['values'].controller.vendorSockets | default('', true), true) | trim).split(',') if _pair | trim | length > 0 -%}
+          {% set _kv = _pair.split('=', 1) -%}
+          {% set _ = _ctrl.update({_kv[0]: _kv[1]}) -%}
+          {% endfor -%}
+          {% set _node = {} -%}
+          {% for _pair in (_s['values'].node.vendorSockets | default('', true) | trim).split(',') if _pair | trim | length > 0 -%}
```
akshaynadkarni: As per my AI review too, there is still a risk of UndefinedError here. The default filter on vendorControllers cannot protect the direct _s['values'].controller access. The fallback expression is also evaluated before default. If an existing Helm release has missing or null values, controller, or node objects, the convergence check can fail instead of falling back to the install path.

The suggestion is to normalize the parent maps before parsing:
(You will have to indent the content, if you decide to accept this)

```suggestion
{% set _values = _s['values'] | default({}, true) -%}
{% set _controller = _values['controller'] | default({}, true) -%}
{% set _node_values = _values['node'] | default({}, true) -%}
{% set _ctrl = {} -%}
{% set _controller_raw = (_controller['vendorControllers'] | default('', true))
   or (_controller['vendorSockets'] | default('', true)) -%}
{% for _pair in (_controller_raw | trim).split(',') if _pair | trim | length > 0 -%}
{% set _kv = _pair.split('=', 1) -%}
{% set _ = _ctrl.update({_kv[0]: _kv[1]}) -%}
{% endfor -%}
{% set _node = {} -%}
{% for _pair in (_node_values['vendorSockets'] | default('', true) | trim).split(',') if _pair | trim | length > 0 -%}
```

osac-aap/collections/ansible_collections/osac/service/roles/csi_driver_install/tasks/main.yaml
```diff
@@ -0,0 +1,379 @@
...
+    # silently return the bound method instead of the chart values.
+    - name: Extract previously-deployed controller/node vendor endpoints strings (empty string if unset)
+      ansible.builtin.set_fact:
+        _csi_driver_install_prior_controller_sockets_raw: >-
+          {{ ((_csi_driver_install_prior_driver_status['values'].controller.vendorControllers | default('', true))
+              or (_csi_driver_install_prior_driver_status['values'].controller.vendorSockets | default('', true)) | trim)
+             if (_csi_driver_install_prior_driver_status is not none) else '' }}
+        _csi_driver_install_prior_node_sockets_raw: >-
+          {{ (_csi_driver_install_prior_driver_status['values'].node.vendorSockets | default('', true) | trim)
+             if (_csi_driver_install_prior_driver_status is not none) else '' }}
```
akshaynadkarni: As per my AI review tool, the same risk remains in the release-value extractor. The current expression only guards whether the release status exists. It still directly dereferences values.controller and values.node, so a customized or incomplete existing release can fail with UndefinedError.

The suggestion is to normalize values, controller, and node before extracting the endpoint strings, then use the normalized maps:

```
- name: Normalize previously deployed csi-driver values
  ansible.builtin.set_fact:
    _csi_driver_install_prior_values: >-
      {{
        _csi_driver_install_prior_driver_status['values'] | default({}, true)
        if _csi_driver_install_prior_driver_status is not none
        else {}
      }}

- name: Normalize previously deployed controller and node values
  ansible.builtin.set_fact:
    _csi_driver_install_prior_controller: >-
      {{ _csi_driver_install_prior_values['controller'] | default({}, true) }}
    _csi_driver_install_prior_node: >-
      {{ _csi_driver_install_prior_values['node'] | default({}, true) }}
```

Then extract safely:

```
_csi_driver_install_prior_controller_sockets_raw: >-
  {{
    (
      (_csi_driver_install_prior_controller['vendorControllers'] | default('', true))
      or (_csi_driver_install_prior_controller['vendorSockets'] | default('', true))
    ) | trim
  }}
_csi_driver_install_prior_node_sockets_raw: >-
  {{ _csi_driver_install_prior_node['vendorSockets'] | default('', true) | trim }}
```

These jinja suggestions are more runtime concerns. If you feel, the scenario can't occur, it's ok to not address them. At the same time, if you think this is good code hardening, please update the code. Thanks.

osac-aap/collections/ansible_collections/osac/service/roles/csi_driver_install/tasks/main.yaml
```diff
@@ -0,0 +1,379 @@
...
+    - name: Install/upgrade csi-driver Helm release
+      kubernetes.core.helm:
+        name: csi-driver
+        chart_ref: "{{ csi_driver_install_chart_registry }}/csi-driver"
+        chart_version: "{{ csi_driver_install_chart_version }}"
+        release_namespace: "{{ csi_driver_install_namespace }}"
+        state: present
+        reuse_values: true
+        reset_values: false
+        values: "{{ _csi_driver_install_driver_values }}"
```
akshaynadkarni, line 383: I understand the intent of the Helm options: repeated AAP dispatches should be idempotent, and previously configured vendor values should be preserved across dispatches.

My concern here is separate from that value-merging behavior. I think we need an enable/disable flag for this AAP Helm installation, similar to storage_provider_csi_backends_enabled, with the default set to false.

The umbrella chart declares both the csi-driver and csi-backends dependencies (https://github.com/osac-project/osac/blob/bdb3a419656a906b7c23bd2c54be2e3e1d17376f/osac-installer/charts/osac/Chart.yaml#L45-L54), and the vmaas-ci profile enables csiDriver (https://github.com/osac-project/osac/blob/bdb3a419656a906b7c23bd2c54be2e3e1d17376f/osac-installer/values/vmaas-ci/instance.yaml#L15-L18).

My understanding is that the umbrella chart should own the csi-driver deployment. AAP should not install a second csi-driver Helm release from the tenant storage playbooks, since this can cause Helm ownership conflicts.

Could we gate these AAP role invocations with a csi_driver_install_enabled flag and leave it disabled by default? This would make the ownership model explicit, similar to the csi-backends change.

osac-aap/tests/integration/targets/csi_driver_install/tasks/main.yml
```diff
@@ -0,0 +1,731 @@
...
+        name: csi-driver-install-lock
+        namespace: "{{ lookup('env', 'POD_NAMESPACE') }}"
+      register: _lease_after_noop_dispatch
+
+    - name: Assert the install lock was never touched (block was skipped, not just a no-op)
+      ansible.builtin.assert:
+        that:
+          - >-
+            (_lease_before_noop_dispatch.resources[0].metadata.resourceVersion | default(''))
+            == (_lease_after_noop_dispatch.resources[0].metadata.resourceVersion | default(''))
```
akshaynadkarni: (As per AI)
One concern with this test is that the Lease comparison does not prove the install block was skipped.

A normal install acquires the Lease and then deletes it in the always block. Therefore, the Lease can have the same state before and after the test whether the role was completely skipped or acquired and released the lock.

Could we compare the Helm release revision before and after instead? The revision would only change if the role actually performed a Helm install or upgrade.

osac-aap/tests/integration/setup_test_env.sh
```diff
@@ -163,20 +168,114 @@ CSIEOF
...
+  # falling back to whichever of docker/podman is actually installed.
+  if [ -n "${KIND_EXPERIMENTAL_PROVIDER:-}" ]; then
+    CSI_DRIVER_TEST_CONTAINER_TOOL="${KIND_EXPERIMENTAL_PROVIDER}"
+  elif command -v docker > /dev/null 2>&1; then
+    CSI_DRIVER_TEST_CONTAINER_TOOL="docker"
+  else
+    CSI_DRIVER_TEST_CONTAINER_TOOL="podman"
+  fi
+
+  # This whole section is best-effort: it must never abort the rest of setup. The
```
akshaynadkarni, line 214: (ai) This is a test-harness reliability issue rather than a production issue. This section is documented as best effort, but the script runs with set -e and the commands inside this block are not guarded.

If certificate generation, sudo, container startup, or registry readiness fails, setup exits and prevents unrelated integration tests from running. The script also proceeds to package and push the charts even when the readiness loop never succeeds.

Could we track registry readiness explicitly, only package and push when the registry is ready, and otherwise leave CSI_DRIVER_INSTALL_TEST_REGISTRY unset so the role-level test falls back to the published chart?

osac-aap/.claude/rules/playbook-patterns.md
```diff
@@ -108,10 +108,11 @@ capabilities:  # informational only here -- see the `capabilities` field note be
 |------|---------|-------|
 | `osac.service.common` | Shared utilities (kubeconfig, credentials) | `tasks_from: get_remote_cluster_kubeconfig` |
 | `osac.service.finalizer` | Finalizer management for CRs | `tasks_from: add_finalizer` |
-| `osac.service.lease` | Bare-metal lease management | Used by cluster/compute workflows |
+| `osac.service.lease` | Generic Kubernetes Lease-based mutex | Used by cluster/compute/storage workflows |
 | `osac.service.wait_for` | Polling utilities | Wait for pods, deployments, CRs |
 | `osac.service.tenant_storage_class` | StorageClass discovery | Find tenant-specific storage |
 | `osac.service.publish_templates` | Template registration | Publishes ComputeClass-family resources from `meta/osac.yaml` |
+| `osac.service.csi_driver_install` | Idempotent OSAC CSI driver Helm install | Called directly from 3 hub-targeting storage dispatch playbooks |
```
rgolangh: nit: not sure I would count the number of playbooks this referece. This would easily drift.

osac-aap/collections/ansible_collections/osac/service/roles/csi_driver_install/defaults/main.yaml
```diff
@@ -0,0 +1,31 @@
...
+csi_driver_install_backends_namespace: "osac-csi-backends"
+
+# Vendors the csi-driver chart knows how to enable. Dispatched providers
+# outside this list are logged and skipped rather than failing the whole
+# install, since storage_provider's own provider validation is separate
+# from what this chart happens to support today.
+csi_driver_install_supported_vendors:
+  - vast
+  - pure
+  - trident
```
rgolangh: pure and trident are still unsupported

osac-aap/playbook_osac_create_compute_instance.yml
```diff
@@ -72,6 +79,26 @@
...
+    # once-per-dispatch OSAC CSI driver install ahead of it.
+    #
+    # Unlike the Tenant call site, no ordering guarantee exists here -- this
+    # JIT path has no osac-operator Stage-gating invariant protecting it.
+    # csi_driver_install's own pre-block fail-fast (checking each dispatched
+    # vendor's controller Service exists) is the actual protection, not
+    # assumed safety. In practice this path almost always dispatches zero
+    # CSI-driver-supported vendors today (typically 'lvms'), so the fail-fast
+    # is expected to be rarely exercised but is a real, live code path.
+    - name: Install OSAC CSI driver for JIT storage (once per dispatch, regardless of provider count)
```
rgolangh: even though this is idempotent, I don't think it makes sense to invoke an installation when a VM is created. I understand that the concern is 
that there can be a storage tier which is missing, and therefor we have install the csi backend, if missing, and create a storage class. 
But this brings me back to the thought that the VM should be scheduled on a cluster only if the clustger is reporting those tiers. Meaning - this is a scheduing decision.
My concern is that doing insallation as part of VM lifecycle simple wonm'e scale. If we want to start 50 VMs at once then this means 50 installation competing on the lease, 49 of them waiting.

osac-aap/playbook_osac_create_tenant_cluster_storage.yml
```diff
@@ -80,6 +87,32 @@
...
+    # this file cannot enforce or verify: osac-operator's storage_controller.go
+    # handleBackendReadiness blocks Stage 2 from running until Stage 1
+    # (playbook_osac_create_tenant_storage_backend.yml's setup dispatch) has
+    # gone green for this tenant. A future change to that controller's
+    # Stage-gating logic could silently break this precondition. That
+    # cross-repo Stage-gating contract is verified in osac-operator's
+    # internal/controller/storage_controller_test.go ("Stage 1: Backend
+    # provisioning" and "Stage 2: Cluster storage provisioning" contexts),
+    # which drive handleBackendReadiness against a real envtest control plane.
+    - name: Install OSAC CSI driver on the hub (once per dispatch, regardless of provider count)
```
rgolangh, line 119: why is a tenant clusters excluded here?

osac-aap/tests/integration/targets/csi_driver_install/tasks/main.yml
```diff
@@ -0,0 +1,731 @@
...
+# plane at all -- it only reads live Service objects' ports (fixtures below) to build
+# DNS-based vendorSockets strings for the csi-driver chart. Vendor controller Services
+# are owned and deployed by the umbrella installer under OSAC-4252.
+- name: "CSI Driver Install - Setup (clean slate)"
+  hosts: localhost
+  gather_facts: false
+
+  vars:
+    csi_driver_install_test_namespace: "osac-csi"
+    csi_driver_install_test_backends_namespace: "osac-csi-backends"
```
rgolangh, line 31: small detail here is that the csi charts use the namespace from values (with those default already there) and not usding the release namespace. In other words setting the anemsapce here will not override it. {{ .Relese.namespace }} vs {{ .Values.namespace }}.

rgolangh (commented):
Overall good but still some questions around installation triggered by VM creation.

osac-aap/collections/ansible_collections/osac/service/roles/csi_driver_install/meta/argument_specs.yaml
```diff
@@ -16,6 +16,15 @@ argument_specs:
       keep storage_provider/tests/test.yml's zero-live-I/O validation
       guarantee intact.
     options:
+      csi_driver_install_enabled:
+        type: bool
+        required: false
+        default: false
```
rgolangh, line 21: why default: false?

osac-aap/collections/ansible_collections/osac/service/roles/csi_driver_install/tasks/main.yaml
```diff
@@ -335,6 +372,7 @@
           namespace: "{{ csi_driver_install_namespace }}"
           controller:
             vendorControllers: "{{ _csi_driver_install_controller_vendor_sockets }}"
+            vendorSockets: ""
```
rgolangh: vendorSockets is irrelevant for the controller. this should be dropped.

akshaynadkarni (approved):
Overall, code changes seem fine. 

Thanks for addressing the previous Jinja issues and Roy's comments. The default ownership gates align with OSAC-4252.

Approving. The remaining AAP serialization, VM hot-path, and registry fallback improvements can be handled later under OSAC-4525, which is where i've captured these enhancements.

osac-aap/playbook_osac_create_tenant_cluster_storage.yml
wgordon17: decided that "umbrella installer" owns CSI installation on hub

PR #370, opened 2026-08-18
NO-ISSUE: fix: preserve comma-separated --additional-disk values

clobrano:
## Problem

`osac create computeinstance --additional-disk "size=50,storage-tier=<tier>"` fails with:

```
Error: --additional-disk "storage-tier=<tier>" must include size=<value>
```

The flag is registered with Cobra's `StringSliceVar`, which splits values on commas at flag-parse time. So `size=50,storage-tier=<tier>` is split into two array elements (`size=50` and `storage-tier=<tier>`), and the parser (`parseAdditionalDisks`) processes each as a standalone disk spec — the second one has no `size=` fragment and fails validation.

This blocks the storage-tier E2E tests in osac-test-infra (11 failures), which pass the documented `size=<GiB>,storage-tier=<name>` form.

## Fix

Change `--additional-disk` from `StringSliceVar` to `StringArrayVar`, which preserves each `--additional-disk` occurrence as a single string without comma-splitting. This matches the documented flag format and the existing `--network-attachment` flag (also `StringArrayVar` because its value contains commas).

## Tests

Added specs that drive the real Cobra flag → `parseAdditionalDisks` path (reading the flag value type-agnostically via pflag's `SliceValue`). Confirmed failing before the change, passing after.

## Coordination

Unblocks the storage-tier E2E tests in osac-project/osac-test-infra#367 — that workflow needs to build the `osac` CLI from this branch, or wait for this to merge and re-run CI.

### DIFF
--- a/fulfillment-service/internal/cmd/cli/create/computeinstance/create_compute_instance_cmd.go
+++ b/fulfillment-service/internal/cmd/cli/create/computeinstance/create_compute_instance_cmd.go
@@ -128,10 +128,10 @@ func Cmd() *cobra.Command {
 		"",
 		bootDiskStorageTierFlagHelp,
 	)
-	flags.StringSliceVar(
+	flags.StringArrayVar(
 		&runner.args.additionalDisks,
 		"additional-disk",
-		[]string{},
+		nil,
 		additionalDiskFlagHelp,
 	)
 	flags.StringVar(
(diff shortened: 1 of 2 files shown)


clobrano:
Addressed the fullsend review findings (amended into the same commit, now `8b2d678`):

- **default-value convention** — changed the `--additional-disk` default from `[]string{}` to `nil` to match the sibling `StringArrayVar` flags (`--network-attachment`, `--set`).
- **test integrity** — reworded the `rawDisks` helper comment to explain it reads through pflag's `SliceValue` (implemented by both slice types) precisely so the assertions catch how the two types differ during `Parse()`.

ygalblum (approved):
/approve
/lgtm

PR #379, opened 2026-08-19
OSAC-4073: enforce one NetworkClass per deployment

SiddarthR56:
NetworkClass is a provider-level singleton per the unified networking design ([OSAC-1433](https://redhat.atlassian.net/browse/OSAC-1433)): tenants never interact with it directly, and only one should ever exist. Reject Create with FailedPrecondition when a NetworkClass already exists, both as a fast application-level check and via a unique partial index (migration 101) that closes the TOCTOU race between two concurrent Create calls.

Refactor test fixtures across internal/servers and one integration test that previously created multiple NetworkClass instances per test to seed via DAO/memoization or explicitly drop the new index where multiple coexisting NetworkClasses are intentionally exercised.

## Summary by CodeRabbit

* **New Features**
  * Enforced a single active NetworkClass per deployment, with replacement allowed after deletion.
  * Added clear API errors for singleton conflicts and invalid filters.
  * Added configurable NetworkClass selection during template publishing, including explicit strategy overrides and default selection.
* **Documentation**
  * Updated API and workflow guidance for NetworkClass replacement and error handling.
* **Bug Fixes**
  * Improved cleanup behavior in integration tests and prevented ambiguous NetworkClass publishing.


danmanor:
The implementation is solid (server-side check + DB unique partial index as the TOCTOU backstop, mirroring the network_classes_single_default pattern), and it matches design decision #6 in principle.

**But it breaks the vmaas full-install e2e at install time** — not in the tests, earlier:

```
Error: failed post-install: 1 error occurred:
  * job osac-publish-templates failed: BackoffLimitExceeded
make: *** [Makefile:90: install-osac] Error 1
```

Root cause: config-as-code's `publish_templates` creates **one NetworkClass per network template role**, and there are 5 (`cudn_net`, `netris`, `openstack`, `network_policy`, `metallb_l2`). The first Create succeeds; the singleton enforcement rejects the rest, the AAP job exhausts its backoff, and the Helm install fails.

This is a deeper tension than a test fix — three things currently disagree about how many NetworkClasses exist:
- Decision #6: "one NetworkClass per deployment"
- config-as-code: publishes a catalog — one NetworkClass per available backend
- the existing `is_default` swap logic: only meaningful if multiple can coexist

To land this, the publishing model has to be reconciled first. Options:
1. Change `publish_templates` to publish only the **one** selected deployment NetworkClass (the true "one per deployment" model). This pairs naturally with the Enclave fabric-manager-selection work ([OSAC-2434](https://redhat.atlassian.net/browse/OSAC-2434)), where the provider picks the backend — config-as-code would publish only that one.
2. Or defer this PR until config-as-code is updated.

Note that once we're truly one-per-deployment, the `is_default` mechanism becomes redundant and could be removed in the same effort.

Suggest holding this PR until the config-as-code publishing change lands, otherwise every deployment that vendors more than one network role fails to install.

SiddarthR56:
**Audit of the gRPC error-code remap** (`GenericServer.Create` singleton UniqueViolation → `FailedPrecondition`): no consumer updates are required.

- **osac-operator** never Creates a NetworkClass. `NetworkClassAdapter` only `Get`s; production `AlreadyExists` usage is Kubernetes `apierrors.IsAlreadyExists` (namespaces and similar), not the gRPC NetworkClass API.
- **osac-test-infra** Lists an already-published NetworkClass. Its `AlreadyExists` matching is for Tenant / ClusterVersion create idempotency, not NetworkClass.
- **osac-aap `publish_templates`** Lists then PATCH/POST; it does not branch on gRPC `AlreadyExists`.
- In-repo tests already assert `FailedPrecondition` for a second NetworkClass Create.

This is also not a silent remap of an existing second-Create failure. Before this PR a second NetworkClass with a different name succeeded; `AlreadyExists`→`FailedPrecondition` remapping applied only to default-swap races on `network_classes_single_default`. The new code is the singleton contract (`FailedPrecondition` when another non-deleted NetworkClass exists; ordinary name uniqueness stays `AlreadyExists`).

Documented that distinction in `fulfillment-service/docs/API.md` so future consumers match on status codes rather than assuming all uniqueness is `AlreadyExists`.

danmanor:
## Feedback: NetworkClass publishing from templates should be removed

The AAP changes in this PR (the `select_network_class` filter, `OSAC_NETWORK_CLASS_IMPLEMENTATION_STRATEGY` env var, and the selection logic in `network_classes.yaml`) add complexity for a flow that should be removed entirely. NetworkClass should not be published from template discovery — it's a provider-level singleton configured at deployment time, not something discovered from Ansible collection roles.

### NetworkClass creation model

NetworkClass should be created (via private API or installer bootstrap) with explicit default k8s manager and fabric manager fields, not derived from template metadata. The deployment decides what managers it uses:

- **K8s-only deployment** (VMaaS with CUDN/MetalLB/NetworkPolicy): `fabric_manager: ""` (none), `k8s_manager: "cudn_net"` or empty — relies on `implementation_strategy` via the legacy fallback path
- **Physical fabric deployment** (Netris): `fabric_manager: "netris"`, dispatched via the two-manager model
- **Mixed**: both managers set, dispatcher handles multi-target dispatch (e.g., Subnet → fabric + k8s)

### Recommendation

1. **Remove** the AAP template publishing changes from this PR (the `select_network_class` filter, `OSAC_NETWORK_CLASS_IMPLEMENTATION_STRATEGY`, and the selection logic in `network_classes.yaml`)
2. **Keep** the fulfillment-service singleton enforcement (DB index + app-level check) — that's correct and valuable
3. Separately, remove NetworkClass publishing from `publish_templates` entirely (the `network_classes.yaml` tasks, `find_network_class_roles` filter, and `NetworkClassMetadata` model)
4. NC-VAL-05 (`"at least one of 'fabric_manager' or 'k8s_manager' is required"`) should be relaxed to allow both empty when `implementation_strategy` is set — the operator's `dispatcher_helpers.go` already handles this case correctly via the legacy fallback path (`ErrNoManagerConfigured` → falls back to `legacyStrategy`)

danmanor:
## Migration Plan Update — Changes Required

This PR is **T12** in the [OSAC-1441](https://redhat.atlassian.net/browse/OSAC-1441) networking migration plan. The fulfillment-service singleton enforcement (DB index + app-level check) is correct and should be kept. However, the AAP changes need to be removed.

### What to keep
- Migration 102: `network_classes_singleton` unique partial index — race-free enforcement at DB level
- `checkSingleNetworkClass()` app-level check — fast clear error for the common case
- `isSingletonConstraintViolation()` + FailedPrecondition error mapping in GenericServer
- All fulfillment-service test updates
- API.md documentation updates

### What to remove
- `select_network_class` filter plugin (67 lines) + its unit tests (67 lines)
- `OSAC_NETWORK_CLASS_IMPLEMENTATION_STRATEGY` env var and its plumbing through playbooks/defaults/argument_specs
- Selection logic changes in `network_classes.yaml` tasks
- Test cases for explicit/ambiguous NC selection
- AGENTS.md / CLAUDE.md changes about NC replacement

**Reason**: NetworkClass will no longer be published from template discovery at all ([OSAC-4270](https://redhat.atlassian.net/browse/OSAC-4270)). NC creation moves to the installer Helm chart ([OSAC-4271](https://redhat.atlassian.net/browse/OSAC-4271)). Adding selection complexity for a flow that's being removed is unnecessary work.

**Dependency**: This PR should merge after [OSAC-4270](https://redhat.atlassian.net/browse/OSAC-4270) (remove NC from template publishing), which itself is blocked by [OSAC-4271](https://redhat.atlassian.net/browse/OSAC-4271) (installer NC creation). The DB-side singleton enforcement can be split out and merged earlier if desired.

See: `docs/superpowers/specs/2026-08-22-networkclass-migration-design.md`

danmanor:
## Migration Plan Update (updated)

This PR is **[OSAC-4073](https://redhat.atlassian.net/browse/OSAC-4073)** in the [OSAC-1441](https://redhat.atlassian.net/browse/OSAC-1441) networking migration plan. The fulfillment-service singleton enforcement (DB index + app-level check) is correct and should be kept. However, the AAP changes need to be removed.

### What to keep
- Migration 102: `network_classes_singleton` unique partial index
- `checkSingleNetworkClass()` app-level check
- `isSingletonConstraintViolation()` + FailedPrecondition error mapping
- All fulfillment-service test updates
- API.md documentation updates

### What to remove
- `select_network_class` filter plugin + unit tests
- `OSAC_NETWORK_CLASS_IMPLEMENTATION_STRATEGY` env var and plumbing
- Selection logic changes in `network_classes.yaml` tasks
- Test cases for explicit/ambiguous NC selection
- AGENTS.md / CLAUDE.md changes about NC replacement

**Reason**: NetworkClass will no longer be published from template discovery ([OSAC-4270](https://redhat.atlassian.net/browse/OSAC-4270)). NC creation moves to the installer Helm chart ([OSAC-4271](https://redhat.atlassian.net/browse/OSAC-4271)). Adding selection complexity for a flow that's being removed is unnecessary.

**Dependency**: Merge after [OSAC-4270](https://redhat.atlassian.net/browse/OSAC-4270) (remove NC publishing), which is blocked by [OSAC-4271](https://redhat.atlassian.net/browse/OSAC-4271) (installer NC creation). The DB singleton enforcement can be split out and merged earlier if desired.

See: `docs/superpowers/specs/2026-08-22-networkclass-migration-design.md`

ori-amizur:
Replaced by https://github.com/osac-project/osac/pull/475

PR #385, opened 2026-08-19
OSAC-2978: Add pull_secret_secret reference field to Cluster and ClusterTemplate

danielerez:
##  Summary

  - Define SecretLocalReference proto type (public + private) following the established
    *LocalReference pattern
  - Add pull_secret_secret field to ClusterSpec and ClusterTemplateSpecDefaults (both public and
    private protos)
  - Add server-side validation: mutual exclusion with inline pull_secret, reference existence check
    via lookupSecret(), and resolved id+name normalization
  - Validation covers all Create paths (template, catalog item) and the Update path
  - Update ApplyClusterSpecDefaults to propagate pull_secret_secret from template defaults with
    correct precedence: user inline > user secret ref > default secret ref > default inline
  - Add 13 unit tests covering reference by id/name, mutual exclusion, nonexistent secret rejection,
    template default propagation, and user overrides

##  Context

  First vertical slice of the secret references migration ([OSAC-2953](https://redhat.atlassian.net/browse/OSAC-2953)). Establishes the pattern for replacing inline credentials with SecretLocalReference fields. Subsequent tickets (Hub kubeconfig, IdP client_secret, StorageBackend password, Tenant break_glass_credentials) follow the same approach.

  No DB migration needed. Resources use GenericDAO with JSON data column, so proto field additions are picked up automatically.

##  Test plan

  - [x] Create cluster with pull_secret_secret by id — resolves and stores id + name
  - [x] Create cluster with pull_secret_secret by name — resolves and stores id + name
  - [x] Create with both pull_secret and pull_secret_secret — returns INVALID_ARGUMENT
  - [x] Create with nonexistent secret reference — returns INVALID_ARGUMENT
  - [x] Create with inline pull_secret only — still works unchanged
  - [x] Update with both fields set — returns INVALID_ARGUMENT
  - [x] Template default pull_secret_secret propagates to cluster
  - [x] User inline pull_secret overrides template pull_secret_secret default
  - [x] Defaults unit tests: 5 cases for precedence logic


fulfillment-service/proto/private/osac/private/v1/cluster_template_type.proto
DakCrowder: Are we able to mark the existing fields as deprecated in the .proto?

fulfillment-service/internal/servers/private_clusters_server.go
```diff
@@ -461,6 +479,71 @@ func (s *PrivateClustersServer) lookupTemplate(ctx context.Context,
 	return
 }
 
+func (s *PrivateClustersServer) lookupSecret(ctx context.Context,
+	key string) (result *privatev1.Secret, err error) {
+	if key == "" {
+		return
+	}
+	response, err := s.secretsDao.List().
+		SetFilter(fmt.Sprintf("this.id == %[1]s || this.metadata.name == %[1]s", strconv.Quote(key))).
```
DakCrowder: Instead of defining our own lookup, can we use code in the reference_lookups?  I  was trying to see if we needed to define project scoping here but I think that there is some more generic reference code defined.  Then we just have to worry about the existing field / mutually exclusive cases.

fulfillment-service/internal/servers/private_clusters_server.go
```diff
@@ -461,6 +479,71 @@ func (s *PrivateClustersServer) lookupTemplate(ctx context.Context,
...
+			grpccodes.InvalidArgument,
+			"there are multiple secrets with identifier or name '%s'",
+			key,
+		)
+	}
+	return
+}
+
+func (s *PrivateClustersServer) validatePullSecretMutualExclusion(spec *privatev1.ClusterSpec) error {
+	if spec.HasPullSecret() && spec.GetPullSecretSecret() != nil {
```
DakCrowder, line 485: In the update case, I think we may need to compare against what is in the db rather than what is being passed - I think currently an update mask could be passed with the new field set while the cluster has the old field still set but not specified in the update mask.

fulfillment-service/internal/servers/private_clusters_server.go
danielerez: Good point. the SecretLocalReference ends with "LocalReference" so the interceptor picks it up automatically. probably just needs a RegisterDAOLookup entry in reference_lookups.go. Will try to refactor.

fulfillment-service/internal/servers/private_clusters_server.go
danielerez: good catch! validatePullSecretMutualExclusion runs against the request object, not the merged result, so it can miss conflicts with existing DB state. I'll handle it.

fulfillment-service/internal/servers/private_clusters_server.go
danielerez: So seems that the validatePullSecretSecret also runs after template defaults are applied (ApplyClusterSpecDefaults). thus, if a template injects pull_secret_secret via spec_defaults, the interceptor never sees it.
I'll try to simplify lookupSecret now by replacing its inline logic with NewDAOLookupFunc(s.secretsDao) and keeping just the error handling on top.

DakCrowder (approved):
Changes LGTM - looks like there is a comments diff being flagged between public / private protos?  Once resolved lmk and I can tag again.

danielerez:
> Changes LGTM - looks like there is a comments diff being flagged between public / private protos? Once resolved lmk and I can tag again.

Thanks Dakota, good catch! 
the generated code was stale because buf generate wasn't rerun after the proto comment change...
fixed here: https://github.com/osac-project/osac/pull/430

PR #391, opened 2026-08-19
OSAC-3758: Helm-manage BCM inventory and management secrets

mennyaboush:
## [OSAC-3758](https://redhat.atlassian.net/browse/OSAC-3758): Helm-manage BCM inventory and management secrets

**Jira:** https://redhat.atlassian.net/browse/OSAC-3758

### Summary

Adds Helm chart support for BCM backend configuration secrets (`osac-inventory-config` and `osac-management-config`) so deployers don't have to create them manually. When `bmf.bcm.enabled` is set to `true`, the BMF operator subchart creates both secrets with the correct BCM backend format and mounts the mTLS credentials volume. Follows the Metal3 pattern from #387.

### Changes

**BMF operator subchart** (`bare-metal-fulfillment-operator/charts/operator/`):
- Added `bcm` config block to `values.yaml` (enabled, url, credentialsSecret, insecureSkipVerify, hostClass, networkClass, metal3Namespace)
- New `bcm-inventory-config.yaml` template — creates `osac-inventory-config` Secret when BCM is enabled
- New `bcm-management-config.yaml` template — creates `osac-management-config` Secret when BCM is enabled (Metal3 management type)
- Updated `deployment.yaml` — conditional `/etc/osac/certs/<credentialsSecret>/` volume mount for BCM mTLS credentials
- All three required fields (url, credentialsSecret, metal3Namespace) have `{{- fail }}` guards

**Umbrella chart** (`osac-installer/charts/osac/`):
- Added `bmf.bcm` schema properties to `values.schema.json`
- Added `bmf.bcm` defaults to `values.yaml` (disabled by default)
- Documented BCM configuration in `values-example.yaml`
- Enabled BCM in `ci/full-values.yaml` for template validation

### Testing

- **Structural validation:** `helm template` verified three behavioral paths:
  - BCM enabled with all required fields → both secrets + certs volume mount render correctly
  - BCM disabled → no BCM resources rendered
  - BCM enabled without required fields → clear `{{- fail }}` errors
- **helm lint:** passes on both subchart and umbrella chart

### Notes

- The BCM mTLS credentials Secret (`tls.crt` + `tls.key`) must be pre-created by the admin in the install namespace — the chart references it by name, it does not template the cert content
- BCM requires Metal3 for power management, so the management config is always `type: metal3`
- Per-host BCM metadata (`resource_class`, `osac_bmc_credentials_secret`, `osac_bmc_address`) are Day-0 admin setup in BCM, not installer scope

## Summary by CodeRabbit

* **New Features**
  * Added optional integration with the Bare Metal Controller (BCM).
  * Added configuration for BCM endpoints, mutual TLS certificates, host classes, namespaces, and TLS verification.
  * Automatically creates the required BCM connection, management, and certificate secrets when enabled.
  * Added validation to prevent deployment when required BCM settings or credentials are missing.
  * Updated installer examples and schema with BCM configuration options.

### DIFF
--- a/bare-metal-fulfillment-operator/charts/operator/templates/bcm-certs.yaml
+++ b/bare-metal-fulfillment-operator/charts/operator/templates/bcm-certs.yaml
@@ -1,0 +1,22 @@
+{{- if .Values.bcm.enabled }}
+{{- if not .Values.bcm.cert }}
+{{- fail "bcm.enabled is true but bcm.cert is not set" }}
+{{- end }}
+{{- if not .Values.bcm.key }}
+{{- fail "bcm.enabled is true but bcm.key is not set" }}
+{{- end }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ .Values.secrets.bcmCerts }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "bare-metal-fulfillment-operator.labels" . | nindent 4 }}
+type: Opaque
+stringData:
+  tls.crt: {{ .Values.bcm.cert | quote }}
+  tls.key: {{ .Values.bcm.key | quote }}
+  {{- if .Values.bcm.caCert }}
+  ca.crt: {{ .Values.bcm.caCert | quote }}
+  {{- end }}
+{{- end }}
--- a/bare-metal-fulfillment-operator/charts/operator/templates/bcm-inventory-config.yaml
+++ b/bare-metal-fulfillment-operator/charts/operator/templates/bcm-inventory-config.yaml
@@ -1,0 +1,26 @@
+{{- if .Values.bcm.enabled }}
+{{- if not .Values.bcm.url }}
+{{- fail "bcm.enabled is true but bcm.url is not set" }}
+{{- end }}
+{{- if not .Values.bcm.bmhNamespace }}
+{{- fail "bcm.enabled is true but bcm.bmhNamespace is not set" }}
+{{- end }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ .Values.secrets.inventoryConfig }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "bare-metal-fulfillment-operator.labels" . | nindent 4 }}
+type: Opaque
+stringData:
+  inventory.yaml: |
+    name: bcm-inventory
+    type: bcm
+    hostClass: {{ .Values.bcm.hostClass | quote }}
+    options:
+      bcm:
+        url: {{ .Values.bcm.url | quote }}
+        credentialsSecret: {{ .Values.secrets.bcmCerts | quote }}
+        insecureSkipVerify: {{ .Values.bcm.insecureSkipVerify }}
+{{- end }}
--- a/bare-metal-fulfillment-operator/charts/operator/templates/bcm-management-config.yaml
+++ b/bare-metal-fulfillment-operator/charts/operator/templates/bcm-management-config.yaml
@@ -1,0 +1,20 @@
+{{- if .Values.bcm.enabled }}
+{{- if not .Values.bcm.bmhNamespace }}
+{{- fail "bcm.enabled is true but bcm.bmhNamespace is not set" }}
+{{- end }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ .Values.secrets.managementConfig }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "bare-metal-fulfillment-operator.labels" . | nindent 4 }}
+type: Opaque
+stringData:
+  management.yaml: |
+    name: metal3-management
+    type: metal3
+    options:
+      metal3:
+        namespace: {{ .Values.bcm.bmhNamespace | quote }}
+{{- end }}
--- a/osac-installer/charts/osac/values-example.yaml
+++ b/osac-installer/charts/osac/values-example.yaml
@@ -298,6 +298,39 @@ aap:
     key: ""
 
 # ---------------------
+# Bare Metal Fulfillment Operator — BCM Backend
+# ---------------------
+# BCM (NVIDIA Base Command Manager) inventory backend configuration for BMaaS.
+# When enabled, the chart creates osac-inventory-config, osac-management-config,
+# and osac-bcm-certs Secrets with BCM backend configuration.
+# The cert volume mount is handled by the operator wiring and sync script.
+bmf:
+  bcm:
+    # Set to true to create BCM config secrets via Helm and mount mTLS certs.
+    # When false (default), secrets must be created manually.
+    enabled: false
+    # [REQUIRED when enabled] BCM head node API endpoint.
+    # Example: "https://bcm-head.example.com:8081"
+    url: ""
+    # [REQUIRED when enabled] PEM-encoded client certificate for BCM mTLS
+    # authentication. The chart creates a Secret from this value.
+    cert: ""
+    # [REQUIRED when enabled] PEM-encoded client private key for BCM mTLS
+    # authentication.
+    key: ""
+    # [OPTIONAL] PEM-encoded CA certificate for verifying BCM server cert.
+    # When empty, system root CAs are used (or InsecureSkipVerify if enabled).
+    caCert: ""
+    # Skip TLS verification of BCM server certificate. Test environments only.
+    insecureSkipVerify: false
+    # Host class identifier written to the inventory config.
+    hostClass: "bcm"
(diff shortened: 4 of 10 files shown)


osac-installer/charts/osac/values-example.yaml
```diff
@@ -297,6 +297,36 @@ aap:
...
+    # Set to true to create BCM config secrets via Helm and mount mTLS certs.
+    # When false (default), secrets must be created manually.
+    enabled: false
+    # [REQUIRED when enabled] BCM head node API endpoint.
+    # Example: "https://bcm-head.example.com:8081"
+    url: ""
+    # [REQUIRED when enabled] Name of existing K8s Secret containing BCM mTLS
+    # client certificate (tls.crt) and private key (tls.key).
+    # This Secret must be pre-created in the install namespace.
+    credentialsSecret: ""
```
trewest: If I'm not mistaken, I think the contents of this secret needs to be provided here. The installer should be able to setup the environment without any prerequisite objects being created on the cluster

osac-installer/charts/osac/values-example.yaml
```diff
@@ -297,6 +297,36 @@ aap:
...
+    # Skip TLS verification of BCM server certificate. Test environments only.
+    insecureSkipVerify: false
+    # Host class identifier written to the inventory config.
+    hostClass: "bcm"
+    # [REQUIRED when enabled] Namespace where BareMetalHost CRs are created.
+    # BCM uses Metal3 for power management via BareMetalHost CRs.
+    # Example: "hardware-inventory"
+    metal3Namespace: ""
```
trewest: A couple of things:
1. The wording of this could be confused for "Which namespace is metal3 running in?" (typically openshift-machine-api). Maybe something better would be bmhNamespace?
2. Should this be nested under BCM or should all hostClasses leveraging metal3 provision BMHs in the same namespace? If the latter, I think it would probably make more sense at bmf.bmhNamespace

bare-metal-fulfillment-operator/charts/operator/templates/bcm-inventory-config.yaml
```diff
@@ -0,0 +1,29 @@
...
+{{- if not .Values.bcm.credentialsSecret }}
+{{- fail "bcm.enabled is true but bcm.credentialsSecret is not set" }}
+{{- end }}
+{{- if not .Values.bcm.metal3Namespace }}
+{{- fail "bcm.enabled is true but bcm.metal3Namespace is not set" }}
+{{- end }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ .Values.secrets.inventoryConfig }}
```
trewest, line 11: So based on this, it seems that only one inventory source can be used at a time per deployment? I.e. you can't use BCM and metal3 inventory sources at the same time?

bare-metal-fulfillment-operator/charts/operator/templates/bcm-management-config.yaml
```diff
@@ -0,0 +1,20 @@
...
+kind: Secret
+metadata:
+  name: {{ .Values.secrets.managementConfig }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "bare-metal-fulfillment-operator.labels" . | nindent 4 }}
+type: Opaque
+stringData:
+  management.yaml: |
+    name: metal3-management
```
trewest, line 15: Can you use more than one management backend at a time? Just wondering if this should be pulled up higher and updated based on which backends are enabled

bare-metal-fulfillment-operator/charts/operator/templates/bcm-inventory-config.yaml
carbonin: Correct

bare-metal-fulfillment-operator/charts/operator/templates/bcm-management-config.yaml
carbonin: Nope it's all just one at a time

osac-installer/charts/osac/values-example.yaml
mennyaboush: I will change it to much what you did in CaaS

osac-installer/charts/osac/values-example.yaml
mennyaboush: Good call on the naming - renamed metal3Namespace ==> bmhNamespace to describe what it controls (where BareMetalHost CRs are created) rather than which subsystem uses it.

On placement: PR #387 already has metal3.namespace for the same purpose in the Metal3-inventory backend. Since only one backend runs at a time keeping it scoped under bcm avoids cross-PR coordination right now. Both bcm.bmhNamespace and metal3.namespace write to the same options.metal3.namespace in the management config - if we want a single shared field later, we can consolidate to bmf.bmhNamespace and have both backend templates read from it.

mennyaboush:
## Response to fullsend review findings

**Addressed in this PR:**
- **value-quoting** — Added `| quote` to all string values in templates (`url`, `hostClass`, `credentialsSecret`, `bmhNamespace`). Boolean `insecureSkipVerify` left unquoted.
- **path-traversal** — No longer applicable. The cert volume mount uses a fixed path and is handled by the operator wiring (PR #261), not this PR.
- **validation-pattern-consistency** — `bmhNamespace` fail guard is intentionally kept in both `bcm-inventory-config.yaml` and `bcm-management-config.yaml`. BCM depends on Metal3 management — validating early in both templates catches missing config at install time rather than at runtime.
- **tls-bypass** — `insecureSkipVerify` defaults to `false` (secure). Same pattern as existing `aapInsecureSkipVerify`.

**Not in scope for this PR:**
- **mutual-exclusion** — PR #387 (Metal3 config) is not merged yet. Will add a fail guard when both are in main.
- **stale-doc** (README, AGENTS.md, helm deployment guide) — Documentation updates for BCM as a registered backend belong with the operator wiring PR (#261) which adds the BCM adapter to the codebase. This PR only creates the Helm config Secrets.

bare-metal-fulfillment-operator/charts/operator/templates/bcm-inventory-config.yaml
```diff
@@ -0,0 +1,26 @@
...
+type: Opaque
+stringData:
+  inventory.yaml: |
+    name: bcm-inventory
+    type: bcm
+    hostClass: {{ .Values.bcm.hostClass | quote }}
+    options:
+      bcm:
+        url: {{ .Values.bcm.url | quote }}
+        credentialsSecret: {{ .Values.secrets.bcmCerts | quote }}
```
trewest, line 24: **Bug: mount path mismatch — BCM mTLS certs will not be found at runtime**

The operator constructs the cert directory from this value:

```go
// internal/inventory/bcm.go:98 / cmd/main.go:509
certDir := filepath.Join("/etc/osac/certs", bcmCfg.CredentialsSecret)
```

So with `credentialsSecret: osac-bcm-certs` it looks for:
```
/etc/osac/certs/osac-bcm-certs/tls.crt
/etc/osac/certs/osac-bcm-certs/tls.key
```

But `deployment.yaml` (on `main`) mounts the secret at a hardcoded path:

```yaml
- name: bcm-certs
  mountPath: /etc/osac/certs/bcm-certs/   # ← should be "osac-bcm-certs"
```

There are two options:

1. The mount path should be derived from `secrets.bcmCerts` so the two stay in sync:

```yaml
mountPath: /etc/osac/certs/{{ .Values.secrets.bcmCerts }}/
```
2. Update secret.bcmCerts to match the hardcoded mount path - `bcm-certs`

Without this fix, BCM mTLS auth fails silently — the certs exist in the pod but at the wrong path.

trewest (dismissed):
Overall looks good, just one thing to address

bare-metal-fulfillment-operator/charts/operator/templates/bcm-inventory-config.yaml
mennyaboush: good catch. added the volume mount

bare-metal-fulfillment-operator/charts/operator/templates/bcm-certs.yaml
```diff
@@ -0,0 +1,19 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ .Values.secrets.bcmCerts }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "bare-metal-fulfillment-operator.labels" . | nindent 4 }}
+type: Opaque
+stringData:
+  tls.crt: {{ .Values.bcm.cert | quote }}
```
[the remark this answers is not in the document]
trewest: @mennyaboush I think you need to address this one as well

trewest (approved):

PR #393, opened 2026-08-19
OSAC-4187: proxy CSI attach/detach directly to vendor controllers (0.2 temporary)

rgolangh:
## Summary
- Implement `ControllerPublishVolume`/`ControllerUnpublishVolume` in `osac-csi-driver` by proxying directly to the vendor CSI controller for the volume's backend, mirroring the existing node-plugin vendor-socket proxy. Routing key is `osac.backend` (the **StorageBackend name** resolved from the volume's StorageTier); the fulfillment volume id is translated to the vendor-side id and the vendor's `PublishContext` is returned to the CO unchanged.
- Add a `none` sentinel endpoint so node-local backends (lvms/topolvm, `attachRequired=false`, no network CSI controller) make attach/detach a no-op instead of dialing a vendor. `AlreadyExists`/`NotFound`/`Unimplemented` vendor responses are treated as success.
- Remove the unused `ControlPlaneClient`/`ControlPlaneStub` indirection (never backed by a real fulfillment attach API); add the `--vendor-controllers` flag and wire `controller.vendorControllers` into the csi-driver chart with a StorageBackend-name-keyed default (`local=none,vast=…,pure=…,ontap=…:50051`).

> **Temporary for milestone 0.2** — the controller talks to the vendor CSI controller directly instead of going through a fulfillment control-plane attach API. Expected to be reworked in 0.3 (marked provisional in code).

## Jira
[OSAC-4187](https://issues.redhat.com/browse/OSAC-4187)

## Test plan
- [x] New/updated unit tests in `pkg/driver/controller_test.go` exercise publish/unpublish against an in-process fake vendor CSI controller: vendor-id translation, `PublishContext` forwarding, context-vs-`GetVolume` resolution, unknown-backend `NotFound`, `AlreadyExists`/`NotFound`/`Unimplemented` idempotency, and the `none` sentinel no-op.
- [x] CSI sanity suite (`test/sanity/`) green with the new `NewDriver` signature.
- [x] Unit tests pass (`make test`, `pkg/driver` 79.9% coverage)
- [x] Lint/format/build pass (`make fmt`, `make lint` 0 issues, `make build`)
- [x] `helm lint charts/csi-driver` clean; `--vendor-controllers` renders on the controller container

## Notes for reviewers
- Chart change included (`charts/csi-driver/values.yaml` default + `controller-deployment.yaml` flag) — review-sensitive per AGENTS.md. The default endpoints target the `osac-csi-backends` Services on gRPC port 50051; keys are the expected StorageBackend naming convention.
- Two ADVISORY security notes (non-blocking, both flagged for the 0.3 rework): publish trusts `osac.backend`/`osac.volume-id` volume-context over authoritative `GetVolume` (matches node-plugin trust model); secrets are forwarded to vendor controllers over the pre-existing plaintext in-cluster gRPC proxy.

---

_This PR description was drafted with AI assistance ([create-pr](https://github.com/osac-project/osac-workspace/tree/main/skills/create-pr) v0.1.3). Review for accuracy_


osac-csi-driver/pkg/driver/controller.go
```diff
@@ -168,40 +187,161 @@ func (c *ControllerServer) ControllerPublishVolume(ctx context.Context, req *csi
...
+		Readonly:         req.GetReadonly(),
+		Secrets:          req.GetSecrets(),
+		VolumeContext:    req.GetVolumeContext(),
+	}
+
+	resp, err := vendorClient.ControllerPublishVolume(ctx, vendorReq)
+	if err != nil {
 		if st, ok := status.FromError(err); ok && st.Code() == codes.AlreadyExists {
 			klog.Infof("Volume %s already published to node %s", req.GetVolumeId(), req.GetNodeId())
 			return &csi.ControllerPublishVolumeResponse{}, nil
```
akshaynadkarni, line 219: `AlreadyExists` returns an empty response here, so we drop the vendor's `PublishContext`. The attacher persists that on the VolumeAttachment and hands it to `NodeStageVolume`, so if this branch fires on a re-publish (e.g. controller restart) we overwrite a good context with an empty one and node staging loses the device/mount attrs.

A compliant vendor returns OK with the same context on re-publish rather than `AlreadyExists`, so the normal path is fine, but that's not guaranteed across Trident/VAST/Pure. Fine to leave for 0.2 with a comment; worth fixing in the 0.3 rework.

rgolangh:
waiting for #405 to be merged, then rebase and push

PR #407, opened 2026-08-20
OSAC-3735: provision a real per-test macOS keychain for the IT harness

wgordon17:
## Summary
- Alternate fix for [OSAC-3735](https://redhat.atlassian.net/browse/OSAC-3735), submitted alongside PR #207 for side-by-side comparison: instead of bypassing the keyring probe via `OSAC_SECRET_STORE=file`, this provisions a real, throwaway per-test macOS keychain (`security create-keychain` + `default-keychain -s`, both HOME-scoped to the test's sandboxed $HOME) so the IT harness genuinely exercises the OS keyring's write path instead of skipping it.
- Build-tag-gated (`it_keychain_darwin.go` / `it_keychain_other.go`) with 3 fast, cluster-free unit tests validating creation/resolution, non-interference with the real developer keychain, and a functional round-trip; hardened with a subprocess timeout, absolute `/usr/bin/security` path, restrictive file permissions, and auto-lock disabled to avoid sleep-triggered flakiness during long IT runs.
- Verified end-to-end: full unit regression suite (91/91 passing) plus a live `cli`-labeled IT run (8 Passed | 0 Failed, no keychain dialog) confirming the real `osac login` → go-keyring → keychain write path now succeeds cleanly on macOS.

## Summary by CodeRabbit

* **Tests**
  * Added macOS integration coverage for temporary keychain provisioning, permissions, locking, credential round trips, and failure scenarios.
  * Verified isolated test keychains become the default without modifying the user’s real keychain settings.
  * Added cross-platform handling so non-macOS builds continue to compile successfully.

* **Chores**
  * Added automated macOS test execution for relevant pull requests and manual runs.
  * Improved cleanup and error reporting when temporary test environments cannot be provisioned.


wgordon17:
Cross-linking for comparison: this is an alternate approach to #207 for the same underlying bug. #207 bypasses the keyring probe via `OSAC_SECRET_STORE=file`; this PR instead provisions a real, throwaway per-test macOS keychain so the probe/save logic in `config_secret_store.go` is actually exercised. Submitted side-by-side for review comparison, not as a replacement.

PR #416, opened 2026-08-20
NO-ISSUE: Import cleanapi BSR

CrystalChun:
Includes dummy package to simulate cleanapi.


fulfillment-service/buf.gen.yaml
```diff
@@ -16,8 +16,26 @@ version: v2
...
+
+  # This is only for the 'cleanapi' module, and only because there is no way
+  # to tell the Go code generator that we will not be generating code for this
+  # module: it will always include an import statement for this package. We
+  # solve that creating a dummy package.
+  - module: buf.build/cleanapi/cleanapi
+    file_option: go_package_prefix
+    value: github.com/osac-project/osac/fulfillment-service/internal/dummy
+
+  # This is for the rest of the modules, as it doesn't have a 'module' field:
```
DakCrowder: nit: this reads a bit confusing, as it states something doesn' thave a module field but all of these entries appear to have module fields

CrystalChun: True, removed it since I'm not sure how to word it lol thank you!

DakCrowder (approved):

PR #419, opened 2026-08-20
NO-ISSUE: Resolve flaky test db timeouts

DakCrowder:
The metering unit test is currently flaky and results in failures like:

https://github.com/osac-project/osac/actions/runs/32372417110/job/96435834733?pr=403

It appears like the root cause is the postgres db shutting down.  Currently the metering test db is called with a parent context from the test setup with a 2 minute timeout - this modifies the caller to not use that context so the db stays up and isn't bound to the test setup timeout.  This is also how the [fulfillment-service](https://github.com/DakCrowder/osac/blob/main/fulfillment-service/internal/database/database_container.go#L190) starts its db container.


CrystalChun (approved):
/lgtm

PR #426, opened 2026-08-20
OSAC-3468: Correct projectmembership deletion

CrystalChun:
The projectmembership deletion loop was incorrectly waiting when there were no memberships actually being deleted.

This ensures we're only waiting for projectmemberships that are actually being removed by the controller.

## Summary by CodeRabbit

* **Bug Fixes**
  * Project deletion now proceeds when all memberships are already marked for deletion.
  * Prevented projects from unnecessarily waiting on membership deletions that are already in progress.
  * Completed cleanup and finalization correctly when no new membership deletion is required.

### DIFF
--- a/fulfillment-service/internal/controllers/project/project_reconciler_function_test.go
+++ b/fulfillment-service/internal/controllers/project/project_reconciler_function_test.go
@@ -1061,6 +1061,69 @@ var _ = Describe("Deletion Cleanup", func() {
 		Expect(project.GetMetadata().GetFinalizers()).To(ContainElement(finalizers.Controller))
 	})
 
+	It("should proceed with deletion when memberships exist but all have deletion timestamps", func() {
+		project := privatev1.Project_builder{
+			Id: "project-1",
+			Metadata: privatev1.Metadata_builder{
+				Name:       "test-project",
+				Tenant:     "acme",
+				Finalizers: []string{finalizers.Controller},
+			}.Build(),
+		}.Build()
+
+		// Expect query for children (returns 0)
+		mockClient.EXPECT().
+			List(gomock.Any(), gomock.Any()).
+			Return(&privatev1.ProjectsListResponse{
+				Size: 0,
+			}, nil)
+
+		// Memberships exist but are already being deleted (have deletion timestamps)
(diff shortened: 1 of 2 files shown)


fulfillment-service/internal/controllers/project/project_reconciler_function_test.go
```diff
@@ -1294,6 +1357,14 @@ var _ = Describe("Deletion Cleanup", func() {
 				}, nil)
 
 			// Should NOT call Delete (already deleting)
```
DakCrowder: Should this comment be removed now that we are asserting it is deleted?

fulfillment-service/internal/controllers/project/project_reconciler_function_test.go
```diff
@@ -1061,6 +1061,69 @@ var _ = Describe("Deletion Cleanup", func() {
...
+		task := &task{
+			r:       functionObj,
+			project: project,
+		}
+
+		err := task.delete(ctx)
+		Expect(err).ToNot(HaveOccurred())
+		// Should complete deletion and remove finalizer
+		Expect(project.GetMetadata().GetFinalizers()).ToNot(ContainElement(finalizers.Controller))
+		// State should NOT be DELETING (was the bug)
```
DakCrowder: Can we remove (was the bug) from the comment, feels awkward to come back in the future and read

fulfillment-service/internal/controllers/project/project_reconciler_function_test.go
CrystalChun: Oh I can see how this is confusing, updated the comment! Does it make more sense now or should I still remove it?

fulfillment-service/internal/controllers/project/project_reconciler_function_test.go
CrystalChun: Good call, removed this part of the comment! Thanks!

fulfillment-service/internal/controllers/project/project_reconciler_function_test.go
DakCrowder: Ty looks good!

DakCrowder (approved):

PR #431, opened 2026-08-21
OSAC-4202: [DEV] OpenStack/Ironic Backend — FindFreeHost Port Filter and GetHostNICs

adriengentil:
## [OSAC-4202](https://redhat.atlassian.net/browse/OSAC-4202): [DEV] OpenStack/Ironic Backend — FindFreeHost Port Filter and GetHostNICs

**Jira:** https://redhat.atlassian.net/browse/OSAC-4202
**Story type:** [DEV]
**Epic:** [OSAC-4200](https://redhat.atlassian.net/browse/OSAC-4200) — Expose NIC MAC Addresses in BareMetalInstance
**Depends on:** #410 (OSAC-4201, merged)

### Summary

Implements `GetHostNICs` for the OpenStack/Ironic inventory backend and extends `FindFreeHost` to skip nodes with no registered Ironic ports. NIC MAC addresses are fetched from Ironic port records via `ports.ListDetail`, lowercased, and returned as `HostNIC` entries. The existing `isAuthError` → reconnect → retry pattern is applied to all port API calls.

### Changes

**`internal/inventory/openstack.go`:**
- `FindFreeHost`: added port filter — calls `ports.ListDetail` per candidate node; skips node on empty result or API error (auth errors propagated for reconnect)
- `GetHostNICs`: fetches all Ironic ports for a node via `ports.ListDetail`, lowercases `Port.Address`, returns `[]HostNIC`; returns error on nil/empty (node guaranteed to have ports after `FindFreeHost` filter)
- Auth/reconnect pattern applied consistently to all port calls

**`internal/inventory/openstack_internal_test.go`:** 7 new test cases using gophercloud fake HTTP server

### Acceptance Criteria

- [x] `FindFreeHost` skips nodes with no registered Ironic ports
- [x] Port API error during `FindFreeHost` treats node as ineligible (skip, not fail)
- [x] `GetHostNICs` calls `ports.ListDetail` and returns lowercased MACs
- [x] Auth/reconnect retry applied to port calls in both methods
- [x] Unit tests for all paths

## Summary by CodeRabbit

* **New Features**
  * Improved bare-metal host discovery by recognizing only hosts with registered network ports.
  * Added support for retrieving host network interfaces from OpenStack.
  * MAC addresses are now returned in a consistent lowercase format.

* **Bug Fixes**
  * Improved recovery when OpenStack authentication expires during port lookups.
  * Added clearer handling for hosts with no available ports and other port retrieval failures.

### DIFF
--- a/bare-metal-fulfillment-operator/internal/inventory/openstack.go
+++ b/bare-metal-fulfillment-operator/internal/inventory/openstack.go
@@ -28,6 +28,7 @@ import (
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/openstack"
 	"github.com/gophercloud/gophercloud/v2/openstack/baremetal/v1/nodes"
+	"github.com/gophercloud/gophercloud/v2/openstack/baremetal/v1/ports"
 	"github.com/gophercloud/gophercloud/v2/pagination"
 	"github.com/gophercloud/utils/v2/openstack/clientconfig"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
@@ -184,6 +185,7 @@ func (c *OpenStackClient) findFreeHost(ctx context.Context, matchExpressions map
 
 	var foundHost *Host
 	err := nodes.List(c.client, listOpts).EachPage(ctx, func(ctx context.Context, page pagination.Page) (bool, error) {
+		log := ctrllog.FromContext(ctx)
 		nodeList, err := nodes.ExtractNodes(page)
 		if err != nil {
 			return false, err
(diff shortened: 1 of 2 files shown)


adriengentil:
Thanks for the thorough review.

**race-condition (reconnect without lock):** This is a pre-existing pattern across the entire OpenStack backend — `c.client` is mutated without synchronization in all existing reconnect paths too. This PR adds `GetHostNICs` which follows the same pattern but doesn't introduce the race. Fixing the sync model is a cross-cutting change that should be addressed in a dedicated issue across all methods, not in this story.

**error-handling-idiom (port errors silently skip node):** The story's acceptance criteria explicitly require: *"A port API error during FindFreeHost treats the node as ineligible (skip, do not fail allocation entirely)"*. Returning the error from the EachPage callback would fail the entire allocation, which contradicts the required behavior. When all nodes fail the port check due to a sustained Ironic API failure, FindFreeHost returns `(nil, nil)` — the allocator stays in Progressing and retries on the next reconcile interval. This is intentional graceful degradation, consistent with the existing behavior for other transient errors in FindFreeHost (e.g., node provisioning state filtering).

tzumainn (approved):
/lgtm

PR #433, opened 2026-08-21
OSAC-2974: Hub secret backend

DakCrowder:
Adds a hub secret backend to the secrets apis

Additionally, this would have been the 3rd location with similar functionality in the server to fetch data from a hub cluster so consolidated some existing logic into a new package with a hub client / ability to fetch data/secrets.

NOTE: There will be conflicts with https://github.com/osac-project/osac/pull/476 - pending that being merged then will resolve.

## Summary by CodeRabbit

* **New Features**
  * Added support for retrieving secrets from Kubernetes-based hubs through the Secrets APIs.
  * Hub-backed secrets are available through private and authenticated public endpoints.
  * Secret retrieval supports selecting specific keys and validates hub, namespace, and secret details.
  * Cluster and console operations now use managed hub connections for consistent access.

* **Bug Fixes**
  * Improved handling of missing, inaccessible, or incorrectly configured hubs and secrets.

* **Tests**
  * Added coverage for hub secret retrieval, error handling, and end-to-end API behavior.


fulfillment-service/internal/servers/hub_client_provider.go
```diff
@@ -0,0 +1,199 @@
...
+type HubClientProviderBuilder struct {
+	hubLookup        HubLookup
+	hubClientFactory HubClientFactory
+}
+
+type hubClientProvider struct {
+	hubLookup        HubLookup
+	hubClientFactory HubClientFactory
+	mu               sync.RWMutex
+	clients          map[string]*HubClientInfo
```
danielerez, line 100: should we have cache invalidation for clients?
I.e. if kubeconfigs are rotated or token are expired, etc .
so subsequent API calls will return 401, but the cache won't be cleared...
maybe it's enough to add a comment about it, or implied some cache eviction on Unauthenticated.

fulfillment-service/internal/servers/hub_client_provider.go
```diff
@@ -0,0 +1,199 @@
...
+
+func (p *hubClientProvider) GetClient(ctx context.Context, hubID string) (*HubClientInfo, error) {
+	p.mu.RLock()
+	if info, ok := p.clients[hubID]; ok {
+		p.mu.RUnlock()
+		return info, nil
+	}
+	p.mu.RUnlock()
+
+	ctx = context.WithoutCancel(ctx)
```
danielerez, line 154: consider add some fixed timeout inside the 'Do'.
just so if the hub's DB record lookup stalls, we would avoid a scenario in which all waiters are blocked indefinitely.

fulfillment-service/internal/servers/console_target_resolver.go
```diff
@@ -169,81 +136,56 @@ func (r *ConsoleTargetResolver) ResolveComputeInstance(ctx context.Context, reso
...
 
-	// Phase 2: Kubernetes API call — no DB connection held.
-	config, err := clientcmd.RESTConfigFromKubeConfig(state.kubeconfig)
+	// Phase 2: Hub access — no DB connection held.
+	hubInfo, err := r.hubClientProvider.GetClient(ctx, state.ciInfo.HubID)
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to parse hub kubeconfig: %v", err)
+		return nil, err
+	}
+	if hubInfo.Namespace == "" {
```
danielerez: this check might be redundant. looks like GetClient already errors on empty namespace, so this can never fire.

danielerez (commented):
looks good, just a few comments.
also, worth fixing the PR's title.
i.e. "osac-2974: hub secret backend..."

fulfillment-service/internal/servers/hub_client_provider.go
DakCrowder: I had been considering something like this as well, I think for now evicting on Unauthenticated makes sense will take a pass at implementing that.

DakCrowder:
Re-structuring with the conflicts from main warranting a new PR for simplicity

PR #463, opened 2026-08-23
OSAC-3767: implement AssignHost BCM assignment write

mennyaboush:
## Summary

- Implements the BCM write portion of `AssignHost` for the bare-metal-fulfillment-operator, replacing the stub with full lifecycle logic
- Extends `BCMAPI` interface with `GetDevice`/`UpdateDevice` and introduces `BMHLifecycleManager` interface for testable BMH operations
- Covers all acceptance criteria: host ID parsing, device-not-found orphan detection, skip-write optimization, race detection (host taken + concurrent writer), full-object replacement with verify-after-write, and input validation

### Key design decisions
- **Verify-after-write**: defense-in-depth since BCM lacks Kubernetes-style optimistic concurrency
- **Only `osac_instance_id` written**: no tenant-identifying data in BCM (NFR-2)
- **Full-object replacement**: required because BCM's `updateDevice` rejects partial objects
- **BMHLifecycleManager interface**: enables mock-based unit testing without K8s API dependency

### Deferred
- **EventRecorder** (BCMHostAssigned/BCMConnectionError events): deferred to a dedicated task — changes the reconciler struct, needs review with Nick. See `osac-1339/OSAC-3767-deferred-event-recorder.md` for the full implementation plan.
- **ConsumerRef ownership check**: deferred to [OSAC-3768](https://redhat.atlassian.net/browse/OSAC-3768) (BMH Creation) where ConsumerRef is first set

## Test plan

- [x] 15 unit tests covering all AC paths: happy path, skip-write, host-taken, device-not-found (with/without BMH), verify-after-write race, device disappearance, full-object preservation, GetDevice/UpdateDevice/validation failures, invalid inputs
- [x] `make lint` — 0 issues
- [x] `make test` — full operator test suite passes

## Summary by CodeRabbit

- **New Features**
  - Improved bare-metal host assignment with validation, ownership checks, and safe handling of missing or conflicting devices.
  - Preserves existing device information during assignment and verifies updates successfully complete.
  - Better distinguishes unavailable hosts from hosts that are already assigned.

- **Bug Fixes**
  - Prevents accidental overwrites when assignments change concurrently or devices disappear.
  - Adds clearer handling for malformed host identifiers and failed update verification.

- **Tests**
  - Expanded coverage for assignment, conflicts, missing devices, validation, field preservation, and update failures.


bare-metal-fulfillment-operator/internal/inventory/bcm.go
```diff
@@ -213,9 +222,118 @@ func (c *BCMClient) FindFreeHost(ctx context.Context, matchExpressions map[strin
...
+
+	if device.ExtraValues != nil {
+		if existingID, ok := device.ExtraValues[bcmclient.ExtraValueInstanceID]; ok {
+			existingIDStr, _ := existingID.(string)
+			if existingIDStr == bareMetalInstanceID {
+				log.Info("BCM host already assigned to this instance, skipping write", "hostname", hostname)
+				return c.buildHost(device), nil
+			}
+			log.Info("BCM host assigned to another instance", "hostname", hostname, "existingInstanceID", existingIDStr)
+			return nil, nil
```
carbonin, line 262: Should this be an error?

bare-metal-fulfillment-operator/internal/inventory/bcm.go
```diff
@@ -213,9 +222,118 @@ func (c *BCMClient) FindFreeHost(ctx context.Context, matchExpressions map[strin
...
+		}
+	}
+
+	confirmed, err := c.writeAndVerifyAssignment(ctx, device, hostname, bareMetalInstanceID)
+	if err != nil {
+		return nil, err
+	}
+	if !confirmed {
+		return nil, nil
```
carbonin, line 271: Same here, should this be an error?

bare-metal-fulfillment-operator/internal/inventory/bcm_test.go
```diff
@@ -27,16 +27,36 @@ import (
...
 
-	"github.com/osac-project/osac/bare-metal-fulfillment-operator/internal/baremetalhost"
 	"github.com/osac-project/osac/bare-metal-fulfillment-operator/internal/bcmclient"
 )
 
 type mockBCMAPI struct{}
 
 func (m *mockBCMAPI) CertWatcher() *certwatcher.CertWatcher                    { return nil }
 func (m *mockBCMAPI) GetDevices(_ context.Context) ([]bcmclient.Device, error) { return nil, nil }
+func (m *mockBCMAPI) GetDevice(_ context.Context, _ string) (*bcmclient.Device, error) {
```
carbonin: Would it be better to generate this so we can inject errors more easily in tests?

bare-metal-fulfillment-operator/internal/inventory/bcm.go
mennyaboush: I do not think so... we may find some host that get assigned until we get to that specific code and in the current flow of the code (at the controller) returning error will retry the process with the same instanceID that was assigned already so we will get to infinity loop while return nil nil will trigger all the process again include findFreeHost that will give us different host.

bare-metal-fulfillment-operator/internal/inventory/bcm.go
mennyaboush: As the previews comment using nil nil will re triggered the flow as we wanted and there is some error the function can trow if needed but it will not trow any exception related to a race condition. 
Why do you think we need to use error here?  there is something I missed?

bare-metal-fulfillment-operator/internal/inventory/bcm_test.go
mennyaboush: will move to Mockgen

bare-metal-fulfillment-operator/internal/inventory/bcm.go
carbonin: Ah okay, I forgot about this behavior. Makes sense.

carbonin (approved):

carbonin (approved):

PR #476, opened 2026-08-24
OSAC-2981: store hub kubeconfig in Secret with dual-path readers

danielerez:
## Summary
- Add `Hub.spec.kubeconfig_secret` (`SecretLocalReference`, mutually exclusive with inline `kubeconfig`) so hub kubeconfigs can live in Kubernetes/Vault Secrets instead of PostgreSQL.
- Validate the referenced Secret exists on Hub Create/Update (by id or name).
- Dual-path kubeconfig resolution (`kubeconfig_secret` first, then inline) in `hub_cache`, `Clusters/GetKubeconfig`, and VM console, so those readers keep working after [OSAC-2984](https://redhat.atlassian.net/browse/OSAC-2984) clears inline kubeconfig.

Parent epic: OSAC-2953.

## Test plan
- [x] `ginkgo run --focus="kubeconfig_secret" internal/controllers`
- [x] `ginkgo run --focus="Kubeconfig secret reference" internal/servers`
- [x] `ginkgo run --focus="createKubeClient|privateServerHubLookup|should create a ticket" internal/servers`
- [x] Create a Hub with `kubeconfig_secret` only and confirm hub_cache, GetKubeconfig, and VM console still resolve the kubeconfig
- [x] Create a Hub with inline `kubeconfig` only and confirm the same paths still work
- [x] Confirm Create/Update reject both fields set, empty refs, and missing secrets

## Summary by CodeRabbit

* **New Features**
  * Hubs can now use kubeconfig stored in a same-tenant Secret.
  * Added validation to prevent using inline kubeconfig and Secret-based kubeconfig together.
  * Console operations and cluster connections can resolve credentials from referenced Secrets.
  * Added clear handling for missing, invalid, or unavailable Secret references.

* **Bug Fixes**
  * Improved kubeconfig refresh behavior after cached entries expire.
  * Prevented failed Secret lookups from being cached.


fulfillment-service/internal/servers/private_hubs_server.go
```diff
@@ -163,3 +191,123 @@ func (s *PrivateHubsServer) Signal(ctx context.Context,
...
+		return nil
+	}
+	ref := spec.GetKubeconfigSecret()
+	if ref == nil {
+		return nil
+	}
+	if ref.GetId() == "" && ref.GetName() == "" {
+		return grpcstatus.Errorf(grpccodes.InvalidArgument, "kubeconfig_secret must specify id or name")
+	}
+	resolved, err := references.NewDAOLookupFunc(s.secretsDao)(ctx, "", "", ref.GetId(), ref.GetName())
```
DakCrowder, line 275: I think the coderabbit feedback here is relevant (although outside the scope of this work) - it appears like  NewDAOLookupFunc doesn't use the tenant / project scoping args its is passed.  I think we should probably file a bug to fix (or not 100% sure if it is in progress or part of some other work)

DakCrowder (approved):

PR #481, opened 2026-08-24
OSAC-2982: Add client_secret_secret to IdentityProvider with dual-path reconciler

danielerez:
## Summary

- Add `SecretLocalReference client_secret_secret` to `OidcConfig` (public + private), mutually exclusive with inline `client_secret`
- Add server-side validation: reject empty `{}`, mutual exclusion (including update-mask-aware conflict vs the stored IDP), existence check via secrets DAO, and write-back of resolved `{id, name}`
- IdentityProvider reconciler prefers the Secret ref (`Secrets.Get` with controller auth, extract `data["value"]`) and falls back to inline `client_secret`
- Add unit tests for Create/Update validation and secret resolution / fallback / missing `value`

## Context

Next vertical slice of the secret references migration ([OSAC-2953](https://redhat.atlassian.net/browse/OSAC-2953)), following [OSAC-2978](https://github.com/osac-project/osac/pull/385). Uses `SecretLocalReference` (id + name) rather than a string field.

No DB migration needed. Resources use GenericDAO with a JSON `data` column, so proto field additions are picked up automatically.

## Jira

https://issues.redhat.com/browse/OSAC-2982

## Test plan

- [x] Create IDP with `client_secret_secret` by id — resolves and stores id + name
- [x] Create IDP with `client_secret_secret` by name — resolves and stores id + name
- [x] Create with both `client_secret` and `client_secret_secret` — returns INVALID_ARGUMENT
- [x] Create with nonexistent secret reference — returns INVALID_ARGUMENT
- [x] Create with empty `{}` ref — returns INVALID_ARGUMENT
- [x] Create with inline `client_secret` only — still works unchanged
- [x] Update with a valid secret ref — succeeds
- [x] Update that adds a secret ref while DB still has inline `client_secret` (mask-aware) — returns INVALID_ARGUMENT
- [x] Update with both fields set — returns INVALID_ARGUMENT
- [x] Reconciler resolves secret via Secrets API `data["value"]`
- [x] Reconciler falls back to inline `client_secret` when no ref is set
- [x] Reconciler errors when secret is missing `data["value"]`
- [x] IdentityProvider unit suite passes

## Summary by CodeRabbit

* **New Features**
  * OIDC identity providers can use a referenced Secret for their client secret.
  * Secret references can be resolved by ID or name in private and public configurations.
* **Validation**
  * Inline and referenced client secrets cannot be configured together.
  * Missing, inaccessible, or incomplete Secret references return clear validation errors.
  * Existing inline client-secret configurations continue to work.
* **Improvements**
  * Secret references are normalized to resolved resource identifiers during processing.


fulfillment-service/internal/servers/private_identity_providers_server.go
```diff
@@ -175,3 +200,128 @@ func (s *PrivateIdentityProvidersServer) Signal(ctx context.Context,
...
+// client_secret_secret.
+func (s *PrivateIdentityProvidersServer) validateClientSecretMutualExclusion(
+	idp *privatev1.IdentityProvider) error {
+	oidc := oidcFrom(idp)
+	if oidc == nil {
+		return nil
+	}
+	if oidc.GetClientSecret() != "" && oidc.GetClientSecretSecret() != nil {
+		return grpcstatus.Errorf(grpccodes.InvalidArgument, clientSecretExclusive)
+	}
```
CrystalChun, line 229: Should we also make sure at least one of these are set?

danielerez: hmm, good question:)
so actually neither field is required today (before this change).
I.e. this PR only adds a secret reference alternative and the mutual exclusion check, to avoid setting both.
Then requiring at least one would be a new constraint, which means a change in behavior...
anyway, we can have a follow up if we indeed want OIDC to always have a secret.

CrystalChun: Gotcha! Makes sense, thanks for checking that!

PR #484, opened 2026-08-24
OSAC-2988: Add pull secret ref to cluster create cli cmd

DakCrowder:
Modifies the --pull-secret arg to take an identifier for a secret ref.  This is a breaking change to existing cli behavior where this flag is for inline secret values.

Note: there is an existing --pull-secret-file that is also supported for passing files as part of the cluster create cmd.  However, the e2e tests currently reference this field.  We will need to merge this, update e2e, then circle back and clean up the file arg.

## Summary by CodeRabbit

* **New Features**
  * Added support for specifying pull secrets either by secret resource or by file.
  * Added validation to prevent using both pull-secret options at the same time.
  * Updated command help text to clarify the differences between the two options.


CrystalChun (approved):
/lgtm

PR #485, opened 2026-08-24
WIP Adding cleanapi annotations to private api

CrystalChun:


fulfillment-service/proto/private/osac/private/v1/authn_capabilities_type.proto
```diff
@@ -15,8 +15,23 @@ syntax = "proto3";
...
+  // This means that the client can select one of this and then use the OAuth discovery endpoint to find the details of
+  // the OAuth server. For example if the value is `https://my.oauth.com` then the client can obtain the details of the
+  // authorizatoin server going to the following URL:
+  //
+  //   https://my.oauth.com/.well-known/oauth-authorization-server
+  //
+  // The discovery process is defined in [RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414).
+  //
+  // Note that having an access token issued by one of this servers doesn't guarantee that the requests from the user
+  // will be accepted: it will still be subject to authorization checks that may grant or deny access.
```
CrystalChun: These comments are only in the public proto. 
Should comments that exist purely in the public proto be migrated to their private protos?

fulfillment-service/proto/private/osac/private/v1/baremetal_instance_catalog_item_type.proto
```diff
@@ -15,12 +15,15 @@ syntax = "proto3";
...
 
+import "cleanapi/cleanapi.proto";
 import "osac/private/v1/field_definition_type.proto";
 import "osac/private/v1/metadata_type.proto";
 import "osac/private/v1/baremetal_instance_template_type.proto";
 
+option (cleanapi.file).package = "osac.public.v1";
+
 // A bare metal instance catalog item defines a curated hardware offering that references an underlying bare metal
-// instance template. Admins control which fields users can set, enforce defaults, and validate input via JSON Schema.
```
CrystalChun, line 23: The removed part of this comment did not exist in the public proto. 
What should we do for these? Should we extend protoc-gen-cleanapi to allow having private comments?

fulfillment-service/proto/public/osac/public/v1/baremetal_instance_catalog_items_service.proto
```diff
@@ -70,29 +69,13 @@ message BareMetalInstanceCatalogItemsDeleteResponse {}
 
 service BareMetalInstanceCatalogItems {
   rpc List(BareMetalInstanceCatalogItemsListRequest) returns (BareMetalInstanceCatalogItemsListResponse) {
-    option (google.api.http) = {get: "/api/fulfillment/v1/baremetal_instance_catalog_items"};
   }
   rpc Get(BareMetalInstanceCatalogItemsGetRequest) returns (BareMetalInstanceCatalogItemsGetResponse) {
-    option (google.api.http) = {
-      get: "/api/fulfillment/v1/baremetal_instance_catalog_items/{id}"
-      response_body: "object"
-    };
```
CrystalChun, line 79: Double-checking that route-handling does not exist yet

CrystalChun: Adding it here https://github.com/jhernand/protoc-gen-cleanapi/pull/4

fulfillment-service/proto/private/osac/private/v1/authn_capabilities_type.proto
jhernand: Yes, I think so.

fulfillment-service/proto/private/osac/private/v1/baremetal_instance_catalog_item_type.proto
jhernand: For this particular case I think that the message can be in the public API. In general I'd say that the comments of public elements should be written with care, so that they can be copied to the pubic API, and so that we don't need to have a mechanism to mark entire comments or parts of comments as private.

fulfillment-service/proto/public/osac/public/v1/baremetal_instance_catalog_items_service.proto
jhernand: Not for now, but these HTTP options could all be automatically generated as they are always the same.

fulfillment-service/proto/private/osac/private/v1/authn_capabilities_type.proto
htayrie-rh: should have been there from the beginning i'd guess (?)

jhernand: Yes.

PR #505, opened 2026-08-25
OSAC-1927: Rename hub secrets to LVMS convention and add provider filter to hubSecretExists

akshaynadkarni:
Resolves remaining items from the [OSAC-1927 investigation comment](https://redhat.atlassian.net/browse/OSAC-1927?focusedCommentId=17978290).

PR [#305](https://github.com/osac-project/osac/pull/305) fixed the CI controller SC fallback. This PR addresses Fix 1 from the investigation plus a forward-compatible enhancement to `hubSecretExists()`.

### Fix 1: Rename hub secret prefix (`tenants.yaml`)
- Renamed `vast-tenant-config-` → `lvms-tenant-config-` to match the Ansible `lvms_storage` role convention
- Changed `data: {}` → `stringData: provider: lvms`
- Added `[osac.openshift.io/storage-provider](http://osac.openshift.io/storage-provider): lvms` label for future multi-backend filtering

### Enhancement: Optional provider filter for `hubSecretExists()`
- Added `provider string` parameter to `hubSecretExists()`
- When non-empty, adds `[osac.openshift.io/storage-provider](http://osac.openshift.io/storage-provider)` to `MatchingLabels` for provider-aware secret lookup
- Both call sites pass `""` today (no behavioral change) — ready for future multi-backend support
- Extracted `osacStorageProviderKey` constant in `tenant_names.go` following existing codebase convention

### Proto regeneration (upstream merge)
- Merged upstream `main` to pick up [OSAC-4195](https://redhat.atlassian.net/browse/OSAC-4195) (cleanapi annotations on proto types)
- Updated `buf.gen.yaml` with cleanapi module redirect
- Created `internal/dummy/cleanapi/` stub package
- Regenerated 12 `.pb.go` files

### ~~Fix 2: Extend register-local-storage hook~~ (dropped)
Adding `kubectl apply` for per-tenant hub Secrets in the hook required K8s RBAC that the hook's ServiceAccount doesn't have (it only uses `curl` to the fulfillment API). The `tenants.yaml` template (Fix 1) already creates these Secrets. Self-containment can be added in a follow-up with proper Role + RoleBinding.

### ~~Fix 3: Probe correct service in init container~~ (dropped)
Changing the `waitForFulfillment` probe from `fulfillment-rest-gateway:8000/healthz` to `fulfillment-internal-api:8001` broke all E2E because the private API requires authentication. The original unauthenticated health endpoint is sufficient since both services back the same pod.

## Test plan
- Updated `createHubSecret` test helper to use `lvms-tenant-config-` prefix, `provider: lvms` data, and storage-provider label
- Updated direct `vast-tenant-config-` reference in deprovisioning test
- Added new test context for `hubSecretExists` provider filtering:
  - `provider="lvms"` → matches LVMS-labeled secret ✅
  - `provider="vast"` → does not match ✅
  - `provider=""` → matches any (backward compat) ✅
- `go vet ./...` passes
- `go test ./...` passes (73.5% coverage on controller package)
- `make fmt` and `make lint` pass

## Summary by CodeRabbit

* **New Features**
  * Added support for LVMS storage provider configuration.
  * Storage resources can now be matched using a specific provider.
  * Added support for clean API package generation.

* **Bug Fixes**
  * Updated tenant configuration and resource adoption to recognize LVMS-specific Secrets.
  * Improved storage Secret detection for matching and non-matching providers.

* **Documentation**
  * Updated storage configuration references to reflect LVMS terminology.


zszabo-rh (changes requested):
The changes look right and address the root causes as intended, so code-wise I'm good (happy to lgtm&approve).
But all three e2e gates are currently failing at the same post-install step which is the exact hook this PR modified, which should be investigated first.

akshaynadkarni:
Superseded by #568 — rebased onto current `main` and dropped the proto-regeneration commit, since [OSAC-4195](https://redhat.atlassian.net/browse/OSAC-4195) (cleanapi annotations) already landed on `main` directly via #417. Closing this one in favor of #568.

PR #507, opened 2026-08-25
NO-ISSUE: Fix merge-queue e2e ref resolution falling back to ephemeral gh-readonly-queue branch

eliorerz:
## Live incident

The merge queue has been failing ~100% of runs since ~17:05 today.

## Root cause

`merge_group` events have no `pull_request` context, so any
`github.event.pull_request.head.ref || github.ref_name` fallback resolves
`github.ref_name` to GitHub's ephemeral `gh-readonly-queue/main/pr-NNN-<sha>`
branch. That branch gets replaced/deleted whenever any queued PR fails or is
dequeued, so any downstream clone/build against it fails with:

```
fatal: Remote branch gh-readonly-queue/main/pr-NNN-<sha> not found in upstream origin
```

Confirmed against real failing merge-queue job logs, not speculation.

## Fix

Fall back to `github.sha` instead of `github.ref_name`. `github.sha` is the
exact commit under test for `merge_group` (the queue's merge-preview commit),
and resolves to the same commit `github.ref_name` would have at trigger time
for `schedule`/`workflow_dispatch` -- just immutable instead of a movable
branch name. `pull_request.head.ref` is unaffected and left as the first
choice in all cases.

Same pattern found and fixed in two places:
- `.github/actions/resolve-e2e-components/action.yaml` -- the originally
  reported root cause, but currently **unreferenced anywhere in this repo**
  (dead code today).
- `.github/workflows/e2e-bmaas-full-install.yml`,
  `e2e-caas-full-install.yml`, `e2e-vmaas-full-install.yml` --
  `installer-ref` and each component's inline `ref` in the `components` JSON.
  These are the actually-wired-in callers and where the live failures are
  coming from.

## Companion fix

A separate agent is fixing the corresponding downstream git-clone step in
`osac-test-infra` in parallel. **Both PRs are needed to fully resolve the
incident**, but this half is independently correct and safe to merge alone --
it does not depend on or block the other one.

## Test plan
- [x] All 4 changed files validated as parseable YAML
- [x] Confirmed no other `pull_request.head.ref || github.ref_name` occurrences remain in `.github/`
- [x] Confirmed the one remaining unrelated `github.ref_name` use (`nightly-build.yaml`'s Slack failure notification) serves a different purpose and was left untouched
- [ ] Merge queue run succeeds end-to-end once both PRs are merged

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved end-to-end workflow reliability when running outside pull requests.
  * Installer and component builds now consistently use the correct branch or commit reference.
  * Prevented reference resolution issues in automated installation and image-building workflows.


eliorerz:
Verified CodeRabbit's finding against the actual downstream code in osac-test-infra (read-only, not this repo):

- The composite action's `REF` and each workflow's inline `components[].ref` do **not** go through `git clone --branch` — they're consumed via `git -C "$src" fetch --depth 1 "$url" "$ref"` + `checkout FETCH_HEAD` in osac-test-infra's e2e-*-full-install.yml build step, which handles a raw commit SHA fine. No checkout-logic change needed for those.
- The one place a bare `git clone --branch` still exists is the **installer clone's fallback path**, gated on `COMPONENT_BUILD_RESULTS` being empty. For `e2e-bmaas-full-install.yml`/`e2e-caas-full-install.yml`, `components[]` is always statically non-empty on a real run, so that fallback is unreachable today -- this PR's `installer-ref` change doesn't newly break anything reachable.
- Companion PR #414 (already merged to osac-test-infra main) already fixed this exact class of issue for **vmaas** (`CLONE_REF="${PINNED_SHA:-${INSTALLER_REF}}"` + `git fetch`), but left bmaas/caas on the old `git clone --branch` fallback. Worth closing out for consistency in a follow-up there, but it's dead code today and doesn't block this PR.

Not making any further changes here based on this -- verified as not applicable to this repo's files.

PR #545, opened 2026-08-26
Edge cases for cleanapi

CrystalChun:
## Summary by CodeRabbit

* **New Features**
  * Added API operations to retrieve cluster Kubeconfig files and administrator passwords through gRPC and HTTP.
  * Expanded event payloads to include additional resource types.
  * Added an event type for changes that trigger reconciliation.

* **API Updates**
  * Improved event visibility controls for private resource data.
  * Updated cluster API documentation and endpoint organization.

### DIFF
--- a/fulfillment-service/proto/public/osac/public/v1/event_type.proto
+++ b/fulfillment-service/proto/public/osac/public/v1/event_type.proto
@@ -15,23 +15,44 @@ syntax = "proto3";
 
 package osac.public.v1;
 
+import "osac/public/v1/baremetal_instance_catalog_item_type.proto";
+import "osac/public/v1/baremetal_instance_template_type.proto";
 import "osac/public/v1/baremetal_instance_type.proto";
+import "osac/public/v1/cluster_catalog_item_type.proto";
 import "osac/public/v1/cluster_template_type.proto";
-import "osac/public/v1/cluster_type.proto";
 import "osac/public/v1/cluster_version_type.proto";
+import "osac/public/v1/cluster_type.proto";
+import "osac/public/v1/compute_instance_catalog_item_type.proto";
 import "osac/public/v1/compute_instance_template_type.proto";
 import "osac/public/v1/compute_instance_type.proto";
 import "osac/public/v1/disk_image_type.proto";
+import "osac/public/v1/external_ip_attachment_type.proto";
+import "osac/public/v1/external_ip_pool_type.proto";
+import "osac/public/v1/external_ip_type.proto";
 import "osac/public/v1/host_type_type.proto";
+import "osac/public/v1/hub_type.proto";
+import "osac/public/v1/identity_provider_type.proto";
 import "osac/public/v1/instance_type_type.proto";
-import "osac/public/v1/role_binding_type.proto";
+import "osac/public/v1/nat_gateway_type.proto";
+import "osac/public/v1/network_class_type.proto";
+import "osac/public/v1/project_membership_type.proto";
 import "osac/public/v1/project_type.proto";
-import "osac/public/v1/secret_type.proto";
-import "osac/public/v1/role_type.proto";
 import "osac/public/v1/tenant_type.proto";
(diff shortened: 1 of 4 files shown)


fulfillment-service/proto/public/osac/public/v1/event_type.proto
```diff
@@ -15,23 +15,44 @@ syntax = "proto3";
...
+import "osac/public/v1/storage_backend_type.proto";
+import "osac/public/v1/storage_tier_type.proto";
+import "osac/public/v1/subnet_type.proto";
+import "osac/public/v1/user_type.proto";
+import "osac/public/v1/virtual_network_type.proto";
+import "osac/public/v1/volume_type.proto";
 
 // Represents events delivered by the server.
 message Event {
+  reserved 16, 17, 18, 25; // 16-17, 25: Previously PublicIP resources, renamed to ExternalIP. 18: Previously Organization, removed in favor of Tenant.
```
CrystalChun, line 54: This was only present on the private proto
Should we extend protoc-gen-cleanapi to handle this? 

Currently it throws an error if I add 
```suggestion
  reserved 16, 17, 18, 25 [(cleanapi.field).private = true]; // 16-17, 25: Previously PublicIP resources, renamed to ExternalIP. 18: Previously Organization, removed in favor of Tenant.
```

Or should reserved fields exist on the public proto?

fulfillment-service/proto/public/osac/public/v1/virtual_network_type.proto
```diff
@@ -50,9 +53,6 @@ message VirtualNetwork {
 // The spec contains user-specified parameters that define how the network should be configured. These fields
 // follow a declarative model where users specify the desired state and the system reconciles to match it.
 message VirtualNetworkSpec {
-  // Reserved field numbers for removed fields (do not reuse).
-  reserved 2;
-  reserved "network_class";
```
CrystalChun, line 55: How about vice-versa if a public proto has reserved fields, should it be copied to the private proto to preserve them? 
What about field numbering? Should we be concerned about field numbering changes?
E.g. https://github.com/osac-project/osac/pull/538/changes/9aed9dcde85720298e5abfba8af90e6ea7a3f15a#r3866618436

fulfillment-service/proto/private/osac/private/v1/clusters_service.proto
```diff
@@ -39,6 +81,30 @@ message ClustersGetResponse {
...
+  string id = 1;
+}
+
+message ClustersGetPasswordResponse {
+  string password = 1;
+}
+
+message ClustersGetPasswordViaHttpRequest {
+  string id = 1;
+}
```
CrystalChun: These functions only existed on the public proto and adding them throws this
```
internal/controllers/cluster/cluster_reconciler_function.go:1: : # github.com/osac-project/osac/fulfillment-service/internal/controllers/cluster [github.com/osac-project/osac/fulfillment-service/internal/controllers/cluster.test]
internal/controllers/cluster/cluster_reconciler_function_test.go:894:27: cannot use clustersClient (variable of type *MockClustersClient) as privatev1.ClustersClient value in structliteral: *MockClustersClient does not implement privatev1.ClustersClient (missing method GetKubeconfig)
```

For the CaaS team: do we want to add these to the cluster client mock?

fulfillment-service/proto/public/osac/public/v1/event_type.proto
jhernand: Reserving and not reusing numbers makes sense when the system has been released and is already in use. We are not yet there, so I think these "reserved ..." should be removed and the fields should be renumbered.

fulfillment-service/proto/public/osac/public/v1/virtual_network_type.proto
jhernand: This will be a concern when we have released the software. When we do that we should indeed copy to the public API the "reserved ..." specification if it affects public fields.

PR #563, opened 2026-08-27
OSAC-4542: add read-only public Volume Get/List API

zszabo-rh:
## [OSAC-4542](https://redhat.atlassian.net/browse/OSAC-4542): read-only public Volume Get/List API

First phase of the public Volume API ([OSAC-984](https://redhat.atlassian.net/browse/OSAC-984)): a read-only public `Volumes`
service (`Get` + `List`, gRPC + REST at `/api/fulfillment/v1/volumes`) so the
console UI and tenants can list and view the standalone volumes provisioned via
the private Volume API ([OSAC-2872](https://redhat.atlassian.net/browse/OSAC-2872)). No create/update/delete here — those stay on
the private API and come under later [OSAC-984](https://redhat.atlassian.net/browse/OSAC-984) tickets.

**Jira:** [OSAC-4542](https://redhat.atlassian.net/browse/OSAC-4542)
**Design:** osac-project/enhancement-proposals#237
**PRD:** osac-project/enhancement-proposals#236

### How

- **proto:** opt the private `Volumes` service + `Volume` type into public
  generation via cleanapi, exposing only `List`/`Get`. Internal routing fields
  (`backend`, `protocol`, `hub`, `vendor_volume_id`) are stripped from the public
  type via `[(cleanapi.field).private = true]`; `StorageProtocol`/`storage_common_type`
  stay private.
- **servers:** new public `VolumesServer` delegates to `PrivateVolumesServer`
  (tenant-scoped reads) and maps private→public, dropping internal fields. The
  private server's tier resolver is now optional (Create-only) and gained
  `SetFilterDesc` so public CEL filters resolve against public fields.
- **wiring:** register the public `Volumes` gRPC service + REST handler.
- **authz:** allow tenant clients on the public `Volumes/Get` + `/List` methods.
- **tests:** public server List/Get + field mapping; authz allow (read) / deny
  (mutations); a filter on a private-only field is rejected; and a generated-schema
  guard asserting no internal field appears in the public `Volume`.

### Scoping decision (⚠️ team input requested)

Reads are **tenant-level scoped**, consistent with every other resource. The
DoD's "Tenant User sees only their own" is a per-creator boundary that **no
platform resource implements today** and would need net-new cross-cutting
tenancy plumbing. It's deferred to a **separate platform tenancy feature** (no
migration needed later — `creator` column + index already exist). This is a
recorded product narrowing, not a cross-tenant leak (every returned row is within
the caller's own tenant). **Team: own that follow-up here, or delegate to a
platform/tenancy owner?** See design § RBAC / Tenancy and Open Questions.

### Notes

- **UI alignment:** `osac-ux` is deprecated (osac-project/osac-workspace#224);
  `osac-ui` generates UI types directly from this proto. No public Volume UI types
  exist yet, so there are **no deviations to reconcile** — the UI regenerates from
  this proto once it lands (confirmed with the UI owner).
- Public proto import pruning was done by hand (cleanapi v0.0.8 doesn't prune;
  `buf lint IMPORT_USED` enforces it) — matches existing read-only public
  resources; a tooling follow-up to automate this is a candidate.
- OpenAPI is derived from the proto HTTP annotations (no in-repo spec).
- **CI note:** the non-required `Check generated code (osac-operator)` job is red
  due to a **pre-existing, repo-wide** staleness of the committed osac-operator
  generated code (reproducible on a clean `main`; tracked separately by infra).
  This PR does not introduce it and it will clear on rebase once the infra fix
  lands; the required checks and `Check generated code (fulfillment-service)` are
  green.

_Part of a 3-PR set for [OSAC-4542](https://redhat.atlassian.net/browse/OSAC-4542): PRD (#236) → design (#237) → **code** (this)._

Assisted-by: Claude Code

## Summary

- **API surface**
  - Added the public, read-only `Volumes` API.
  - Added gRPC and REST `List` and `Get` methods at `/api/fulfillment/v1/volumes`.
  - Added pagination, filtering, ordering, totals, access modes, and lifecycle states.
  - Excluded internal routing and vendor fields from public responses.
  - Added validation for storage tier, capacity, and access mode.

- **Service registration and deployment**
  - Added tenant-scoped public volume reads.
  - Registered public gRPC and REST handlers when VMaaS is enabled.
  - Added service-flag gating for CaaS, VMaaS, and BMaaS registration.
  - Added filter-descriptor support to the private volume server.
  - Allowed read-only construction without a tier resolver. `Create` returns an internal error when the resolver is missing.

- **Authorization**
  - Allowed tenant-scoped users to call `Volumes/List` and `Volumes/Get`.
  - Denied public `Create` and `Delete` operations.

- **Tests**
  - Added tests for mapping, tenant visibility, authorization, CEL filtering, private-field filter rejection, schema protection, builder validation, server registration, and missing tier-resolver behavior.

- **Backward compatibility**
  - The private API remains available.
  - The public API is additive and read-only.
  - CleanAPI now removes internal fields from the generated public `VolumeStatus` type.
  - Per-creator visibility remains deferred.
  - VMaaS and BMaaS full-install E2E workflows failed without diagnostic artifacts, so they did not identify a code-specific cause.
  - The operator generated-code check remains affected by pre-existing repository-wide staleness.

## Risk classification

- **risk:show** — The change adds a public API and authorization rules. It is read-only, tenant-scoped, validated, and covered by focused tests.
- It does not qualify for **risk:ship** because it changes the public API surface and access policy.
- It does not qualify for **risk:ask** because it does not modify existing data, expose private fields, or enable public mutation.


fulfillment-service/internal/cmd/service/start/grpcserver/unknown_service_handler_test.go
```diff
@@ -92,6 +92,23 @@ var _ = Describe("UnknownServiceHandler", func() {
...
+
+		err := invokeMethod(conn, "/osac.public.v1.Volumes/List")
+		Expect(err).To(HaveOccurred())
+
+		st, ok := status.FromError(err)
+		Expect(ok).To(BeTrue())
+		Expect(st.Code()).To(Equal(codes.Unavailable))
+		Expect(st.Message()).To(Equal("the VMaaS service is not enabled on this server"))
+	})
+
```
rgolangh: why do we block an API if vmaas is disabled? what about BMaaS or any other client that may need access? not sure what is the requirement here, can you explain @zszabo-rh ? 
@redhat-chai-bot can you explain this requirement perhaps?

akshaynadkarni: @zszabo-rh I checked with chai-bot on the existing code structure, and my interpretation is that we don't have to limit volumes to VMaaS. Can you take another look to see if we can remove the `if VMaaS` condition? I will connect with you tomorrow.

zszabo-rh: done!

akshaynadkarni (approved):
Thanks for updating the code to remove the `if VMaaS` condition. 

/lgtm

PR #582, opened 2026-08-27
OSAC-2988: Deprecate --pull-secret-file inline arg

DakCrowder:
Removes the --pull-secret-file argument that places file contents inline for a given cluster.

I had thought this would need to be a back and forth updating e2e then the cli, but it turns out now e2e tests have moved into the monorepo as well as nothing in the e2e actually using the pull-secret-file arg.

## Summary by CodeRabbit

* **New Features**
  * Cluster creation now references pull secrets by Kubernetes Secret name.
  * Added support for creating secrets from one or more named files.

* **Bug Fixes**
  * Updated guidance and validation messages to reflect Secret-based pull-secret configuration.

* **Documentation**
  * Updated catalog creation examples and dependency guidance for the revised pull-secret workflow.


CrystalChun (approved):
/lgtm

PR #584, opened 2026-08-27
OSAC-1574: Add e2e tests for projects

CrystalChun:
Actually running this test suite will be added in a later patch once the migration from test infra to this repo is complete.
https://redhat.atlassian.net/browse/OSAC-3593

### DIFF
--- a/tests/core/grpc_client.py
+++ b/tests/core/grpc_client.py
@@ -750,3 +750,22 @@ class GRPCClient:
 
     def delete_storage_tier(self, *, tier_id: str) -> None:
         self.call(service=f"{PRIVATE_API}.StorageTiers/Delete", data={"id": tier_id})
+
+    # Project operations (public API)
+
+    def create_project(self, *, name: str, parent_project_id: str | None = None) -> str:
+        obj: dict[str, Any] = {"metadata": {"name": name}}
+        if parent_project_id is not None:
+            obj["spec"] = {"parent_project": {"id": parent_project_id}}
+        response: dict[str, Any] = self.call(service=f"{PUBLIC_API}.Projects/Create", data={"object": obj})
+        return response["object"]["id"]
+
+    def get_project(self, *, project_id: str) -> dict[str, Any]:
+        return self.call(service=f"{PUBLIC_API}.Projects/Get", data={"id": project_id})
+
+    def list_project_ids(self) -> list[str]:
+        response: dict[str, Any] = self.call(service=f"{PUBLIC_API}.Projects/List")
+        return [item["id"] for item in response.get("items", [])]
+
+    def delete_project(self, *, project_id: str) -> None:
+        self.call(service=f"{PUBLIC_API}.Projects/Delete", data={"id": project_id})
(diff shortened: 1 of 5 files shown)


tests/core/grpc_client.py
```diff
@@ -750,3 +750,21 @@ def create_storage_tier(self, *, name: str, backend_id: str) -> str:
 
     def delete_storage_tier(self, *, tier_id: str) -> None:
         self.call(service=f"{PRIVATE_API}.StorageTiers/Delete", data={"id": tier_id})
+    # Project operations (public API)
```
DakCrowder, line 754: Nit: can we add a newline above this

tests/projects/README.md
```diff
@@ -0,0 +1,50 @@
...
+7. **Tenant Isolation**: Verify users from other tenants cannot access projects
+8. **Deletion Constraints**: Parent project with children cannot be deleted
+9. **Cascade Cleanup**: Child deletion removes from both gRPC and Keycloak
+10. **Complete Cleanup**: Parent deletion after children are removed
+
+## Running Tests
+
+```bash
+# Run all projects tests
+make test-projects
```
DakCrowder: Does this makefile target exist (or will exist?)

tests/projects/README.md
```diff
@@ -0,0 +1,50 @@
...
+- **Keycloak**: Must be accessible for organization/group verification
+- **Multiple Tenants**: tenant1 and tenant2 must exist
+- **JWT Authentication**: tenant1_admin, tenant1_user, tenant2_user must be configured
+- **Environment Variables**:
+  - `OSAC_KEYCLOAK_URL`: Keycloak base URL
+  - `OSAC_KEYCLOAK_ADMIN_PASSWORD`: Keycloak admin password (default: "admin")
+  - `OSAC_JWT_PASSWORD`: Password for JWT test users (default: "foobar")
+  - `OSAC_SKIP_KEYCLOAK_SYNC`: Set to "true" to skip Keycloak sync verification (default: "false"). Useful when the project sync feature isn't deployed yet.
+
+## Architecture Notes
```
DakCrowder: I think we can probably remove these notes, I imagine there are other docs/sources of truth that are better for why Projects are what they are than here

DakCrowder (commented):
I think the test structure itself looks good but have a couple comments - also looks like pre commit linting is failing and it wanta a newline at the end of `keycloak_admin.py`

tests/projects/README.md
CrystalChun: Oops this was a relic from when I was adding it to osac-test-infra before the migration to this repo. Thanks for catching it! Removed it.

tests/projects/README.md
CrystalChun: Makes sense, thanks for the review! Removed it.

tests/core/grpc_client.py
CrystalChun: Yep! The new line wasn't added after I rebased off of main. Added, thanks!

PR #585, opened 2026-08-27
OSAC-4196: Undo type changes where field numbers were updated/changed

CrystalChun:
Backing out changes where field numbers were updated/changed in https://github.com/osac-project/osac/pull/538

Causes breakage such as https://github.com/osac-project/osac/actions/runs/33093913927/attempts/1

The problem is a wire format incompatibility between the server and client:
     
  1. Server (e.g. osac.private.v1.ComputeInstance): Sends messages with field 4 (hub)
     populated
  2. Client/CLI (e.g. osac.public.v1.ComputeInstance): Expects messages where field 4
     doesn't exist

## Summary by CodeRabbit

* **New Features**
  * Cluster status now supports kubeconfig and password secret references.
  * OAuth authentication supports configurable audience values.
  * Tenant setup provisions the required Vault namespace before storing break-glass credentials.
  * OpenBao lifecycle access now supports namespace, mount, and policy management.

* **API Updates**
  * Updated status schemas and field numbering across bare-metal, cluster, compute, and external IP resources.
  * Previously restricted status and configuration fields are now available through the API.
  * Reserved a field number in bare-metal instance type specifications.


CrystalChun:
/hold

This didn't work...

PR #587, opened 2026-08-27
NO-ISSUE: Revert api types

CrystalChun:
Field changes causes wiring failures in e2e tests. Backing out the changes for now, but keeping the fix for the spacing issue in the last commit

## Summary by CodeRabbit

* **API Changes**
  * Updated resource schemas and field numbering across compute, cluster, bare-metal, and external IP status responses.
  * Added a restart-required compute instance condition.
  * Made selected network attachment fields mutable.
  * Reserved retired field numbers to prevent reuse.
* **Validation**
  * Relaxed identifier and secret data validation rules.
  * Updated external IP pool configuration requirements.
* **Documentation**
  * Clarified resource lifecycles, administrator-managed pools, secret backends, and template parameters.
  * Refined system-only API definitions and descriptions.


CrystalChun:
/close

no longer needed

PR #683, opened 2026-09-01
OSAC-4741: Pass enable-ai-diagnostic through to scheduled E2E full-install runs

eliorerz:
## Summary

Companion to [osac-test-infra#435](https://github.com/osac-project/osac-test-infra/pull/435), which adds the `enable-ai-diagnostic` input (AI-powered root-cause diagnosis on failure, written to the job summary) to the three reusable E2E full-install workflows. This PR passes that input through from the mono-repo's own callers, true only for `schedule`-triggered runs — matching the existing `notify-on-failure` pattern already used by
`osac-test-infra`'s own callers (this mono-repo's callers never wired that one up either; pre-existing asymmetry, not touched here).

## ⚠️ Merge-order dependency — do not merge before osac-test-infra#435

This repo's callers pin to `osac-test-infra/...@main`. GitHub validates `workflow_call` inputs strictly — passing `enable-ai-diagnostic` here before the reusable workflow on `osac-test-infra`'s `main` actually defines it would break **every** scheduled (and PR-triggered) E2E run in this repo immediately with an "unexpected input" error.

**[osac-test-infra#435](https://github.com/osac-project/osac-test-infra/pull/435)
must merge and be confirmed live on `main` before this PR merges.**

## Test plan

- [x] YAML syntax validated on all three changed workflow files
- [ ] Blocked on osac-test-infra#435 landing first — then confirm a real scheduled run picks this up cleanly

## Summary by CodeRabbit

* **Chores**
  * AI-powered failure diagnostics are now enabled for scheduled end-to-end test runs.
  * Pull request and merge-group test runs continue without AI diagnostics, ensuring compatibility with forked execution environments.
  * End-to-end test workflows now consistently use the appropriate repository and revision when sourcing shared tests.

### DIFF
--- a/.github/workflows/e2e-bmaas-full-install.yml
+++ b/.github/workflows/e2e-bmaas-full-install.yml
@@ -157,6 +157,13 @@ jobs:
       test-source: "osac"
       test-repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
       test-ref: ${{ github.event.pull_request.head.sha || github.sha }}
+      # AI failure diagnosis only for scheduled runs: those always run
+      # against main in this repo's own context (no fork involved), unlike
+      # pull_request/merge_group runs from PRs, which come from genuine
+      # forks and cannot get OIDC/WIF access at all regardless of this
+      # input (a GitHub Actions security policy, not something this input
+      # can override).
+      enable-ai-diagnostic: ${{ github.event_name == 'schedule' }}
 
   e2e-bmaas-gate:
     if: always()
--- a/.github/workflows/e2e-caas-full-install.yml
+++ b/.github/workflows/e2e-caas-full-install.yml
@@ -169,6 +169,13 @@ jobs:
       # flavor-name, namespace, etc.) are left at the reusable workflow's own
       # defaults -- none are required and this caller has no need to override
       # them today.
+      # AI failure diagnosis only for scheduled runs: those always run
+      # against main in this repo's own context (no fork involved), unlike
+      # pull_request/merge_group runs from PRs, which come from genuine
+      # forks and cannot get OIDC/WIF access at all regardless of this
+      # input (a GitHub Actions security policy, not something this input
+      # can override).
+      enable-ai-diagnostic: ${{ github.event_name == 'schedule' }}
 
   e2e-caas-gate:
     if: always()
--- a/.github/workflows/e2e-vmaas-full-install.yml
+++ b/.github/workflows/e2e-vmaas-full-install.yml
@@ -165,6 +165,13 @@ jobs:
       test-source: "osac"
       test-repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
       test-ref: ${{ github.event.pull_request.head.sha || github.sha }}
+      # AI failure diagnosis only for scheduled runs: those always run
+      # against main in this repo's own context (no fork involved), unlike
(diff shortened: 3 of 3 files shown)


eliorerz:
Rebased onto latest main — conflicts (from OSAC-3593's dual test-source support landing in the meantime) resolved by keeping both blocks; confirmed `mergeable: MERGEABLE` now.

On the two findings:

- **"Ensure reusable workflows declare the input before retaining callers' scheduled-run expressions, then run one scheduled job per suite to validate"** — this restates the merge-order dependency already called out explicitly in this PR's description (and mirrored in [osac-test-infra#435](https://github.com/osac-project/osac-test-infra/pull/435)'s description): #435 must merge and be live on `osac-test-infra`'s `main` before this PR merges, since these callers pin to `@main` and GitHub validates `workflow_call` inputs strictly. No additional code change needed here — the live-validation-per-suite step is already in this PR's test plan, blocked on #435 landing first, not something to do before that.
- **"Replace the mutable `@main` reusable-workflow references with pinned commit SHAs"** — this is pre-existing code on all three lines (`e2e-{caas,bmaas,vmaas}-full-install.yml`'s `uses: osac-project/osac-test-infra/.github/workflows/...@main`), not something this PR's diff touches or introduces, and it's the same convention used by every other cross-repo reference in this file (e.g. `check-e2e-readiness@main`). Skipping it here: (1) out of scope for this PR's actual change, (2) pinning to a SHA *right now* would break this exact PR, since it would point at a commit before `enable-ai-diagnostic` exists on `osac-test-infra` — the whole point of `@main` here is to pick up #435's merge automatically once it lands, which is the intentional design for this pair of PRs. Worth considering as a separate, repo-wide convention change later, not bundled into this one.

eliorerz:
Verified both findings against the actual current code — both concern pre-existing code from [OSAC-3593](https://redhat.atlassian.net/browse/OSAC-3593) (#433, already merged independently of this PR), and neither holds up:

- **SHA-pin the `@main` reusable-workflow refs** (`e2e-{bmaas,vmaas}-full-install.yml`'s `uses: osac-project/osac-test-infra/.github/workflows/...@main`) — pre-existing, same convention used elsewhere in these files (`check-e2e-readiness@main`) and across the repo. The dependency risk that made pinning actively dangerous before this PR is gone now that #435 is merged, but pinning just these lines while leaving every other identical cross-repo `@main` reference untouched would be an inconsistent partial fix, not a real answer to "unpinned cross-repo refs" as a systemic concern. Skipping — better as its own deliberate, repo-wide policy change, not bundled into this one.

- **merge_group loads untrusted test code with secrets while authorization is empty** — checked this specifically and it doesn't hold. `github.event.pull_request` is undefined for `merge_group` events (that event has its own `merge_group` payload key, no `pull_request` key). So `test-repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}` and `test-ref: ${{ github.event.pull_request.head.sha || github.sha }}` both correctly fall through to `github.repository`/`github.sha` for merge_group — the trusted base repo, at GitHub's own server-computed merge-queue test commit, not arbitrary fork content. The `Authorize fork PR` step being skipped there is correct, not a gap: entry into the merge queue already requires the PR to have passed required review/approval — that's the real authorization checkpoint, and it happens before merge_group, not at it. No code change made here — the described vulnerability doesn't exist in the current code.

PR #697, opened 2026-09-01
OSAC-4741: Add AI diagnostic Check Run for fork PR E2E failures

eliorerz:
## Summary
Phase 2 of AI failure diagnostics: fork PR E2E failures get a diagnosis too, via a Check Run instead of a job summary. Mirrors [osac-test-infra#443](https://github.com/osac-project/osac-test-infra/pull/443) 1:1 (`vertex-ai-auth`, `ai-diagnose-failure.py`, and `ai-diagnostic-e2e.yml`) -- `workflow_run` only fires for runs in the repo the listener lives in, so each repo needs its own copy to cover PRs against itself.

Phase 1 (ref-fixed in #696) writes the diagnosis into the failing job's own Step Summary via WIF auth in the same job -- but that path structurally can't work for fork PRs: GitHub gives `pull_request` jobs whose head is a fork a read-only token with **no OIDC minting at all**, unconditionally, by policy. Every real PR on this repo is fork-originated today, so Phase 1 alone never reaches fork PRs regardless of any input.

`.github/workflows/ai-diagnostic-e2e.yml`:
1. Fires on completion of `E2E CaaS/BMaaS/VMaaS Full Install`, gated to `conclusion == 'failure' && event == 'pull_request'` (schedule/workflow_dispatch already get Phase 1; merge_group has no single PR to attach to).
2. Resolves the originating PR from the failed run's head SHA by listing open PRs and requiring a unique match -- `github.event.workflow_run.pull_requests[]` is unreliable/empty for genuine cross-fork PRs, same gap `e2e-on-approval-fork.yml` already works around.
3. Downloads that run's own `osac-logs-*` artifact (already gitleaks-scanned/redacted by `gather-artifacts` in osac-test-infra's reusable workflow).
4. Authenticates via a new `vertex-ai-auth` composite action (this listener runs in the repo's own context regardless of what triggered the original run -- no fork code is ever checked out or executed here) and calls `ai-diagnose-failure.py`.
5. Publishes the result as a `neutral` (non-blocking, informational-only) Check Run on the PR's head SHA.

## Test plan
- [x] `python3 -c "import yaml; yaml.safe_load(...)"` on the new workflow
- [ ] **Genuinely new ground, needs live validation**: confirm the default `GITHUB_TOKEN` with `checks: write` can actually create a Check Run against a fork PR's head SHA from a `workflow_run` context -- no existing precedent for `check-runs` API usage in this repo. Will validate against the next real fork PR that fails E2E.
- [ ] Confirm the Check Run appears in the PR's Checks list, renders correctly, and is NOT required/blocking.

## Summary by CodeRabbit

- **New Features**
  - Added automated AI-powered diagnosis for failed end-to-end test runs.
  - Reports now consider changed files, pull-request differences, test results, and nested logs for more relevant findings.
  - Findings are published as concise workflow summaries and commit statuses.
  - Added secure Vertex AI authentication for supported CI workflows without static credentials.
  - Added persistent pull-request comments with per-suite diagnostic updates.

- **Bug Fixes**
  - Improved reporting for fork-based pull requests by publishing results on the relevant commit.
  - Added recovery updates when previously failing test suites pass.
  - Added graceful handling for missing artifacts, malformed or oversized test results, and unavailable diagnostic context.


eliorerz:
Pushed a fix on top of this branch after live-validating against a real fork PR in osac-test-infra:

1. **Check Runs never appear on fork PRs.** Confirmed empirically: a Check Run posted against a fork PR's head SHA is queryable via the API but invisible in the PR's own Checks-tab, because GitHub can only link a check_suite to a PR when the head branch lives in the same repo -- never true for a genuine fork. Switched to the commit Status API instead (no such linkage requirement, verified live via `statusCheckRollup`).
2. **Wrong-CI-system hallucination.** The prompt never said "GitHub Actions," so on sparse evidence Gemini guessed Jenkins/GitLab CI/Tekton. Fixed by naming the workflow and pointing it at the real run URL.

Full writeup: osac-test-infra#446 (same fix, that repo).

eliorerz:
Added three more fixes on top of this branch after the user noticed real diagnoses looked like the model "didn't know the reason":

1. **Recursive log scan fix.** `extract_log_signal()` only scanned the top level of the artifact directory, but most of the real signal (AAP job stdout in `aap-jobs/`, operator logs in `osac-operators/`, etc.) is nested under subdirectories `gather-osac-logs.sh` creates -- silently never read. Now recursive, labeled by relative path.
2. **OSAC architecture primer** added to the prompt, mapping component names to those artifact paths, so the model can actually use the labels. (Checked `osac-ai-skills` -- nothing reusable there, it's dev-workflow guides not failure-diagnosis knowledge.)
3. **PR changed-files + bounded diff** added as extra context (Phase 2 already resolves the PR number). Both JSON-encoded rather than newline/heredoc-joined, since PR content is attacker-controlled -- avoids delimiter-collision output injection.

Full writeup: osac-test-infra#448 (same fix, that repo).

eliorerz:
Added a sticky, edited-in-place PR comment on top of this branch (per your request), mirroring osac-test-infra#449:

- One persistent comment per PR (found via a hidden HTML marker, PATCHed not reposted), one section per E2E suite.
- On failure: create-or-update that suite's section with the **full** diagnosis (no ~140-char cap like the commit status).
- On a later pass: if the comment exists, edit that suite's section to a "now passing" message. If it never existed, post nothing -- never comment just to say things are fine.
- Fixed a real race: two suites failing near-simultaneously would otherwise run their comment-edit steps concurrently and clobber each other. `concurrency.group` is now keyed by `head_sha` (was `workflow_run.id`) so they queue instead.

Full writeup: osac-test-infra#449 (same fix, that repo).

eliorerz:
Same fixes applied here -- see osac-test-infra#449 for the full writeup (4 review findings fixed: embedded-delimiter truncation bug + regression test, job split for least-privilege, PR-stable concurrency key, comment-author impersonation check) plus the switch to a dedicated `AI_DIAGNOSTIC_TOKEN` bot PAT instead of `GITHUB_TOKEN`. Still pending: setting `AI_DIAGNOSTIC_TOKEN` as a repo secret here and in osac-test-infra (token is stored in Vault at `secret/osac/e2e/osac-ai-diagnostic-bot`).

eliorerz:
Added a commit removing the duplication -- this repo's own `ai-diagnostic-e2e.yml` is now a ~15-line wrapper calling osac-test-infra's copy via `workflow_call`, instead of a full duplicated pipeline. Full writeup + merge-order note: osac-test-infra#450 (must merge first, before this PR).

PR #698, opened 2026-09-01
OSAC-4759: Prevent deletion of user suppliedsecrets referenced by other objects

DakCrowder:
Prevents user supplied secrets (like a cluster pull secret) from being deleted while a cluster (or other object) still references the secret.

Does not apply to system owned/created secrets like HUB secrets or break glass creds.

Follows patterns introduced by other resources / references - https://github.com/osac-project/osac/pull/157/changes#diff-63efb27c99e3b251954e45250bf64a558897eed81ac99580ebd9b181e5d5e6e2

## Summary

- **Database:** Added indexes and a trigger that blocks deletion of user-supplied secrets referenced by active objects.
- **Protection rules:** The trigger covers clusters, cluster templates, hubs, identity providers, and storage backends. It excludes HUB-backed and system-managed secrets.
- **API behavior:** Deletion of an in-use Vault secret returns `FailedPrecondition` with an “in use” message.
- **Authorization:** Tests verify platform administrator access to shared Vault secret data and tenant restrictions.
- **Tests:** Added migration and server tests for active references, exempt secrets, soft-deleted dependents, error code `Z0003`, and shared-secret authorization.
- **Maintenance:** Updated the database migration checksum.

## Backward compatibility

This changes deletion behavior. Clients can no longer delete user-supplied secrets while active objects reference them. System-managed, HUB-backed, unreferenced, and otherwise exempt secrets retain their previous deletion behavior.

## Risk classification

**risk:show** — Applied because the change introduces a user-visible deletion restriction and database trigger behavior, but limits the effect to referenced user-supplied secrets and includes focused migration and API tests.

This is below **risk:ask** because the change does not alter exported entities, deployment configuration, or access for system-managed secrets.


jhernand (approved):

danielerez:
Looks good. 
just maybe worth handling the secret deletion issue that coderabbit mentioned? https://github.com/osac-project/osac/pull/698#discussion_r3919461459
or as a follow-up PR...

DakCrowder:
> Looks good. just maybe worth handling the secret deletion issue that coderabbit mentioned? [#698 (comment)](https://github.com/osac-project/osac/pull/698#discussion_r3919461459) or as a follow-up PR...

Added some handling/test case for what was mentioned.  I believe this would affect a number of other migrations (since this basically copied the style) so will get a bug raised there too.

PR #705, opened 2026-09-02
OSAC-4704: Enable OpenBao by default

obochan-rh:
Enable bundled OpenBao (bundledVault) throughout osac-installer so every CI and dev environment gets a Vault-compatible secret store without per-profile opt-in.

P0: Remove bundledVault blocks from CI infra.yaml files (caas-ci, vmaas-ci, bmaas-ci, full-ci) -- these were misplaced because the bundled-openbao.yaml template belongs to the Phase 3 osac chart, but infra.yaml is only passed to osac-deps and osac-infra charts.

P1: Flip defaults in charts/osac/values.yaml (enabled: true, devRootToken: "dev-root-token") and update values.schema.json to match.

P2: Update helm-deployment-guide.md CI/Dev features table to reflect the new default, and add bundled-openbao to AGENTS.md Phase 3 template listing.

P3: Retained bundledVault in dev/kind-instance.yaml for self-documentation (matches the pattern of other explicit overrides in that file).

Assisted-by: Claude <noreply@anthropic.com>

## Summary

* **Deployment and configuration**
  * Bundled OpenBao is enabled by default.
  * The default development root token is `"dev-root-token"`.
  * CI profiles no longer override the chart defaults.
  * The development Kind profile retains explicit bundled OpenBao settings for self-documentation.
  * Bundled PostgreSQL settings remain unchanged.

* **Documentation**
  * Helm deployment guidance now documents the default CI/development service settings.
  * The guide states that these settings should be disabled in production.
  * `AGENTS.md` now lists the `bundled-openbao` template in the chart architecture.

* **API surface, controllers, database, auth, and tests**
  * No API, controller, database, or test changes were identified.
  * OpenBao provides a Vault-compatible secret store for CI and development environments.
  * Production deployments must disable bundled OpenBao and use an external Vault instance.

* **Backward compatibility**
  * CI and development deployments that relied on profile-specific OpenBao configuration now use the chart defaults.
  * Production deployments should explicitly disable bundled OpenBao. Otherwise, the new default changes deployment behavior and uses a development token.

## Risk classification

* **risk:show** — The change affects deployment defaults and secret-store configuration, but it does not modify application logic, APIs, controllers, or database behavior.
* The change is close to **risk:ask** because enabling a bundled secret store and setting a default token can affect production security. It does not qualify because the documentation identifies these settings as CI/development features and instructs production deployments to disable bundled OpenBao.

PR #730, opened 2026-09-03
OSAC-4704: Fix OpenBao OpenShift SCC, namespace endpoint mismatch, and Keycloak issuer docs

obochan-rh:
## Summary

Follow-up to PR #705 ([OSAC-4704](https://redhat.atlassian.net/browse/OSAC-4704) enable OpenBao by default).
Fixes three runtime failures discovered once OpenBao actually deploys (OSAC-4643, [OSAC-4795](https://redhat.atlassian.net/browse/OSAC-4795)).

---

## Fix 1: OpenShift SCC incompatibility (bundled-openbao.yaml)

OpenShift's `restricted-v2` SCC assigns namespace-scoped UIDs (≥ 1000000000). The hardcoded `runAsUser: 100 / runAsGroup: 1000 / fsGroup: 1000` falls outside that range, causing pod admission failure.

- Remove `runAsUser`, `runAsGroup`, `fsGroup` from pod `securityContext`
- `runAsNonRoot: true` is sufficient — OpenShift's assigned UID is always non-root

`bao server -dev` and the bootstrap shell also write lock files to paths outside explicit volume mounts. `readOnlyRootFilesystem: true` causes EROFS at runtime.

- Set `readOnlyRootFilesystem: false` on both `server` and `bootstrap` containers

---

## Fix 2: Vault endpoint namespace mismatch (values.yaml + 4× CI instance.yaml)

All CI profiles hardcoded `service.vault.endpoint: https://openbao.**osac**.svc...` but:
- `caas-ci` installs with `NS=osac-e2e-ci` → actual service is at `openbao.osac-e2e-ci.svc...`
- `standalone` and other non-default namespaces have the same problem

The fulfillment-service deployment template already applies `tpl` to the endpoint value, so Go template syntax works in it.

- Default `service.vault.endpoint` to `https://openbao.{{ .Release.Namespace }}.svc.cluster.local:8200`
- Remove the hardcoded override from all 4 CI `instance.yaml` profiles

---

## Fix 3: Keycloak OIDC issuer mismatch (values.yaml comment)

On OpenShift, Keycloak's `KC_HOSTNAME` is the external route URL. The OIDC discovery document returns this external URL as the `issuer` field. OpenBao validates that `oidc_discovery_url == issuer_from_discovery_doc`; using the internal service URL (`keycloak.keycloak.svc...`) for `keycloakIssuerUrl` causes this check to fail.

Cannot be fixed by a default — the external route URL is deployment-specific. Added prominent comment in `values.yaml` to document the requirement.

---

## Testing

Validated against `192.168.56.19` standalone deployment — all 4 tenants reached `SYNCED`, controller logs show vault namespace provisioned successfully.

Fixes: OSAC-4643, [OSAC-4795](https://redhat.atlassian.net/browse/OSAC-4795)
Companion: #705 (should be merged first or together)

## Deployment

- Removed fixed OpenBao UID, GID, and filesystem group settings.
- Added Kubernetes and OpenShift pod security context overrides.
- Added writable `home` volumes and set `HOME=/tmp` for OpenBao containers.
- Updated CI deployment values for OpenShift `restricted-v2` SCC compatibility.

## Configuration

- Changed the default Vault endpoint to use the Helm release namespace.
- Removed hardcoded Vault endpoints from the BMaaS, CaaS, full, and VMaaS CI profiles.
- Documented that an empty Vault endpoint disables Vault integration.
- Documented that `keycloakIssuerUrl` must match the deployment-specific external Keycloak OIDC issuer URL.
- Enabled bundled OpenBao by default with a development root token.

## CI

- Configured all four CI profiles to use platform-assigned OpenBao identities.
- Removed duplicate bundled OpenBao settings from CI infrastructure values.

## Tests

- Standalone deployment testing confirmed that all four tenants reached `SYNCED`.
- Vault namespaces were provisioned successfully.

## Backward compatibility

- Deployments that use the hardcoded `osac` namespace endpoint must use the chart-derived endpoint or provide an explicit override.
- Deployments with an incorrect `keycloakIssuerUrl` must update it to the external Keycloak OIDC issuer URL.
- OpenBao deployments may now use platform-assigned identities instead of fixed numeric IDs.
- The bundled OpenBao default and development root token change the default deployment configuration. Production deployments should provide secure credentials and suitable security context settings.

## Risk classification

- **risk:ship** — The changes are limited to deployment configuration, CI values, security context compatibility, endpoint defaults, and documentation. They do not change application APIs, controllers, databases, or authentication flows. Standalone deployment testing validated tenant synchronization and Vault provisioning.
- The PR is not close to `risk:show` because it does not add a user-visible feature or broad application behavior change. It is not close to `risk:ask` because deployment testing validated the runtime configuration changes.

### DIFF
--- a/osac-installer/charts/osac/templates/bundled-openbao.yaml
+++ b/osac-installer/charts/osac/templates/bundled-openbao.yaml
@@ -169,10 +169,19 @@ spec:
     spec:
       automountServiceAccountToken: false
       securityContext:
+        {{- if .Values.bundledVault.podSecurityContext }}
+        {{- toYaml .Values.bundledVault.podSecurityContext | nindent 8 }}
+        {{- else }}
+        # Standard Kubernetes default: explicit numeric UID required because the
+        # OpenBao image uses a named user ("openbao") — Kubernetes cannot satisfy
+        # runAsNonRoot: true without a numeric UID at admission time.
+        # On OpenShift, override podSecurityContext in the profile's instance.yaml
+        # to just {runAsNonRoot: true}; the restricted-v2 SCC allocates the UID.
         runAsNonRoot: true
         runAsUser: 100
         runAsGroup: 1000
         fsGroup: 1000
+        {{- end }}
       volumes:
       - name: server-cert
         secret:
@@ -193,6 +202,14 @@ spec:
       - name: data
         emptyDir: {}
       - name: tmp
+        emptyDir: {}
+      # home-openbao: writable home dir for the openbao user.
+      # The Alpine image sets HOME to / (no explicit home dir) while the UBI
+      # image uses /home/openbao. We map both by setting HOME=/tmp in each
+      # container's env and mounting this emptyDir there, so any HOME-relative
+      # writes (e.g. ~/.vault-token from bao CLI) land in an emptyDir rather
+      # than the read-only root filesystem.
+      - name: home
         emptyDir: {}
       containers:
       - name: server
@@ -225,6 +242,11 @@ spec:
           {{- end }}
         - name: SKIP_SETCAP
           value: "true"
+        # HOME=/tmp: the Alpine image has no explicit home dir; redirecting HOME
+        # to the already-mounted /tmp emptyDir prevents any HOME-relative writes
+        # from hitting the read-only root filesystem.
+        - name: HOME
+          value: "/tmp"
         volumeMounts:
         - name: server-cert
           mountPath: /vault/tls
@@ -238,6 +260,8 @@ spec:
           mountPath: /vault/data
         - name: tmp
           mountPath: /tmp
+        - name: home
+          mountPath: /home/openbao
         ports:
         - name: api
           protocol: TCP
@@ -283,6 +307,10 @@ spec:
           value: {{ .Values.service.vault.keycloakIssuerUrl | default "" | quote }}
         - name: KEYCLOAK_AUDIENCE
           value: {{ .Values.service.vault.keycloakAudience | default "" | quote }}
+        # HOME=/tmp: ensures HOME-relative writes (e.g. bao CLI token cache) land
+        # on the mounted /tmp emptyDir rather than the read-only root filesystem.
+        - name: HOME
+          value: "/tmp"
         volumeMounts:
         - name: init-script
           mountPath: /scripts
@@ -294,6 +322,8 @@ spec:
         {{- end }}
         - name: tmp
           mountPath: /tmp
+        - name: home
+          mountPath: /home/openbao
         resources:
           requests:
             cpu: 50m
--- a/osac-installer/charts/osac/templates/bundled-openbao.yaml
+++ b/osac-installer/charts/osac/templates/bundled-openbao.yaml
@@ -304,7 +304,7 @@ spec:
           value: "/vault/tls/ca.crt"
           {{- end }}
         - name: KEYCLOAK_ISSUER_URL
-          value: {{ .Values.service.vault.keycloakIssuerUrl | default "" | quote }}
+          value: {{ tpl (.Values.service.vault.keycloakIssuerUrl | default "") . | quote }}
         - name: KEYCLOAK_AUDIENCE
           value: {{ .Values.service.vault.keycloakAudience | default "" | quote }}
         # HOME=/tmp: ensures HOME-relative writes (e.g. bao CLI token cache) land
--- a/fulfillment-service/charts/service/templates/controller/deployment.yaml
+++ b/fulfillment-service/charts/service/templates/controller/deployment.yaml
@@ -12,9 +12,9 @@ specific language governing permissions and limitations under the License.
 */}}
 
 {{- $caBundleConfigMap := required "certs.caBundle.configMap is required" .Values.certs.caBundle.configMap }}
-{{- $authIssuerUrl := required "auth.issuerUrl is required" .Values.auth.issuerUrl }}
+{{- $authIssuerUrl := tpl (required "auth.issuerUrl is required" .Values.auth.issuerUrl) . }}
 {{- $authControllerCredentials := required "auth.controllerCredentials is required" .Values.auth.controllerCredentials }}
-{{- $idpUrl := required "idp.url is required" .Values.idp.url }}
+{{- $idpUrl := tpl (required "idp.url is required" .Values.idp.url) . }}
 {{- $idpCredentials := required "idp.credentials is required" .Values.idp.credentials }}
 
 apiVersion: apps/v1
@@ -176,14 +176,14 @@ spec:
         - --grpc-listener-tls-key=/etc/fulfillment-controller/tls/tls.key
         - --metrics-listener-address=0.0.0.0:8002
         {{- if and .Values.vault.endpoint .Values.vault.credentials }}
-        - --vault-endpoint={{ .Values.vault.endpoint }}
+        - --vault-endpoint={{ tpl .Values.vault.endpoint . }}
         - --vault-namespace={{ .Values.vault.namespace }}
         - --vault-kv-mount-path={{ .Values.vault.kvMountPath }}
         - --vault-lifecycle-role={{ .Values.vault.lifecycleRole }}
         - --vault-lifecycle-mount-path={{ .Values.vault.lifecycleMountPath }}
         - --vault-keycloak-client-id={{ .Values.vault.keycloakClientId }}
         - --vault-keycloak-client-secret-file=/etc/fulfillment-controller/vault/client-secret
-        - --vault-keycloak-issuer-url={{ .Values.vault.keycloakIssuerUrl }}
+        - --vault-keycloak-issuer-url={{ tpl .Values.vault.keycloakIssuerUrl . }}
         - --vault-keycloak-audience={{ .Values.vault.keycloakAudience }}
         {{- if .Values.vault.caBundle.configMap }}
         - --vault-ca-cert-file=/etc/fulfillment-controller/vault-cas/bundle.pem
--- a/fulfillment-service/charts/service/templates/grpc-server/deployment.yaml
+++ b/fulfillment-service/charts/service/templates/grpc-server/deployment.yaml
@@ -165,7 +165,7 @@ spec:
         - --grpc-listener-address=0.0.0.0:8000
         - --grpc-listener-tls-crt=/etc/fulfillment-grpc-server/tls/tls.crt
         - --grpc-listener-tls-key=/etc/fulfillment-grpc-server/tls/tls.key
-        - --grpc-authn-trusted-token-issuers={{ required "auth.issuerUrl is required" .Values.auth.issuerUrl }}
+        - --grpc-authn-trusted-token-issuers={{ tpl (required "auth.issuerUrl is required" .Values.auth.issuerUrl) . }}
         - --emergency-service-accounts={{ join "," .Values.auth.emergencyServiceAccounts }}
         - --token-signer-crt=/etc/fulfillment-token-signer/tls.crt
         - --token-signer-key=/etc/fulfillment-token-signer/tls.key
@@ -173,13 +173,13 @@ spec:
         - --token-issuer={{ include "fulfillment-api.tokenIssuerUrl" . }}
         - --metrics-listener-address=0.0.0.0:8002
         {{- if and .Values.vault.endpoint .Values.vault.credentials }}
-        - --vault-endpoint={{ .Values.vault.endpoint }}
+        - --vault-endpoint={{ tpl .Values.vault.endpoint . }}
         - --vault-namespace={{ .Values.vault.namespace }}
         - --vault-kv-mount-path={{ .Values.vault.kvMountPath }}
         {{- if .Values.vault.caBundle.configMap }}
         - --vault-ca-cert-file=/etc/fulfillment-grpc-server/vault-cas/bundle.pem
         {{- end }}
-        - --vault-keycloak-issuer-url={{ .Values.vault.keycloakIssuerUrl }}
+        - --vault-keycloak-issuer-url={{ tpl .Values.vault.keycloakIssuerUrl . }}
         - --vault-keycloak-client-id={{ .Values.vault.keycloakClientId }}
         - --vault-keycloak-client-secret-file=/etc/fulfillment-grpc-server/vault/client-secret
         {{- end }}
--- a/osac-installer/Makefile
+++ b/osac-installer/Makefile
@@ -198,7 +198,7 @@ endif
 	helm upgrade --install osac $(OSAC_CHART) \
 		-f $(INSTANCE_VALUES) \
 		--namespace $(NS) --create-namespace \
-		$(if $(DOMAIN),--set service.externalHostname=fulfillment-api-$(NS).$(DOMAIN) --set service.internalHostname=fulfillment-internal-api-$(NS).$(DOMAIN) --set service.auth.issuerUrl=https://keycloak-keycloak.$(DOMAIN)/realms/osac --set service.idp.url=https://keycloak-keycloak.$(DOMAIN)) \
+		$(if $(DOMAIN),--set service.externalHostname=fulfillment-api-$(NS).$(DOMAIN) --set service.internalHostname=fulfillment-internal-api-$(NS).$(DOMAIN) --set global.clusterDomain=$(DOMAIN)) \
 		$(EXTRA_HELM_ARGS) \
 		--wait --timeout 40m
 
--- a/osac-installer/charts/osac/values.schema.json
+++ b/osac-installer/charts/osac/values.schema.json
@@ -394,7 +394,7 @@
           "properties": {
             "endpoint": {
               "type": "string",
-              "description": "Vault API endpoint URL (e.g. http://openbao.openbao.svc.cluster.local:8200)"
+              "description": "Vault API endpoint URL. Supports Helm tpl so {{ .Release.Namespace }} expands to the install namespace (e.g. https://openbao.{{ .Release.Namespace }}.svc.cluster.local:8200). Leave empty to disable Vault integration."
             },
             "namespace": {
               "type": "string",
@@ -1485,17 +1485,28 @@
       "properties": {
         "enabled": {
           "type": "boolean",
-          "description": "Deploy a bundled OpenBao instance for development/CI use",
-          "default": false
+          "description": "Deploy a bundled OpenBao instance. Enabled by default — set to false only when pointing at an external Vault/OpenBao instance.",
+          "default": true
         },
         "devRootToken": {
           "type": "string",
-          "description": "Root token for OpenBao dev mode (required when enabled)"
+          "description": "Root token for OpenBao dev mode (required when enabled)",
+          "default": "dev-root-token"
         },
         "image": {
           "type": "string",
           "description": "OpenBao container image (must be 2.5+ for namespace support)",
           "default": "ghcr.io/openbao/openbao:2.6.1"
+        },
+        "podSecurityContext": {
+          "type": "object",
+          "description": "Pod-level securityContext for the OpenBao deployment. Leave absent (nil) to use the template's built-in standard-Kubernetes defaults (runAsNonRoot: true, runAsUser: 100, runAsGroup: 1000, fsGroup: 1000). Any non-nil value here FULLY REPLACES the template default — partial overrides must include every key you want rendered. On OpenShift with restricted-v2 SCC, set to {runAsNonRoot: true} only.",
+          "properties": {
+            "runAsNonRoot": { "type": "boolean" },
+            "runAsUser": { "type": "integer" },
+            "runAsGroup": { "type": "integer" },
+            "fsGroup": { "type": "integer" }
+          }
         },
         "resources": {
           "type": "object",
--- a/osac-installer/charts/osac/values.schema.json
+++ b/osac-installer/charts/osac/values.schema.json
@@ -4,6 +4,16 @@
   "description": "Configuration values for Open Sovereign AI Cloud deployment",
   "type": "object",
   "properties": {
+    "global": {
+      "type": "object",
+      "description": "Values propagated unchanged into every subchart, shared between the umbrella chart's own templates and its subcharts",
+      "properties": {
+        "clusterDomain": {
+          "type": "string",
+          "description": "OpenShift cluster ingress domain (e.g. apps.mycluster.example.com). Backs the tpl-templated defaults for service.auth.issuerUrl, service.idp.url, and service.vault.keycloakIssuerUrl. Set via --set global.clusterDomain=$(oc get ingresses.config/cluster -o jsonpath='{.spec.domain}'). Leave empty on platforms (e.g. kind) that set those three values explicitly instead."
+        }
+      }
+    },
     "cliImage": {
       "type": "string",
       "description": "Default OpenShift CLI image used by hook jobs"
@@ -421,7 +431,7 @@
             },
             "keycloakIssuerUrl": {
               "type": "string",
-              "description": "Keycloak OIDC issuer URL for configuring JWT auth in tenant Vault namespaces"
+              "description": "Keycloak OIDC issuer URL for configuring JWT auth in tenant Vault namespaces. Supports Helm tpl -- defaults to a global.clusterDomain-backed external route URL matching Keycloak's actual OIDC issuer on OpenShift. Must match the issuer (iss claim) Keycloak puts in its JWT tokens, or OIDC discovery validation fails."
             },
             "keycloakAudience": {
               "type": "string",
--- a/osac-installer/charts/osac/values.yaml
+++ b/osac-installer/charts/osac/values.yaml
@@ -1,6 +1,19 @@
 # Open Sovereign AI Cloud - Unified Helm Values
 # This values file configures all OSAC components via the umbrella chart.
 
+# global.* is propagated unchanged into every subchart's own .Values, so it's
+# the only way to share a single value (like the cluster's ingress domain)
+# between the umbrella chart's own templates and its subcharts without
+# duplicating it as separate --set flags per value.
+global:
+  # [REQUIRED on OpenShift, via --set global.clusterDomain=$(oc get ingresses.config/cluster -o jsonpath='{.spec.domain}')]
+  # Backs the tpl-templated defaults below (service.auth.issuerUrl,
+  # service.idp.url, service.vault.keycloakIssuerUrl) so they automatically
+  # match Keycloak's actual external route hostname (KC_HOSTNAME) on any
+  # OpenShift cluster. Leave empty for platforms (e.g. kind) that set these
+  # three values explicitly instead.
+  clusterDomain: ""
+
 cliImage: quay.io/openshift/origin-cli:4.20.0
 
 # Shared PostgreSQL host for the db-init hook.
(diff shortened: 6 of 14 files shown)


osac-installer/charts/osac/templates/bundled-openbao.yaml
```diff
@@ -200,7 +202,9 @@ spec:
         imagePullPolicy: IfNotPresent
         securityContext:
           allowPrivilegeEscalation: false
-          readOnlyRootFilesystem: true
+          # readOnlyRootFilesystem disabled: bao server -dev writes lock files
+          # and internal state to paths not covered by explicit volume mounts.
+          readOnlyRootFilesystem: false
```
danielerez: maybe we can use emptyDir volumes instead of disabling read-only root entirely?
I mean, add targeted emptyDir volume mounts for the specific paths bao writes to (as /home/openbao, /tmp, etc). that way the container can only write to explicitly allowed paths, not the entire filesystem...

danielerez:
Looks good, just one issue if you want to tackle as part of this PR: https://github.com/osac-project/osac/pull/730#discussion_r3922684074

danielerez:
Looks good.
just need to fix pre-commit: https://github.com/osac-project/osac/actions/runs/33739287084/job/100660617633?pr=730
```
  fix end of files.........................................................Failed
```

ygalblum:
## CI failure confirms the OIDC issuer fix doesn't actually work

The CaaS E2E run's AI diagnosis confirms the exact failure mode I want to flag here:

> **Root cause**: The OpenBao (Vault) instance failed to initialize due to an OIDC issuer URL mismatch when attempting to connect to Keycloak, caused by configuring the internal service URL instead of the external route URL for Keycloak on OpenShift.
>
> `pod-openbao-...-server.log`:
> ```
> [ERROR] auth.jwt.auth_jwt_...: error checking oidc discovery URL: error="issuer did not match the returned issuer, expected \"https://keycloak.keycloak.svc.cluster.local/realms/osac\" got \"https://keycloak-keycloak.apps.test-infra-cluster-c5118747-caassanity.redhat.com/realms/osac\""

([full run](https://github.com/osac-project/osac/actions/runs/33895118747))

This is exactly the bug **Fix 3** in this PR's description says it's addressing — but Fix 3 only adds a documentation comment to `values.yaml`; it doesn't actually change `keycloakIssuerUrl`'s value, so nothing here fixes the mismatch. CI confirms it's still broken.

## Fix 2 has the same problem, for a different reason

**Fix 2**'s description says:

> The fulfillment-service deployment template already applies `tpl` to the endpoint value, so Go template syntax works in it.

I checked — this isn't true. There's no `tpl` call anywhere in the `fulfillment-service` chart. I confirmed by rendering this branch directly:

```
$ helm template osac charts/osac -f values/vmaas-ci/instance.yaml ... --namespace osac-e2e-ci | grep vault-endpoint
        - --vault-endpoint=https://openbao.{{ .Release.Namespace }}.svc.cluster.local:8200
```

`{{ .Release.Namespace }}` renders **literally** — Helm only auto-templates template *files*, not Go-template syntax embedded inside a plain string value from `values.yaml`. Without an explicit `tpl` call at the point of use, this produces an invalid hostname containing literal `{{ }}`, which would fail DNS resolution just like the original bug (this may not have surfaced in the CaaS E2E run above since OpenBao's JWT/issuer bootstrap failed first and blocked before the controller got far enough to hit vault-endpoint's own DNS lookup).

## A tested, working fix for both

I added the missing `tpl` calls and verified them by rendering. One gotcha worth calling out: `helm dependency build` (`make helm-deps`) must be re-run after editing `fulfillment-service`'s chart templates, since `charts/osac/charts/fulfillment-service-*.tgz` is a vendored snapshot — edits to the source chart don't take effect until it's rebuilt.

For Fix 3 (Keycloak issuer), a `tpl`-only fix using `.Release.Namespace` (like Fix 2) isn't quite enough on its own, since the external cluster domain isn't intrinsically known to Helm the way `.Release.Namespace` is — it needs a value threaded in from outside (same way `install-infra` already does for `keycloak.hostname`). I added a `global.clusterDomain` value, since Helm's `global.*` values propagate identically into every subchart's own `.Values` — this sidesteps a real scope-crossing headache: `keycloakIssuerUrl` needs to render correctly both from the umbrella chart's own `bundled-openbao.yaml` template *and* from the `fulfillment-service` subchart's deployment templates, and those two contexts address the same underlying value at different relative paths (`.Values.service.idp.url` vs `.Values.idp.url`). A `global.*` value is the only thing that resolves identically in both places without extra plumbing.

With this, `install-osac`'s three separate `--set service.auth.issuerUrl=... --set service.idp.url=... ` flags collapse into a single `--set global.clusterDomain=$(DOMAIN)`, and all 4 CI profiles' now-stale hardcoded `issuerUrl`/`url`/`keycloakIssuerUrl` overrides can be removed entirely (the chart's own tpl'd defaults take over correctly).

Verified end-to-end via `helm template` across all 4 CI profiles (each resolves correctly with just `--set global.clusterDomain=...`), the bundled OpenBao's own `KEYCLOAK_ISSUER_URL` env var, and confirmed the `kind`/`kind-instance-devfull` profiles are unaffected (their own static overrides still take precedence, unchanged). `yamllint --strict` and `pre-commit run --all-files` both pass on top of this branch; `make helm-lint`/`make helm-validate` fail only on the pre-existing, unrelated `service.externalHostname`/`internalHostname` schema error already present on `main`.

<details>
<summary>Full diff (relative to this PR's branch)</summary>

```diff
diff --git a/fulfillment-service/charts/service/templates/controller/deployment.yaml b/fulfillment-service/charts/service/templates/controller/deployment.yaml
index 050ee110b..05c395bc5 100644
--- a/fulfillment-service/charts/service/templates/controller/deployment.yaml
+++ b/fulfillment-service/charts/service/templates/controller/deployment.yaml
@@ -12,9 +12,9 @@ specific language governing permissions and limitations under the License.
 */}}
 
 {{- $caBundleConfigMap := required "certs.caBundle.configMap is required" .Values.certs.caBundle.configMap }}
-{{- $authIssuerUrl := required "auth.issuerUrl is required" .Values.auth.issuerUrl }}
+{{- $authIssuerUrl := tpl (required "auth.issuerUrl is required" .Values.auth.issuerUrl) . }}
 {{- $authControllerCredentials := required "auth.controllerCredentials is required" .Values.auth.controllerCredentials }}
-{{- $idpUrl := required "idp.url is required" .Values.idp.url }}
+{{- $idpUrl := tpl (required "idp.url is required" .Values.idp.url) . }}
 {{- $idpCredentials := required "idp.credentials is required" .Values.idp.credentials }}
 
 apiVersion: apps/v1
@@ -176,14 +176,14 @@ spec:
         - --grpc-listener-tls-key=/etc/fulfillment-controller/tls/tls.key
         - --metrics-listener-address=0.0.0.0:8002
         {{- if and .Values.vault.endpoint .Values.vault.credentials }}
-        - --vault-endpoint={{ .Values.vault.endpoint }}
+        - --vault-endpoint={{ tpl .Values.vault.endpoint . }}
         - --vault-namespace={{ .Values.vault.namespace }}
         - --vault-kv-mount-path={{ .Values.vault.kvMountPath }}
         - --vault-lifecycle-role={{ .Values.vault.lifecycleRole }}
         - --vault-lifecycle-mount-path={{ .Values.vault.lifecycleMountPath }}
         - --vault-keycloak-client-id={{ .Values.vault.keycloakClientId }}
         - --vault-keycloak-client-secret-file=/etc/fulfillment-controller/vault/client-secret
-        - --vault-keycloak-issuer-url={{ .Values.vault.keycloakIssuerUrl }}
+        - --vault-keycloak-issuer-url={{ tpl .Values.vault.keycloakIssuerUrl . }}
         - --vault-keycloak-audience={{ .Values.vault.keycloakAudience }}
         {{- if .Values.vault.caBundle.configMap }}
         - --vault-ca-cert-file=/etc/fulfillment-controller/vault-cas/bundle.pem
diff --git a/fulfillment-service/charts/service/templates/grpc-server/deployment.yaml b/fulfillment-service/charts/service/templates/grpc-server/deployment.yaml
index 2b46dc4de..ebfe231e7 100644
--- a/fulfillment-service/charts/service/templates/grpc-server/deployment.yaml
+++ b/fulfillment-service/charts/service/templates/grpc-server/deployment.yaml
@@ -165,7 +165,7 @@ spec:
         - --grpc-listener-address=0.0.0.0:8000
         - --grpc-listener-tls-crt=/etc/fulfillment-grpc-server/tls/tls.crt
         - --grpc-listener-tls-key=/etc/fulfillment-grpc-server/tls/tls.key
-        - --grpc-authn-trusted-token-issuers={{ required "auth.issuerUrl is required" .Values.auth.issuerUrl }}
+        - --grpc-authn-trusted-token-issuers={{ tpl (required "auth.issuerUrl is required" .Values.auth.issuerUrl) . }}
         - --emergency-service-accounts={{ join "," .Values.auth.emergencyServiceAccounts }}
         - --token-signer-crt=/etc/fulfillment-token-signer/tls.crt
         - --token-signer-key=/etc/fulfillment-token-signer/tls.key
@@ -173,13 +173,13 @@ spec:
         - --token-issuer={{ include "fulfillment-api.tokenIssuerUrl" . }}
         - --metrics-listener-address=0.0.0.0:8002
         {{- if and .Values.vault.endpoint .Values.vault.credentials }}
-        - --vault-endpoint={{ .Values.vault.endpoint }}
+        - --vault-endpoint={{ tpl .Values.vault.endpoint . }}
         - --vault-namespace={{ .Values.vault.namespace }}
         - --vault-kv-mount-path={{ .Values.vault.kvMountPath }}
         {{- if .Values.vault.caBundle.configMap }}
         - --vault-ca-cert-file=/etc/fulfillment-grpc-server/vault-cas/bundle.pem
         {{- end }}
-        - --vault-keycloak-issuer-url={{ .Values.vault.keycloakIssuerUrl }}
+        - --vault-keycloak-issuer-url={{ tpl .Values.vault.keycloakIssuerUrl . }}
         - --vault-keycloak-client-id={{ .Values.vault.keycloakClientId }}
         - --vault-keycloak-client-secret-file=/etc/fulfillment-grpc-server/vault/client-secret
         {{- end }}
diff --git a/osac-installer/Makefile b/osac-installer/Makefile
index 30045f1f9..4941ccccd 100644
--- a/osac-installer/Makefile
+++ b/osac-installer/Makefile
@@ -198,7 +198,7 @@ endif
 	helm upgrade --install osac $(OSAC_CHART) \
 		-f $(INSTANCE_VALUES) \
 		--namespace $(NS) --create-namespace \
-		$(if $(DOMAIN),--set service.externalHostname=fulfillment-api-$(NS).$(DOMAIN) --set service.internalHostname=fulfillment-internal-api-$(NS).$(DOMAIN) --set service.auth.issuerUrl=https://keycloak-keycloak.$(DOMAIN)/realms/osac --set service.idp.url=https://keycloak-keycloak.$(DOMAIN)) \
+		$(if $(DOMAIN),--set service.externalHostname=fulfillment-api-$(NS).$(DOMAIN) --set service.internalHostname=fulfillment-internal-api-$(NS).$(DOMAIN) --set global.clusterDomain=$(DOMAIN)) \
 		$(EXTRA_HELM_ARGS) \
 		--wait --timeout 40m
 
diff --git a/osac-installer/charts/osac/templates/bundled-openbao.yaml b/osac-installer/charts/osac/templates/bundled-openbao.yaml
index 540f93675..0bdb67ea4 100644
--- a/osac-installer/charts/osac/templates/bundled-openbao.yaml
+++ b/osac-installer/charts/osac/templates/bundled-openbao.yaml
@@ -304,7 +304,7 @@ spec:
           value: "/vault/tls/ca.crt"
           {{- end }}
         - name: KEYCLOAK_ISSUER_URL
-          value: {{ .Values.service.vault.keycloakIssuerUrl | default "" | quote }}
+          value: {{ tpl (.Values.service.vault.keycloakIssuerUrl | default "") . | quote }}
         - name: KEYCLOAK_AUDIENCE
           value: {{ .Values.service.vault.keycloakAudience | default "" | quote }}
         # HOME=/tmp: ensures HOME-relative writes (e.g. bao CLI token cache) land
diff --git a/osac-installer/charts/osac/values.schema.json b/osac-installer/charts/osac/values.schema.json
index d1a980ae7..0fedd2e43 100644
--- a/osac-installer/charts/osac/values.schema.json
+++ b/osac-installer/charts/osac/values.schema.json
@@ -4,6 +4,16 @@
   "description": "Configuration values for Open Sovereign AI Cloud deployment",
   "type": "object",
   "properties": {
+    "global": {
+      "type": "object",
+      "description": "Values propagated unchanged into every subchart, shared between the umbrella chart's own templates and its subcharts",
+      "properties": {
+        "clusterDomain": {
+          "type": "string",
+          "description": "OpenShift cluster ingress domain (e.g. apps.mycluster.example.com). Backs the tpl-templated defaults for service.auth.issuerUrl, service.idp.url, and service.vault.keycloakIssuerUrl. Set via --set global.clusterDomain=$(oc get ingresses.config/cluster -o jsonpath='{.spec.domain}'). Leave empty on platforms (e.g. kind) that set those three values explicitly instead."
+        }
+      }
+    },
     "cliImage": {
       "type": "string",
       "description": "Default OpenShift CLI image used by hook jobs"
@@ -421,7 +431,7 @@
             },
             "keycloakIssuerUrl": {
               "type": "string",
-              "description": "Keycloak OIDC issuer URL for configuring JWT auth in tenant Vault namespaces"
+              "description": "Keycloak OIDC issuer URL for configuring JWT auth in tenant Vault namespaces. Supports Helm tpl -- defaults to a global.clusterDomain-backed external route URL matching Keycloak's actual OIDC issuer on OpenShift. Must match the issuer (iss claim) Keycloak puts in its JWT tokens, or OIDC discovery validation fails."
             },
             "keycloakAudience": {
               "type": "string",
diff --git a/osac-installer/charts/osac/values.yaml b/osac-installer/charts/osac/values.yaml
index 1e6d55922..9cf869c00 100644
--- a/osac-installer/charts/osac/values.yaml
+++ b/osac-installer/charts/osac/values.yaml
@@ -1,6 +1,19 @@
 # Open Sovereign AI Cloud - Unified Helm Values
 # This values file configures all OSAC components via the umbrella chart.
 
+# global.* is propagated unchanged into every subchart's own .Values, so it's
+# the only way to share a single value (like the cluster's ingress domain)
+# between the umbrella chart's own templates and its subcharts without
+# duplicating it as separate --set flags per value.
+global:
+  # [REQUIRED on OpenShift, via --set global.clusterDomain=$(oc get ingresses.config/cluster -o jsonpath='{.spec.domain}')]
+  # Backs the tpl-templated defaults below (service.auth.issuerUrl,
+  # service.idp.url, service.vault.keycloakIssuerUrl) so they automatically
+  # match Keycloak's actual external route hostname (KC_HOSTNAME) on any
+  # OpenShift cluster. Leave empty for platforms (e.g. kind) that set these
+  # three values explicitly instead.
+  clusterDomain: ""
+
 cliImage: quay.io/openshift/origin-cli:4.20.0
 
 # Shared PostgreSQL host for the db-init hook.
@@ -73,11 +86,15 @@ service:
     caBundle:
       configMap: ca-bundle
   auth:
-    issuerUrl: "https://REPLACE_ME"
+    # tpl is applied at render time -- {{ .Values.global.clusterDomain }}
+    # expands via global.clusterDomain. Platforms that don't set
+    # global.clusterDomain (e.g. kind) must override this explicitly instead.
+    issuerUrl: "https://keycloak-keycloak.{{ .Values.global.clusterDomain }}/realms/osac"
     controllerCredentials: []
   idp:
     provider: keycloak
-    url: ""
+    # tpl is applied at render time -- see service.auth.issuerUrl above.
+    url: "https://keycloak-keycloak.{{ .Values.global.clusterDomain }}"
     credentials: []
   database:
     connection: []
@@ -93,25 +110,26 @@ service:
     headers: false
     bodies: false
   vault:
-    # Vault API endpoint. When bundledVault.enabled is true, set this to the
-    # in-cluster OpenBao service URL for the install namespace. The deployment
-    # template applies tpl at render time, so {{ .Release.Namespace }} expands
-    # to the actual namespace — use that pattern in per-profile overrides.
-    # Leave empty to disable Vault integration (no --vault-endpoint flag is
-    # passed to the controller when endpoint is empty, even if credentials
-    # are configured).
-    endpoint: ""
+    # Vault API endpoint. tpl is applied at render time, so
+    # {{ .Release.Namespace }} expands to the actual install namespace --
+    # correct by default for the in-cluster bundled OpenBao service on any
+    # namespace. Leave empty to disable Vault integration entirely (no
+    # --vault-endpoint flag is passed to the controller when endpoint is
+    # empty, even if credentials are configured). Override with a real
+    # external Vault URL when not using the bundled OpenBao.
+    endpoint: "https://openbao.{{ .Release.Namespace }}.svc.cluster.local:8200"
     namespace: osac
     kvMountPath: secret
     lifecycleRole: ""
     lifecycleMountPath: "jwt"
     keycloakClientId: ""
     # Must match the issuer URL (iss claim) Keycloak puts in its JWT tokens.
-    # On OpenShift, Keycloak's KC_HOSTNAME is typically the external route URL
-    # (e.g. https://keycloak-keycloak.apps.<cluster>/realms/osac), NOT the
-    # internal service URL. Using the internal URL causes OIDC discovery issuer
-    # mismatch and vault bootstrap failure. Set this to the external route URL.
-    keycloakIssuerUrl: ""
+    # tpl is applied at render time -- backed by global.clusterDomain, same as
+    # service.auth.issuerUrl above, so this automatically matches Keycloak's
+    # actual external route hostname (KC_HOSTNAME) on any OpenShift cluster.
+    # Platforms that don't set global.clusterDomain (e.g. kind) must override
+    # this explicitly instead.
+    keycloakIssuerUrl: "https://keycloak-keycloak.{{ .Values.global.clusterDomain }}/realms/osac"
     keycloakAudience: "osac-api"
     caBundle:
       configMap: ""
diff --git a/osac-installer/values/bmaas-ci/infra.yaml b/osac-installer/values/bmaas-ci/infra.yaml
index 7354ff8c0..3fefb8fbd 100644
--- a/osac-installer/values/bmaas-ci/infra.yaml
+++ b/osac-installer/values/bmaas-ci/infra.yaml
@@ -34,4 +34,3 @@ bundledPostgres:
   database:
     name: service
     user: service
-
diff --git a/osac-installer/values/bmaas-ci/instance.yaml b/osac-installer/values/bmaas-ci/instance.yaml
index 173c42a91..4cadea92b 100644
--- a/osac-installer/values/bmaas-ci/instance.yaml
+++ b/osac-installer/values/bmaas-ci/instance.yaml
@@ -45,7 +45,9 @@ service:
     caBundle:
       configMap: ca-bundle
   auth:
-    issuerUrl: https://keycloak.keycloak.svc.cluster.local/realms/osac
+    # issuerUrl defaults to a global.clusterDomain-backed external route URL
+    # (see charts/osac/values.yaml) -- correct automatically once
+    # install-osac sets --set global.clusterDomain=$(DOMAIN).
     emergencyServiceAccounts:
     - admin
     - osac-operator
@@ -62,7 +64,9 @@ service:
           param: client-secret
   idp:
     provider: keycloak
-    url: https://keycloak.keycloak.svc.cluster.local
+    # url defaults to a global.clusterDomain-backed external route URL (see
+    # charts/osac/values.yaml) -- correct automatically once install-osac
+    # sets --set global.clusterDomain=$(DOMAIN).
     credentials:
     - secret:
         name: fulfillment-controller-credentials
@@ -90,16 +94,13 @@ service:
   log:
     level: debug
   vault:
-    # tpl is applied to this value at render time — {{ .Release.Namespace }}
-    # expands to the actual install namespace, avoiding DNS failures when
-    # deploying with a non-default NS.
-    endpoint: "https://openbao.{{ .Release.Namespace }}.svc.cluster.local:8200"
+    # endpoint and keycloakIssuerUrl both default correctly (see
+    # charts/osac/values.yaml) -- no override needed.
     namespace: osac
     kvMountPath: secret
     lifecycleRole: lifecycle
     lifecycleMountPath: jwt
     keycloakClientId: osac-controller
-    keycloakIssuerUrl: "https://keycloak.keycloak.svc.cluster.local/realms/osac"
     keycloakAudience: osac-api
     caBundle:
       configMap: ca-bundle
diff --git a/osac-installer/values/caas-ci/infra.yaml b/osac-installer/values/caas-ci/infra.yaml
index 62583cbb3..56f41d649 100644
--- a/osac-installer/values/caas-ci/infra.yaml
+++ b/osac-installer/values/caas-ci/infra.yaml
@@ -40,4 +40,3 @@ bundledPostgres:
   enabled: true
   database:
     name: service
-
diff --git a/osac-installer/values/caas-ci/instance.yaml b/osac-installer/values/caas-ci/instance.yaml
index 20455efa0..dc38e7576 100644
--- a/osac-installer/values/caas-ci/instance.yaml
+++ b/osac-installer/values/caas-ci/instance.yaml
@@ -43,7 +43,9 @@ service:
     caBundle:
       configMap: ca-bundle
   auth:
-    issuerUrl: https://keycloak.keycloak.svc.cluster.local/realms/osac
+    # issuerUrl defaults to a global.clusterDomain-backed external route URL
+    # (see charts/osac/values.yaml) -- correct automatically once
+    # install-osac sets --set global.clusterDomain=$(DOMAIN).
     emergencyServiceAccounts:
     - admin
     - osac-operator
@@ -60,7 +62,9 @@ service:
           param: client-secret
   idp:
     provider: keycloak
-    url: https://keycloak.keycloak.svc.cluster.local
+    # url defaults to a global.clusterDomain-backed external route URL (see
+    # charts/osac/values.yaml) -- correct automatically once install-osac
+    # sets --set global.clusterDomain=$(DOMAIN).
     credentials:
     - secret:
         name: fulfillment-controller-credentials
@@ -88,16 +92,13 @@ service:
   log:
     level: debug
   vault:
-    # tpl is applied to this value at render time — {{ .Release.Namespace }}
-    # expands to the actual install namespace. caas-ci installs with
-    # NS=osac-e2e-ci; the old hardcoded "osac" endpoint caused DNS failures.
-    endpoint: "https://openbao.{{ .Release.Namespace }}.svc.cluster.local:8200"
+    # endpoint and keycloakIssuerUrl both default correctly (see
+    # charts/osac/values.yaml) -- no override needed.
     namespace: osac
     kvMountPath: secret
     lifecycleRole: lifecycle
     lifecycleMountPath: jwt
     keycloakClientId: osac-controller
-    keycloakIssuerUrl: "https://keycloak.keycloak.svc.cluster.local/realms/osac"
     keycloakAudience: osac-api
     caBundle:
       configMap: ca-bundle
diff --git a/osac-installer/values/full-ci/infra.yaml b/osac-installer/values/full-ci/infra.yaml
index 51c23e978..73568f0a6 100644
--- a/osac-installer/values/full-ci/infra.yaml
+++ b/osac-installer/values/full-ci/infra.yaml
@@ -37,4 +37,3 @@ bundledPostgres:
   database:
     name: service
     user: service
-
diff --git a/osac-installer/values/full-ci/instance.yaml b/osac-installer/values/full-ci/instance.yaml
index 319675080..bc37c12c5 100644
--- a/osac-installer/values/full-ci/instance.yaml
+++ b/osac-installer/values/full-ci/instance.yaml
@@ -44,7 +44,9 @@ service:
     caBundle:
       configMap: ca-bundle
   auth:
-    issuerUrl: https://keycloak.keycloak.svc.cluster.local/realms/osac
+    # issuerUrl defaults to a global.clusterDomain-backed external route URL
+    # (see charts/osac/values.yaml) -- correct automatically once
+    # install-osac sets --set global.clusterDomain=$(DOMAIN).
     emergencyServiceAccounts:
     - admin
     - osac-operator
@@ -61,7 +63,9 @@ service:
           param: client-secret
   idp:
     provider: keycloak
-    url: https://keycloak.keycloak.svc.cluster.local
+    # url defaults to a global.clusterDomain-backed external route URL (see
+    # charts/osac/values.yaml) -- correct automatically once install-osac
+    # sets --set global.clusterDomain=$(DOMAIN).
     credentials:
     - secret:
         name: fulfillment-controller-credentials
@@ -89,16 +93,13 @@ service:
   log:
     level: debug
   vault:
-    # tpl is applied to this value at render time — {{ .Release.Namespace }}
-    # expands to the actual install namespace, avoiding DNS failures when
-    # deploying with a non-default NS.
-    endpoint: "https://openbao.{{ .Release.Namespace }}.svc.cluster.local:8200"
+    # endpoint and keycloakIssuerUrl both default correctly (see
+    # charts/osac/values.yaml) -- no override needed.
     namespace: osac
     kvMountPath: secret
     lifecycleRole: lifecycle
     lifecycleMountPath: jwt
     keycloakClientId: osac-controller
-    keycloakIssuerUrl: "https://keycloak.keycloak.svc.cluster.local/realms/osac"
     keycloakAudience: osac-api
     caBundle:
       configMap: ca-bundle
diff --git a/osac-installer/values/vmaas-ci/infra.yaml b/osac-installer/values/vmaas-ci/infra.yaml
index cb683f8bf..158ab0419 100644
--- a/osac-installer/values/vmaas-ci/infra.yaml
+++ b/osac-installer/values/vmaas-ci/infra.yaml
@@ -36,4 +36,3 @@ bundledPostgres:
   database:
     name: service
     user: service
-
diff --git a/osac-installer/values/vmaas-ci/instance.yaml b/osac-installer/values/vmaas-ci/instance.yaml
index b938c08a6..d4bc46f2e 100644
--- a/osac-installer/values/vmaas-ci/instance.yaml
+++ b/osac-installer/values/vmaas-ci/instance.yaml
@@ -56,7 +56,9 @@ service:
     caBundle:
       configMap: ca-bundle
   auth:
-    issuerUrl: https://keycloak.keycloak.svc.cluster.local/realms/osac
+    # issuerUrl defaults to a global.clusterDomain-backed external route URL
+    # (see charts/osac/values.yaml) -- correct automatically once
+    # install-osac sets --set global.clusterDomain=$(DOMAIN).
     emergencyServiceAccounts:
     - admin
     - osac-operator
@@ -73,7 +75,9 @@ service:
           param: client-secret
   idp:
     provider: keycloak
-    url: https://keycloak.keycloak.svc.cluster.local
+    # url defaults to a global.clusterDomain-backed external route URL (see
+    # charts/osac/values.yaml) -- correct automatically once install-osac
+    # sets --set global.clusterDomain=$(DOMAIN).
     credentials:
     - secret:
         name: fulfillment-controller-credentials
@@ -101,16 +105,13 @@ service:
   log:
     level: debug
   vault:
-    # tpl is applied to this value at render time — {{ .Release.Namespace }}
-    # expands to the actual install namespace, avoiding DNS failures when
-    # deploying with a non-default NS (e.g. NS=osac-e2e-ci, NS=osac-devel).
-    endpoint: "https://openbao.{{ .Release.Namespace }}.svc.cluster.local:8200"
+    # endpoint and keycloakIssuerUrl both default correctly (see
+    # charts/osac/values.yaml) -- no override needed.
     namespace: osac
     kvMountPath: secret
     lifecycleRole: lifecycle
     lifecycleMountPath: jwt
     keycloakClientId: osac-controller
-    keycloakIssuerUrl: "https://keycloak.keycloak.svc.cluster.local/realms/osac"
     keycloakAudience: osac-api
     caBundle:
       configMap: ca-bundle
```

</details>

I also noticed the VMaaS and BMaaS E2E runs failed separately with missing `cert-manager`/`cert-manager-cainjector` ServiceAccounts — that looks like an unrelated CI infra issue, not something this PR's changes caused.

Happy to open this as a PR against `main` if that's easier than incorporating it here, or push it to this branch directly if I'm given access to your fork.

DakCrowder (approved):

ygalblum (approved):
/approve
/lgtm

PR #752, opened 2026-09-04
OSAC-4913,OSAC-4915: fix bundled OpenBao OIDC issuer + vault endpoint namespace mismatch on OpenShift

ygalblum:
## Summary

- On OpenShift, Keycloak is deliberately configured to always report the external route hostname as its OIDC `issuer` (portable tokens), and `install-osac`'s `install-osac` target already overrides `service.auth.issuerUrl`/`service.idp.url` at install time to match. However `service.vault.keycloakIssuerUrl` — the OIDC issuer URL the bundled OpenBao's `init.sh` uses for `auth/jwt/config oidc_discovery_url=...` — was missing from that same `$(DOMAIN)`-based override list, so it stayed pinned to the static internal cluster-local hostname from each CI profile's `instance.yaml`. This mismatched Keycloak's actual discovery-document issuer and made OpenBao's OIDC discovery validation fail forever (`issuer did not match the returned issuer`), so the `lifecycle` JWT auth role never got created on any OpenShift-deployed cluster.
- Fix: extend the existing `$(if $(DOMAIN), ...)` `--set` block in `osac-installer/Makefile`'s `install-osac` target to also set `service.vault.keycloakIssuerUrl=https://keycloak-keycloak.$(DOMAIN)/realms/osac`, matching the pattern already used for `service.auth.issuerUrl`. The static per-profile `instance.yaml` values become dead defaults at install time, same as `service.auth.issuerUrl` today — left as-is for consistency/fallback.

## Test plan

- [x] `yamllint --strict -c .yamllint.yaml osac-installer` passes
- [x] `pre-commit run --all-files` passes
- [x] `make helm-lint` / `make helm-validate` fail only with the pre-existing, unrelated `service.externalHostname`/`service.internalHostname` schema error already present on `main` (confirmed before this change)
- [x] `helm template charts/osac -f values/vmaas-ci/instance.yaml` with the same `--set` flags `install-osac` would pass (mocked `DOMAIN`) confirms `KEYCLOAK_ISSUER_URL` in `bundled-openbao.yaml`'s init container now renders as the external route URL (`https://keycloak-keycloak.<domain>/realms/osac`), matching Keycloak's actual discovery-document issuer

Fixes: https://redhat.atlassian.net/browse/OSAC-4913

## Deployment and authentication

- Updated `install-osac` Helm overrides.
- OpenBao now uses the external Keycloak issuer on OpenShift.
- OpenBao now uses the service endpoint in the install namespace through `$(NS)`.
- This fixes OIDC discovery validation and fulfillment-controller login failures.

## Tests and CI

- YAML linting and pre-commit checks pass.
- Helm template testing confirms the issuer override.
- Helm lint and validation still report a pre-existing schema error.

## API surface, controllers, database, and documentation

- No API, controller, database, or documentation changes.

## Backward compatibility

- The overrides apply only when `DOMAIN` is available.
- Other deployment paths keep their existing configuration.

## Risk classification

- **risk:ship** — The change is limited to deployment configuration and has focused template validation.
- It does not qualify as **risk:show** because it adds no broad user-visible feature.
- It does not qualify as **risk:ask** because it changes no application logic or data handling.

### DIFF
--- a/osac-installer/Makefile
+++ b/osac-installer/Makefile
@@ -236,7 +236,7 @@ endif
 	helm upgrade --install osac $(OSAC_CHART) \
 		-f $(INSTANCE_VALUES) $(INSTANCE_VALUES_EXTRA) \
 		--namespace $(NS) --create-namespace \
-		$(if $(DOMAIN),--set service.externalHostname=fulfillment-api-$(NS).$(DOMAIN) --set service.internalHostname=fulfillment-internal-api-$(NS).$(DOMAIN) --set service.auth.issuerUrl=https://keycloak-keycloak.$(DOMAIN)/realms/osac --set service.idp.url=https://keycloak-keycloak.$(DOMAIN)) \
+		$(if $(DOMAIN),--set service.externalHostname=fulfillment-api-$(NS).$(DOMAIN) --set service.internalHostname=fulfillment-internal-api-$(NS).$(DOMAIN) --set service.auth.issuerUrl=https://keycloak-keycloak.$(DOMAIN)/realms/osac --set service.idp.url=https://keycloak-keycloak.$(DOMAIN) --set service.vault.keycloakIssuerUrl=https://keycloak-keycloak.$(DOMAIN)/realms/osac) \
 		$(EXTRA_HELM_ARGS) \
 		--wait --timeout 40m
 
--- a/osac-installer/Makefile
+++ b/osac-installer/Makefile
@@ -236,7 +236,7 @@ endif
 	helm upgrade --install osac $(OSAC_CHART) \
 		-f $(INSTANCE_VALUES) $(INSTANCE_VALUES_EXTRA) \
 		--namespace $(NS) --create-namespace \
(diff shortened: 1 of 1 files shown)


ygalblum:
## Additional fix: [OSAC-4915](https://redhat.atlassian.net/browse/OSAC-4915)

Same root cause category and same Makefile line — `service.vault.endpoint` was hardcoded to `https://openbao.osac.svc.cluster.local:8200` in every CI profile's `instance.yaml`, but CI actually installs OSAC into the `osac-e2e-ci` namespace, not `osac`. This made `fulfillment-controller` fail to log into vault entirely:

```
Failed to obtain vault token: failed to login to vault with JWT: vault login request failed:
Post "https://openbao.osac.svc.cluster.local:8200/v1/auth/jwt/login":
dial tcp: lookup openbao.osac.svc.cluster.local on 10.128.0.10:53: no such host
```

Fixed by extending the same `$(DOMAIN)`-conditional `--set` block in `install-osac` to also set `service.vault.endpoint=https://openbao.$(NS).svc.cluster.local:8200`, using the Makefile's existing `$(NS)` variable.

Verified via `helm template` (renders `--vault-endpoint=https://openbao.osac-e2e-ci.svc.cluster.local:8200` correctly) and live on a running OpenShift dev cluster (fulfillment-controller's vault login errors disappeared after applying).

Fixes: https://redhat.atlassian.net/browse/OSAC-4915

ygalblum:
Closing as it is addressed in https://github.com/osac-project/osac/pull/730

PR #764, opened 2026-09-05
OSAC-4875: block networking provisioning without dispatcher manager

SiddarthR56:
## Controllers

- Provisioning now requires a dispatcher-resolved implementation strategy.
- ExternalIP, ExternalIPAttachment, ExternalIPPool, SecurityGroup, and VirtualNetwork controllers block resources when dispatcher configuration is unavailable.
- Controllers set `Ready=False` with `ReasonNoManagerConfigured` and return reconciliation errors.
- Legacy resource and pool strategy fallbacks were removed.
- Kubernetes fallback now requires the explicitly named `k8s_only` manager.
- Missing required managers use `ErrRequiredManagerUnavailable`.

## API surface

- Added `dispatcher.K8sOnlyManagerName`.
- Added `ResourceDispatchConfig.K8sFallbackManager`.
- Added `dispatcher.ErrRequiredManagerUnavailable`.
- Resource schemas remain unchanged.

## Documentation

- Documented the `k8s_only` dispatcher contract and its Kubernetes-only scope.
- Documented Kubernetes-only fallback behavior for `VirtualNetwork` and `Subnet`.

## Tests

- Added coverage for explicit `k8s_only` fallback resolution.
- Rejected arbitrary Kubernetes managers for fabric-owned resources.
- Updated controller tests for dispatcher resolution, blocking conditions, errors, and no requeue behavior.

## Backward compatibility

- Resources that relied on legacy implementation-strategy fallback can remain unprovisioned until valid dispatcher configuration is present.
- Kubernetes fallback is narrower because only `k8s_only` can satisfy the fallback role.
- Controllers return reconciliation errors instead of periodically requeueing for missing dispatcher configuration.

## Unaffected areas

- No database, authentication, deployment, or CI changes were made.

## Risk classification

`risk:ask` — applied because the change alters provisioning behavior across multiple controllers, removes legacy fallbacks, and can block existing resources when dispatcher configuration is missing. The change is close to `risk:show`, but the cross-controller provisioning impact and backward-compatibility risk require explicit review.

### DIFF
--- a/osac-aap/collections/ansible_collections/osac/templates/roles/k8s_only/README.md
+++ b/osac-aap/collections/ansible_collections/osac/templates/roles/k8s_only/README.md
@@ -26,9 +26,10 @@ it composes the existing, independently-tested k8s-native roles:
 
 `k8s_only` is an explicit Kubernetes-only profile, not a substitute for a
 unified fabric manager. The operator dispatcher permits the Kubernetes fallback
-for SecurityGroup and ExternalIP-family resources only when the NetworkClass
-selects the manager named `k8s_only`. Other Kubernetes managers are overlay
-bridges and still require a `fabricManager` for those resources.
+for VirtualNetwork, Subnet, SecurityGroup, and ExternalIP-family resources only
+when the NetworkClass selects the manager named `k8s_only`. Other Kubernetes
+managers are overlay bridges and still require a `fabricManager` for those
+resources.
 
 As a result, this profile provides Kubernetes-native behavior for VMaaS
 deployments without a physical fabric; it does not provide fabric-wide ACLs,
--- a/osac-aap/collections/ansible_collections/osac/templates/roles/k8s_only/README.md
+++ b/osac-aap/collections/ansible_collections/osac/templates/roles/k8s_only/README.md
@@ -22,6 +22,18 @@ it composes the existing, independently-tested k8s-native roles:
 - **SecurityGroup** → `network_policy` (Kubernetes NetworkPolicy)
 - **ExternalIPPool / ExternalIP / ExternalIPAttachment** → `metallb_l2` (MetalLB L2 advertisement)
 
+## Dispatcher contract
+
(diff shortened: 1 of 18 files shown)


danmanor:
I agree that backward compatibility with the legacy strategy path is not required here, so I am withdrawing the earlier P1 concern. Dispatcher and NetworkClass configuration are intentional prerequisites.\n\nOne remaining design question: the new guard only checks that `resolveImplementationStrategy()` returns a non-empty manager. That helper falls back from `plan.FabricTarget()` to `plan.K8sTarget()`, while the dispatcher table currently marks `SecurityGroup`, `ExternalIP`, `ExternalIPPool`, and `ExternalIPAttachment` with `K8sFallback: true`.\n\nThe unified networking design assigns fabric-manager responsibility for security groups, IPAM, and DNAT. Therefore, a NetworkClass with only `k8sManager` configured could pass the new guard and start a fabric-only operation using a Kubernetes manager. Should these resources require `plan.FabricTarget()` (or should the fallback entries be removed)? Please add a k8s-only NetworkClass test or document that this behavior is intentional.

danmanor:
The missing-manager branch currently sets `Ready=False` but returns `RequeueAfter`, so every affected resource is polled forever. Given that a dispatcher manager and NetworkClass are mandatory prerequisites, this should be treated as a configuration failure rather than a transient precondition. Please avoid the periodic requeue here and fail the reconciliation. If dynamic recovery is required, reconcile on NetworkClass/dispatcher configuration events instead of polling; alternatively, reject the invalid operator configuration at startup. Note that returning an error alone is not terminal in controller-runtime—it will still retry with backoff.

danmanor:
I don’t think we should support a Kubernetes fallback here at all. If the required manager is not explicitly configured for the `NetworkClass`, reconciliation should fail and mark the resource `Ready=False`. Restricting the fallback to `k8s_only` still leaves fallback behavior in place and can hide invalid configuration. Please remove the fallback path and require an explicitly configured, compatible manager; otherwise return an error.

PR #784, opened 2026-09-07
OSAC-4944: Require admin auth for gRPC reflection

mlorenzofr:
Stop treating gRPC reflection as an anonymous endpoint. Remove reflection from the anonymous method regex, restrict OPA authorization to administrators, and update tests and docs accordingly.

## Changes

* Remove gRPC reflection from `anonymousMethodsRegex` so the authn interceptor requires a JWT before reflection requests reach the handler.
* Replace the unauthenticated OPA allow rule with an admin-only rule for `/grpc.reflection.*`.
* Add authn/authz unit tests covering guest rejection, client `PermissionDenied`, and admin access.
* Update `README.md` and `docs/AUTH.md` to document admin-only reflection and show authenticated `grpcurl` examples.

### API and authentication
- gRPC reflection no longer permits anonymous access.
- JWT and service-account identities can access reflection.
- Capabilities, JWKS, metadata, and health endpoints remain anonymous.
- Unauthenticated reflection clients must provide valid credentials.
- This change is backward-incompatible for anonymous reflection clients.

### Authorization
- OPA now requires an authenticated identity for reflection.
- Tests cover tenant-scoped users and administrators.
- No database, controller, or deployment changes are reported.

### Tests
- Added authentication and authorization coverage for reflection.
- Test execution results are not supplied.

### Documentation
- Updated `grpcurl` examples to require authentication.
- Restricted plaintext examples to local development.
- Added TLS examples for remote access.
- Documented anonymous endpoints and authenticated reflection access.

### CI
- No CI changes are reported.
- [OSAC-4944](https://redhat.atlassian.net/browse/OSAC-4944) has no target version according to the supplied CI comment.

## Risk classification

**risk:show** — The change modifies the gRPC reflection access-control boundary. The scope is limited to authentication, authorization, tests, and documentation. No data, deployment, or core service contract changes are reported.

The change is close to **risk:ask** because it changes externally visible security behavior. It does not qualify based on the supplied context because it does not alter data, deployment, or core service contracts. Formal labeling criteria were not supplied.


jhernand (commented):
I'd say that the risk in exposing gRPC reflection isn't that the full service catalog is exposed: it is already exposed because the .proto files are public. The risk is rather that unauthenticated users could use vulnerabilities in the gRPC reflection code to attack the service. Note also that the gRPC reflection service is read only (if there are no vulnerabilities) which reduces a bit the risk. Based on that I'd suggest to either keep unauthenticated access or else allow it to any authenticated user, but not only to admnistrators. If we only allow admnistrators then access with tools like `grpcurl` (which require the gRCP reflection service) will only work for admnistrators (and that is probably why the tests are failing).

mlorenzofr:
> I'd say that the risk in exposing gRPC reflection isn't that the full service catalog is exposed: it is already exposed because the .proto files are public. The risk is rather that unauthenticated users could use vulnerabilities in the gRPC reflection code to attack the service. Note also that the gRPC reflection service is read only (if there are no vulnerabilities) which reduces a bit the risk. Based on that I'd suggest to either keep unauthenticated access or else allow it to any authenticated user, but not only to admnistrators. If we only allow admnistrators then access with tools like `grpcurl` (which require the gRCP reflection service) will only work for admnistrators (and that is probably why the tests are failing).

We can allow requests from authenticated users, but we should remove access for anonymous users. According to security regulations, anonymous access will always be flagged as an anomaly or an indicator of compromise (IOC). I'll refactor the changes to allow requests from authenticated users, not just administrators

PR #790, opened 2026-09-08
OSAC-4678: enforce disabled-service gRPC behavior and add E2E coverage

htayrie-rh:
### Summary

This PR verifies per-service enablement in a real full-install deployment and fixes the gRPC request path for known disabled services.

The gRPC server now checks known-but-disabled service methods in an early tap handler, before authentication, JIT provisioning, and service interceptors. Such requests return `codes.Unavailable` and increment the existing disabled-service Prometheus counter. Genuinely unknown methods continue to return `codes.Unimplemented`.

### E2E setup and coverage

The VMaaS full-install E2E job builds the OSAC components from this PR and installs the VMaaS profile with:

- CaaS enabled;
- VMaaS enabled;
- BMaaS disabled;
- MaaS disabled.

Against that installation, the enablement suite verifies:

- disabled BMaaS gRPC requests return `codes.Unavailable`;
- disabled BMaaS is absent from gRPC reflection;
- the disabled BMaaS REST endpoint returns HTTP `503`;
- disabled BMaaS controllers are not running;
- shared infrastructure remains available;
- enabled services continue to work;
- disabled services are not advertised by Capabilities or HostTypes.

The test-infra workflow then performs a Helm upgrade enabling BMaaS and verifies that BMaaS gRPC, the operator/controller, the API, and Capabilities become available after rollout.

The current CI run passed all enablement tests. Its VMaaS job failed separately in the pre-existing GPU regression test while waiting for a ComputeInstance VM namespace; that failure is unrelated to this PR's enablement tests.

### Files changed

- `fulfillment-service/internal/cmd/service/start/grpcserver/` — disabled-service gRPC tap handler and unit tests.
- `tests/e2e/enablement/` — initial disabled-service and post-upgrade E2E checks.
- `.github/workflows/e2e-vmaas-full-install.yml` — runs the enablement suite in the VMaaS full-install job.
- `osac-installer/values/full-ci/infra.yaml` and E2E fixtures — provide the full-install test setup.

### Scope limits

This PR tests deployed end results. It does not duplicate Helm template rendering or chart validation, test every possible service-flag combination, validate invalid Helm combinations, or test a default all-services installation.

### Dependency

The reusable workflow and test-infra support are provided by [osac-test-infra#535](https://github.com/osac-project/osac-test-infra/pull/535).


tests/e2e/conftest.py
```diff
@@ -51,6 +50,8 @@ def _bmaas_requires_serial_xdist(args: list[str]) -> bool:
         or a.endswith("tests/e2e/bmaas")
         or a.endswith("/e2e/bmaas")
         or a == "e2e/bmaas"
+        or a.endswith("e2e/enablement")
+        or "/e2e/enablement/" in (a + "/")
```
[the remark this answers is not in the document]
htayrie-rh: The broad-selector gap is pre-existing and documented in the function's docstring ("Broader invocations like `pytest tests/` are not detected"). Adding `tests/e2e` as a match would force serial execution for _all_ suites, not just enablement and bmaas — that's not the right fix. Both suites are invoked individually in CI (`pytest tests/e2e/enablement/`, `pytest tests/e2e/bmaas/serial/`), never via the parent path. The scope of this PR is the enablement suite, not redesigning the serial detection mechanism.

tests/e2e/enablement/test_disabled_services.py
```diff
@@ -0,0 +1,121 @@
...
+    output, _rc = run_unchecked(
+        "curl", "-sk", "-o", "-", "-w", "\n%{http_code}", f"https://{host}/api/fulfillment/v1/bare-metal-instances"
+    )
+    lines = output.strip().splitlines()
+    status_code = lines[-1] if lines else ""
+    body = "\n".join(lines[:-1])
+    assert "bareMetalInstances" not in body and "bare_metal_instances" not in body, (
+        f"Disabled service REST endpoint should not return valid BMaaS data, got: {body[:200]}"
+    )
+    assert status_code != "200", f"Disabled service REST endpoint should not return 200, got body: {body[:200]}"
```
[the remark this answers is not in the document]
htayrie-rh: Good catch — fixed in 68e0457. Now rejects the full 2xx range with `not status_code.startswith("2")`.

tests/e2e/enablement/test_disabled_services.py
```diff
@@ -0,0 +1,121 @@
...
+    )
+    pods = json.loads(pods_json)
+    assert pods.get("items"), "osac-operator pod(s) should exist"
+    for pod in pods["items"]:
+        for container in pod.get("spec", {}).get("containers", []):
+            if "manager" in container.get("name", ""):
+                env_map = {e["name"]: e.get("value", "") for e in container.get("env", [])}
+                assert env_map.get("OSAC_ENABLE_BAREMETAL_INSTANCE_CONTROLLER") == "false", (
+                    f"Expected OSAC_ENABLE_BAREMETAL_INSTANCE_CONTROLLER=false, got env: {env_map}"
+                )
```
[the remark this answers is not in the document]
htayrie-rh: Good catch — fixed in 68e0457. Both `test_disabled_services.py` and `test_service_upgrade.py` now track a `found_manager` flag and assert it after the loop. Also redacted the full `env_map` from assertion messages (only shows the specific flag value now) to avoid leaking sensitive env vars in CI logs.

tests/e2e/enablement/test_helm_validation.py
```diff
@@ -0,0 +1,97 @@
...
+    assert env_vars.get("OSAC_ENABLE_BAREMETAL_INSTANCE_CONTROLLER") == "false"
+
+    bmf_deploy = _find_deployment(manifests, "bare-metal-fulfillment")
+    assert bmf_deploy is None, "BMF operator deployment should be absent when BMaaS is disabled"
+
+
+def test_default_values_enable_all_services(helm_template: HelmTemplate) -> None:
+    manifests = helm_template.render()
+
+    fs_deploy = _find_deployment(manifests, "fulfillment")
```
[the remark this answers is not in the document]
htayrie-rh: Good catch — fixed in 68e0457. Changed to `_find_deployment(manifests, "fulfillment-service")` and updated `_get_container_args`/`_get_container_env` calls to match on `"fulfillment-service"` as well, avoiding the BMF deployment collision.

tests/e2e/enablement/test_service_upgrade.py
```diff
@@ -0,0 +1,112 @@
...
+        f"{PUBLIC_API}.ComputeInstances/List",
+        f"{PUBLIC_API}.BareMetalInstances/List",
+    ):
+        output, rc = grpc.call_unchecked(service=svc)
+        assert rc == 0, f"{svc} should succeed after enabling all services, got rc={rc}: {output}"
+
+    response = grpc.call(service=f"{PUBLIC_API}.Capabilities/Get")
+    enabled = response.get("enabledServices", response.get("enabled_services", []))
+    for svc_name in ("caas", "vmaas", "bmaas", "maas"):
+        assert svc_name in enabled, f"Capabilities should include {svc_name} after upgrade, got: {enabled}"
```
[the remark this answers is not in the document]
htayrie-rh: Fixed in 68e0457. Updated to `global.services.bmaas.enabled=true` and removed `maas` from the expected capabilities list since `--reuse-values` preserves the disabled state from the initial deployment. Also updated `test_helm_validation.py` to use `global.services.*` paths consistently.

fulfillment-service/internal/cmd/service/start/grpcserver/unknown_service_handler.go
```diff
@@ -81,6 +83,36 @@ func buildDisabledServiceMap(svcFlags *services.Flags) map[string]string {
...
+				group,
+			)
+		}
+	}
+	return nil
+}
+
+// NewDisabledServiceTapHandler returns a gRPC tap handler that rejects requests
+// for known-but-disabled services before they enter the interceptor chain.
+func NewDisabledServiceTapHandler(
```
jhernand: Would be nice if this component had, like all other components, a builder and setter methods. One of those setters should set the disabled services, so it can be tested without using the "disabledMap" global variable.

fulfillment-service/internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go
```diff
@@ -531,6 +532,7 @@ func (c *runnerContext) run(cmd *cobra.Command, argv []string) error { //nolint:
 			MinTime:             keepaliveConfig.MinTime,
 			PermitWithoutStream: true,
 		}),
+		grpc.InTapHandle(disabledServiceTapHandler),
 		grpc.UnknownServiceHandler(unknownHandler),
```
jhernand: Should we remove this ^ ?

fulfillment-service/internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go
```diff
@@ -518,6 +518,7 @@ func (c *runnerContext) run(cmd *cobra.Command, argv []string) error { //nolint:
 		Help: "Total requests to disabled services.",
 	}, []string{"service"})
 	metricsRegisterer.MustRegister(disabledServiceCounter)
+	disabledServiceTapHandler := NewDisabledServiceTapHandler(c.args.services, disabledServiceCounter)
 	unknownHandler := NewUnknownServiceHandler(c.args.services, disabledServiceCounter)
```
jhernand: Should we remove this ^ ?

fulfillment-service/internal/cmd/service/start/grpcserver/unknown_service_handler.go
```diff
@@ -81,6 +83,36 @@ func buildDisabledServiceMap(svcFlags *services.Flags) map[string]string {
 	return disabled
 }
 
+func disabledServiceError(method string, disabledMap map[string]string, counter *prometheus.CounterVec) error {
```
jhernand: Can/should this be a method of the component, to avoid polluting the package namespace?

fulfillment-service/internal/cmd/service/start/grpcserver/unknown_service_handler.go
```diff
@@ -81,6 +83,36 @@ func buildDisabledServiceMap(svcFlags *services.Flags) map[string]string {
...
+
+// NewDisabledServiceTapHandler returns a gRPC tap handler that rejects requests
+// for known-but-disabled services before they enter the interceptor chain.
+func NewDisabledServiceTapHandler(
+	svcFlags *services.Flags,
+	counter *prometheus.CounterVec,
+) tap.ServerInHandle {
+	disabledMap := buildDisabledServiceMap(svcFlags)
+
+	return func(ctx context.Context, info *tap.Info) (context.Context, error) {
```
jhernand: This function can also be a method of the coponent, and then we can just return it:

```go
func (h *disabledServiceHandler) handle(ctx context.Context, info *tap.Info)  (context.Context, error) {
        ...
}
```

Then:

```go
func (h *disabledServiceHandlerBuilder) Build() (result tao.ServerInHandle, err error) {
        // Create the handler:
        handler := &disabledServiceHandler{...}

        // Return the method:
        result = handler.handle
        return
}
```

fulfillment-service/internal/cmd/service/start/grpcserver/unknown_service_handler.go
htayrie-rh: Done - added a builder and setters, including SetDisabledServices for test injection.

fulfillment-service/internal/cmd/service/start/grpcserver/unknown_service_handler.go
htayrie-rh: Done - made disabledServiceError a method on disabledServiceHandler.

fulfillment-service/internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go
htayrie-rh: Done - removed the custom UnknownServiceHandler.

fulfillment-service/internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go
htayrie-rh: Done - removed its construction; disabled requests now use the tap handler.

fulfillment-service/internal/cmd/service/start/grpcserver/unknown_service_handler.go
htayrie-rh: Done - Build now returns the tap handler directly.

fulfillment-service/internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go
```diff
@@ -512,14 +512,19 @@ func (c *runnerContext) run(cmd *cobra.Command, argv []string) error { //nolint:
...
 	disabledServiceCounter := prometheus.NewCounterVec(prometheus.CounterOpts{
 		Name: "fulfillment_disabled_service_requests_total",
 		Help: "Total requests to disabled services.",
 	}, []string{"service"})
 	metricsRegisterer.MustRegister(disabledServiceCounter)
-	disabledServiceTapHandler := NewDisabledServiceTapHandler(c.args.services, disabledServiceCounter)
-	unknownHandler := NewUnknownServiceHandler(c.args.services, disabledServiceCounter)
+	disabledServiceHandler, err := NewDisabledServiceHandler().
+		SetDisabledServices(buildDisabledServiceMap(c.args.services)).
+		SetCounter(disabledServiceCounter).
```
jhernand: I think it may be better if this component received the metrics registerer, like other components, and created itself the counter. My rationale is that the metrics are an internal implementation detail, and the component may want to offer other metrics in the future.

fulfillment-service/internal/cmd/service/start/grpcserver/unknown_service_handler.go
```diff
@@ -83,10 +84,60 @@ func buildDisabledServiceMap(svcFlags *services.Flags) map[string]string {
 	return disabled
 }
 
-func disabledServiceError(method string, disabledMap map[string]string, counter *prometheus.CounterVec) error {
-	for prefix, group := range disabledMap {
+// disabledServiceHandlerBuilder contains the data and logic needed to create a disabled-service handler. Don't create
+// objects of this type directly; use NewDisabledServiceHandler instead.
+type disabledServiceHandlerBuilder struct {
```
jhernand: I think it is better to make the builder package-public, so that it can be created from other packages if needed.

fulfillment-service/internal/cmd/service/start/grpcserver/unknown_service_handler.go
```diff
@@ -97,40 +148,10 @@ func disabledServiceError(method string, disabledMap map[string]string, counter
...
-			return status.Error(codes.Unimplemented, "unknown service")
-		}
-
-		if err := disabledServiceError(method, disabledMap, counter); err != nil {
-			return err
-		}
-
-		return status.Errorf(codes.Unimplemented, "unknown service %s", method)
+// handle rejects requests to known-but-disabled services before they enter the interceptor chain.
+func (h *disabledServiceHandler) handle(ctx context.Context, info *tap.Info) (context.Context, error) {
```
jhernand: This method is now very short, can we put the code of "disableServiceError" here directly?

fulfillment-service/internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go
htayrie-rh: Done

fulfillment-service/internal/cmd/service/start/grpcserver/unknown_service_handler.go
htayrie-rh: Done

fulfillment-service/internal/cmd/service/start/grpcserver/unknown_service_handler.go
htayrie-rh: Done, although Uncle bob would be mad :)

jhernand (approved):

jhernand (approved):

PR #791, opened 2026-09-08
OSAC-4945: Complete OSAC-4694 storage_tier migration (catalog defaults + e2e assertions)

slintes:
## Summary

Fixes the 8 VMaaS regression e2e failures introduced by [OSAC-4694](https://redhat.atlassian.net/browse/OSAC-4694) (#748, Sep 6), which retyped `ComputeInstanceDisk.storage_tier` from bare `string` to `StorageTierReference {id, name}`. That change migrated vmaas/sanity but left gaps in vmaas/regression (runs only on `schedule`), so PRs merged green while nightly broke.

## Changes

**Server normalization** (real product bug):
- `catalog_item_validation.go`: Added bare-string → `{"name": s}` normalization for `boot_disk.storage_tier` and `additional_disks[N].storage_tier` catalog-item field-definition defaults, mirroring the existing `disk_image` fallback. Before this, a string catalog default became an invalid typed reference → rejected. Real users with string storage-tier catalog defaults hit this.

**E2E test fixes**:
- `test_compute_instance_storage_tier.py`: Fixed catalog-default shape (2 tests) and made error-string assertions param-aware (2 tests):
  1. Catalog defaults → typed `{"name": tier}` (aligned with the server normalization and #748's documented format).
  2. Param-aware assertions: `None` → "storage_tier is required" (validateDisk); `{"name": ""}` → "storagetier reference must specify id or name" (reference interceptor).

## Testing

- Pre-commit hooks pass (Python ruff + Go golangci-lint).
- Scheduled VMaaS e2e will verify green nightly after merge.

## Related

- Jira: [OSAC-4945](https://redhat.atlassian.net/browse/OSAC-4945)
- Root cause: #748 / [OSAC-4694](https://redhat.atlassian.net/browse/OSAC-4694)
- osac-test-infra#478: separate issue (osac-test-infra's own e2e uses latest CLI release (0.0.95) → wire-format errors); not addressed here (osac-project/osac builds CLI from source, no such issue)

## ⚠️ Currently blocked

This PR's vmaas E2E gate is **blocked** by a pre-existing, unrelated failure in the references test suite: cluster/baremetal reference tests require CaaS/BMaaS, which are disabled in the vmaas-ci profile. Tracked in **[OSAC-5006](https://redhat.atlassian.net/browse/OSAC-5006)** (Blocker, OSAC-Infra).

This PR will be **rebased as soon as a fix for [OSAC-5006](https://redhat.atlassian.net/browse/OSAC-5006) is available**, at which point CI should go green.

_Note: a separate low-frequency metering flake ([OSAC-5001](https://redhat.atlassian.net/browse/OSAC-5001)) may also surface intermittently but is not a hard blocker._

## Affected areas

- **Validation:** Normalize legacy string defaults for `disk_image` and `storage_tier`, including `additional_disks` arrays.
- **Tests:** Update VMaaS regression tests for typed references, parameter-aware errors, and legacy array defaults.
- **CI:** Remove duplicate workflow keys and enable AI diagnostics only on `main`.
- **Backward compatibility:** Legacy string catalog defaults remain supported through normalization. Empty references keep distinct validation errors.
- **API, database, auth, and deployment:** No changes.

## Risk classification

**risk:show** — The change affects provisioning validation and workflow behavior, but it is localized and includes regression coverage.

No repository risk-label criteria were found. The change is close to **risk:ship** because it has focused changes and reported green regression tests, but the VMaaS e2e gate remains blocked by [OSAC-5006](https://redhat.atlassian.net/browse/OSAC-5006).

### DIFF
--- a/.github/workflows/e2e-bmaas-full-install.yml
+++ b/.github/workflows/e2e-bmaas-full-install.yml
@@ -247,7 +247,6 @@ jobs:
       # is fork-originated and gets no OIDC/WIF access at all by GitHub
       # Actions policy. merge_group is excluded the same way (its
       # github.ref is the readonly-queue ref).
-      notify-on-failure: ${{ github.event_name == 'schedule' }}
       enable-ai-diagnostic: ${{ github.ref == 'refs/heads/main' }}
 
   # Skip this required check on unreadiness so it stays pending (not red).
--- a/.github/workflows/e2e-caas-full-install.yml
+++ b/.github/workflows/e2e-caas-full-install.yml
@@ -264,7 +264,6 @@ jobs:
       # is fork-originated and gets no OIDC/WIF access at all by GitHub
       # Actions policy. merge_group is excluded the same way (its
       # github.ref is the readonly-queue ref).
-      notify-on-failure: ${{ github.event_name == 'schedule' }}
       enable-ai-diagnostic: ${{ github.ref == 'refs/heads/main' }}
 
   # Skip this required check on unreadiness so it stays pending (not red).
--- a/.github/workflows/e2e-vmaas-full-install.yml
+++ b/.github/workflows/e2e-vmaas-full-install.yml
@@ -263,7 +263,6 @@ jobs:
       # is fork-originated and gets no OIDC/WIF access at all by GitHub
       # Actions policy. merge_group is excluded the same way (its
       # github.ref is the readonly-queue ref).
-      notify-on-failure: ${{ github.event_name == 'schedule' }}
       enable-ai-diagnostic: ${{ github.ref == 'refs/heads/main' }}
 
   # Skip this required check on unreadiness so it stays pending (not red).
--- a/fulfillment-service/internal/servers/catalog_item_validation.go
+++ b/fulfillment-service/internal/servers/catalog_item_validation.go
@@ -191,6 +191,29 @@ func validateCatalogItemAccess(item catalogItem, ref string) error {
 	return nil
 }
 
+// normalizeDiskField wraps bare-string disk field values (disk_image, storage_tier)
+// as typed reference objects for backward compatibility with legacy catalog item defaults.
+func normalizeDiskField(v any, fieldName string) any {
+	switch val := v.(type) {
+	case string:
+		return map[string]any{"name": val}
+	case map[string]any:
+		if fieldVal, ok := val[fieldName]; ok {
+			if s, ok := fieldVal.(string); ok {
+				val[fieldName] = map[string]any{"name": s}
+			}
+		}
+		return val
+	case []any:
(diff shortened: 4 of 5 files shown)


slintes:
pushed to origin as well in order to run regression tests: https://github.com/osac-project/osac/actions/runs/34198962034

slintes:
regression tests are green :)

clobrano:
There seems to be a scope gap in the server normalization: the boot disk case is handled, but additional disk defaults are not. `applyDefault` receives `path == "additional_disks"` and the complete array, while the new condition only matches indexed paths such as `additional_disks[0].storage_tier`. A legacy default like `[{"storage_tier": "fast"}]` therefore remains a string and fails typed proto unmarshalling.

The tests cover these cases:

1. Boot disk default with an already-typed value:

```python
"default": {"name": default_storage_tier}
```

2. Additional disk default with an already-typed value:

```python
"path": "additional_disks"
"default": [{"storage_tier": {"name": fast_tier}}]
```

That second test does exercise `applyDefault` with `path == "additional_disks"`, but there is no string to convert, so the new condition is skipped.

The only remaining bare-string default is overridden by an explicit user value, so `applyDefault` is not called for it.

Therefore, no test exercises:

```python
"path": "additional_disks"
"default": [{"storage_tier": "fast"}]
```

which is the failing legacy case.

slintes:
[AI-Assisted] Great catch! You're absolutely right — the array-default case was missed.

The initial fix only normalized scalar leaf paths (`boot_disk.storage_tier`, `additional_disks[N].storage_tier`), but `applyDefault` can also receive whole **objects** (`path == "boot_disk"`) or **arrays** (`path == "additional_disks"`) as defaults. When it receives the array, the path-prefix condition never matched, and even if it did, the string type-assertion would fail on the `[]any` value — so legacy defaults like `[{"size_gib": 10, "storage_tier": "fast"}]` stayed bare strings and broke proto unmarshalling.

Fixed in 780553f50a5:
- Extracted a `normalizeDiskField` helper that normalizes at all three granularities (scalar, object, array).
- Applied to both `disk_image` and `storage_tier` at `boot_disk`/`additional_disks*` paths.
- Added a regression test (`test_compute_instance_additional_disks_from_catalog_item_array_default_legacy_string`) that exercises the exact legacy shape you described.

Thanks for the thorough review!

slintes:
triggered new regression test: https://github.com/osac-project/osac/actions/runs/34213008477

clobrano (approved):
/lgtm

slintes:
blocked by https://redhat.atlassian.net/browse/OSAC-5006. will rebase when a fix is available

clobrano (approved):
/lgtm

PR #805, opened 2026-09-08
OSAC-5006: filter service-specific VMaaS references

danielerez:
Exclude CaaS and BMaaS reference tests from the VMaaS profile while preserving per-test markers for reuse by the corresponding suites.

Assisted-by: Codex <noreply@openai.com>

## Affected areas

- **CI:** The VMaaS full-install workflow now excludes tests marked `requires_caas` or `requires_bmaas` by default. A user-supplied `test-marker` still overrides this default.
- **Tests:** Pytest now registers `requires_caas` and `requires_bmaas`. Reference tests now identify their required service.
- **API, controllers, database, auth, deployment, and documentation:** No changes.

## Compatibility

This change does not alter production APIs or runtime behavior. It changes the default VMaaS E2E test selection. CaaS- and BMaaS-specific tests will not run in the VMaaS workflow unless a user supplies an overriding marker.

The CaaS and BMaaS workflows must include the reference test path to execute the excluded tests. This PR does not add that workflow coverage.

## Risk classification

`risk:show` applies because the change affects CI test selection and can hide service-specific E2E coverage from the VMaaS workflow.

It does not qualify as `risk:ask` because it does not change production code, APIs, authentication, data, or deployment behavior. It is close to `risk:ship` because the code change is limited and reversible, but the missing CaaS and BMaaS workflow coverage prevents classification as low-risk.

### DIFF
--- a/.github/workflows/e2e-vmaas-full-install.yml
+++ b/.github/workflows/e2e-vmaas-full-install.yml
@@ -222,7 +222,11 @@ jobs:
           'vmaas/sanity'
         }}
       test-filter: ${{ inputs.test-filter || '' }}
-      test-marker: ${{ inputs.test-marker || '' }}
+      # The references module contains CaaS and BMaaS tests too. Those
+      # services are disabled in the VMaaS profile, so exclude tests marked
+      # with either dependency unless an explicit workflow-dispatch marker is
+      # supplied.
+      test-marker: ${{ inputs.test-marker || 'not requires_caas and not requires_bmaas' }}
       # osac-installer now lives as a subdirectory of this same monorepo checkout
       # (its Chart.yaml deps resolve fulfillment-service/osac-operator/osac-aap/BMF
       # as ../../../<name> siblings within it), so a PR touching those components
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -82,6 +82,8 @@ def pytest_configure(config: pytest.Config) -> None:
     e2e.log artifact.
     """
     config.addinivalue_line("markers", "metering: test verifies metering events via the test adapter HTTP API")
+    config.addinivalue_line("markers", "requires_caas: test requires the CaaS service to be enabled")
+    config.addinivalue_line("markers", "requires_bmaas: test requires the BMaaS service to be enabled")
     config.addinivalue_line("markers", "sanity: fast, low-risk smoke test suitable for every PR")
     config.addinivalue_line("markers", "regression: broader/slower coverage, run on a schedule or on demand")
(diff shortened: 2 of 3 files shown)


htayrie-rh:
I agree with adding per-test markers, but this PR does not yet guarantee that every affected test runs in at least one suite.

The current change guarantees only that VMaaS will stop running the CaaS/BMaaS reference tests. However, the CaaS and BMaaS workflows are unchanged and do not add `tests/e2e/references/` to their test paths. As a result:

- The tests marked `requires_caas` are excluded from VMaaS, but this PR does not show that they are executed by CaaS.
- The test marked `requires_bmaas` is excluded from VMaaS, but this PR does not show that it is executed by BMaaS.
- The other reference modules remain unmarked and continue to run in VMaaS.

Suggested solution:

1. Keep the individual markers in `test_cluster_baremetal_references.py`; the per-test distinction between CaaS cluster tests and the BMaaS bare-metal test is correct.
2. Add `tests/e2e/references/` to the CaaS and BMaaS reusable workflow invocations.
3. Apply service filters in those workflows as well, while combining them with any user-supplied marker instead of replacing the service filter. For example:
   - VMaaS: `(<user marker>) and not requires_caas and not requires_bmaas`
   - CaaS: `(<user marker>) and not requires_bmaas`
   - BMaaS: `(<user marker>) and not requires_caas`
4. Add a collection/coverage check that maps every E2E test path to at least one active suite and fails CI when a test becomes orphaned. This is the only reliable way to guarantee future coverage; workflow path lists alone can silently leave files unexecuted.

If “all tests” means every file under `tests/e2e`, the same ownership check must also cover catalog, projects, storage, and any optional Netris-specific tests. This PR currently addresses only the VMaaS reference-test failure, not complete E2E test ownership.

slintes:
we have to either merge this or https://github.com/osac-project/osac/pull/791 manually, and then rebase the other one,  to get green vmaas e2e...

danielerez:
> we have to either merge this or #791 manually, and then rebase the other one, to get green vmaas e2e...

oh, so these failures are expected for now?
```
=========================== short test summary info ============================
FAILED tests/e2e/vmaas/regression/test_compute_instance_storage_tier.py::test_compute_instance_boot_disk_tier_required[storage_tier1]
FAILED tests/e2e/vmaas/regression/test_compute_instance_storage_tier.py::test_compute_instance_additional_disk_tier_required[storage_tier1]
FAILED tests/e2e/vmaas/regression/test_compute_instance_storage_tier.py::test_compute_instance_boot_disk_tier_from_catalog_item_default
FAILED tests/e2e/vmaas/regression/test_compute_instance_storage_tier.py::test_compute_instance_explicit_additional_disks_without_catalog_item_default
```

Then I guess either is fine:/

PR #812, opened 2026-09-09
Retry transient Go module downloads in image builds

omer-vishlitzky:
## Problem
E2E image builds run `go mod download` inside Podman. The existing host-side Go retry action cannot affect those layers, and proxy.golang.org HTTP/2 stream errors have failed builds across multiple runners.

## Fix
Wrap each OSAC image build module download in five bounded attempts. Retry only transport signatures such as HTTP/2 stream errors, connection resets, timeouts, proxy connection failures, DNS failures, and 4xx/5xx transient responses. Non-transport failures preserve their original exit status immediately.

Checksum verification remains enabled; no dependency or module error is hidden.

## Validation
- `git diff --check`
- Shell simulation verifies transient errors retry and non-transient module errors fail immediately.
- `podman --version` verified on the build host; a full image build was not run because it requires the external registries and dependency downloads.

## Summary

- **Deployment:** Adds bounded retries for `go mod download` in seven OSAC `Containerfile` build stages.
- Retries cover transient transport, proxy, DNS, timeout, HTTP 408/429, and HTTP 5xx failures.
- Non-transport failures retain their original status and fail immediately.
- Checksum verification remains enabled.
- Validation covered formatting and shell simulations. Full image builds were not run because they require external registries and dependency downloads.
- No changes affect APIs, controllers, databases, authentication, runtime behavior, CI, or documentation.
- **Backward compatibility:** No runtime or API compatibility impact. The change improves build resilience.

## Risk classification

**risk:ship** — The change is limited to bounded dependency-download retries. It preserves checksum verification and fail-fast behavior for non-transport errors.

The change does not qualify as **risk:show** because it does not alter deployed runtime behavior. It does not qualify as **risk:ask** because retry count and backoff are bounded, and no unresolved data-integrity or security risk is evident.

### DIFF
--- a/bare-metal-fulfillment-operator/Containerfile
+++ b/bare-metal-fulfillment-operator/Containerfile
@@ -17,7 +17,22 @@ COPY --chown=1001:1001 osac-csi-driver/go.mod osac-csi-driver/go.sum osac-csi-dr
 COPY --chown=1001:1001 osac-metering/schema/go.mod osac-metering/schema/go.sum osac-metering/schema/
 COPY --chown=1001:1001 osac-metering/metering-service/go.mod osac-metering/metering-service/go.sum osac-metering/metering-service/
 COPY --chown=1001:1001 osac-metering/adapters/go.mod osac-metering/adapters/go.sum osac-metering/adapters/
-RUN cd bare-metal-fulfillment-operator && go mod download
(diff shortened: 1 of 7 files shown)


masayag:
A different approach is needed.

PR #822, opened 2026-09-09
OSAC-4922: Default networking resources enter PENDING instead of rejecting deletion

SiddarthR56:
## Summary

- **API and resource lifecycle:** Protects the default label on default virtual networks, subnets, security groups, and NAT gateways. Updates that remove the label now fail. Deletion of default resources remains blocked.
- **Controllers and cleanup:** Verifies that root-project deletion removes tenant virtual networks, subnets, and security groups.
- **Tests:** Adds regression coverage for rejected deletions, rejected label removal, preserved resource state, and cleanup of system-owned networking resources.
- **Database and deployment:** No changes.
- **Auth and CI:** No changes.
- **Documentation:** No changes.

## Backward compatibility

Clients that attempt to remove the default label from an existing default resource will now receive an error. Other label updates and unchanged default labels remain allowed. No public declarations changed.

## Risk classification

**risk:show** — The change affects networking resource update and deletion behavior, but it is narrowly scoped and includes regression tests for each affected resource type.

The change does not qualify for **risk:ship** because it changes API behavior for existing update requests and affects resource lifecycle protection. It does not qualify for **risk:ask** because it does not change authentication, deployment, data schemas, or public API declarations.

### DIFF
--- a/fulfillment-service/internal/servers/default_networking_provisioner.go
+++ b/fulfillment-service/internal/servers/default_networking_provisioner.go
@@ -23,6 +23,7 @@ import (
 	grpccodes "google.golang.org/grpc/codes"
 	grpcstatus "google.golang.org/grpc/status"
 	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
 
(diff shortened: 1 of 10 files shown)


danmanor (changes requested):
Please add integration tests to prove it works

danmanor (approved):

PR #833, opened 2026-09-09
OSAC-5099: Add Secret Types

DakCrowder:
Adds a type to the `Secret` api

Types include:
- Pull secret
- Kubeconfig
- User data
- Opaque
  - Takes k/v pairs, user can set whatever they want and we have no requirements
- Value
  - Intended for secrets that are just a single value, we make sure a value key exists and consumers can count on that when reading

We perform some basic validation at the secrets api layer - making sure a pull secret has a .dockerconfigjson key for example.  Validation also occurs in apis that consume secret references, for instance making sure that a secret assigned to a ComputeInstance is indeed of type pull secret.

## Summary

- **API surface:** Added `Secret.type` to the public and private APIs. Supported types include pull secret, kubeconfig, user data, opaque, and value.
- **Validation:** Added type-specific required data-key validation. Secret types are immutable. Unspecified types default to `OPAQUE`.
- **Reference resolution:** Centralized local secret reference resolution and type validation. Consuming APIs now require the expected secret type.
- **Controllers:** Updated cluster and tenant controllers to create typed secrets. Updated storage credentials to require `SECRET_TYPE_VALUE` and read `data["value"]`.
- **Compatibility:** Existing secrets without a type remain compatible through the `OPAQUE` default. Existing storage credentials that use `data["password"]` must migrate to `SECRET_TYPE_VALUE` and `data["value"]`.
- **Tests and documentation:** Added CLI, unit, integration, controller, API, and validation coverage. Updated API documentation and filter examples.
- **Database, deployment, and CI:** No database schema, deployment, or CI changes are described.

## Risk classification

- Applied label: `risk:ask`.
- The supplied evidence does not include the labeling criteria or the specific reason for this label.
- The available evidence does not support comparison with `risk:ship` or `risk:show`.


fulfillment-service/internal/servers/private_identity_providers_server.go
```diff
@@ -218,54 +212,15 @@ func (s *PrivateIdentityProvidersServer) validateClientSecretSecret(
...
-		data = fetched
-	}
-	if value, ok := data[secretValueKey]; !ok || len(value) == 0 {
-		return grpcstatus.Errorf(grpccodes.InvalidArgument,
-			"secret '%s' referenced by client_secret_secret must contain a non-empty '%s' entry",
-			refKey(ref), secretValueKey)
-	}
 	resolvedRef := &privatev1.SecretLocalReference{}
 	resolvedRef.SetId(resolved.ID)
 	resolvedRef.SetName(resolved.Name)
```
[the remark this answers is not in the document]
DakCrowder: The created hub secrets aren't storing data directly in the create action.  We are always at the mercy of having to fetch and/or map the retrieved data from k8s secrets to data["value"] if needed in the future but I don't think it is worth fetching at create time given these are externally managed system secrets created through the admin api at this time.

fulfillment-service/it/it_cluster_pull_secret_test.go
```diff
@@ -74,6 +75,25 @@ var _ = Describe("Cluster pull_secret_secret", Label("secrets", "cluster"), func
...
+				Type: publicv1.SecretType_SECRET_TYPE_OPAQUE,
+				Metadata: publicv1.Metadata_builder{
+					Name: name,
+				}.Build(),
+			}.Build(),
+		}.Build())
+		Expect(err).ToNot(HaveOccurred())
+		id = response.GetObject().GetId()
+		DeferCleanup(func(ctx context.Context) {
+			_, _ = secretsClient.Delete(ctx, publicv1.SecretsDeleteRequest_builder{Id: id}.Build())
```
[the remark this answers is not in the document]
DakCrowder: This is a common pattern throughout the it test - we try as best we can but not worth failing on cleanup

fulfillment-service/internal/cmd/cli/create/secret/create_secret_cmd.go
```diff
@@ -128,6 +141,26 @@ func (c *runnerContext) run(cmd *cobra.Command, args []string) error {
...
+	case "pull-secret":
+		return publicv1.SecretType_SECRET_TYPE_PULL_SECRET, nil
+	case "kubeconfig":
+		return publicv1.SecretType_SECRET_TYPE_KUBECONFIG, nil
+	case "user-data":
+		return publicv1.SecretType_SECRET_TYPE_USER_DATA, nil
+	case "value":
+		return publicv1.SecretType_SECRET_TYPE_VALUE, nil
+	default:
+		return publicv1.SecretType_SECRET_TYPE_UNSPECIFIED, fmt.Errorf(
```
ygalblum, line 157: IIUC the backend has a default type: https://github.com/osac-project/osac/pull/833/changes#diff-fccd7529502fb788d9512553b2aa99910844f98ab9825463d7aff74cd377dd58R415
If I'm right, why block the CLI?

ygalblum (commented):
Small comment about the CLI. Otherwise, all good

fulfillment-service/internal/cmd/cli/create/secret/create_secret_cmd.go
DakCrowder: The CLI isn't actually blocked, this is just a catch for invalid flags.  The flag registration defines "opaque" as the default value higher up so this would only trigger if something like `--type=some-invalid-type` gets attempted

ygalblum: I see it now. Thanks

ygalblum (approved):
/approve
/lgtm

PR #898, opened 2026-09-10
OSAC-5183: Add release mode to nightly-build.yaml

eliorerz:
## Summary

Two independent implementations of "build/package/publish the OSAC
umbrella chart" exist today: `nightly-build.yaml` (scheduled + manual,
continuous, mature) and `publish-osac-installer-chart.yaml` (manual-only,
intended as the official release path, but effectively dormant -- no
clean non-nightly `osac/vX.Y.Z` tag exists since 2026-07). This
duplication already caused a real bug: [OSAC-5178](https://redhat.atlassian.net/browse/OSAC-5178) had to fix the same
OCI-dependency-rewrite regression in both workflows separately, because
a fix existed in one and not the other.

This PR is **step 1 of 2** (per the ticket's explicit sequencing): it adds
an opt-in "release mode" to `nightly-build.yaml`, covering what
`publish-osac-installer-chart.yaml` does today that nightly doesn't.
`publish-osac-installer-chart.yaml` is deliberately left in place and
untouched -- see "Deliberately left as a human decision" below.

## What changed

Two new `workflow_dispatch` inputs on `nightly-build.yaml`:

- **`release_version`** -- a human-chosen umbrella chart version (e.g.
  `0.1.0`), replacing the auto-generated
  `nightly.<date>.<sha>.<run>.<attempt>` suffix for that run. This is
  what makes `osac/v<version>` a real, clean release tag again. Validated
  as a real semver via a new shared `validate_semver` helper in `lib.sh`
  (same grammar `publish-osac-installer-chart.yaml` already validates
  inline -- now available to both instead of duplicated in only one, the
  exact class of drift [OSAC-5178](https://redhat.atlassian.net/browse/OSAC-5178) fixed).
- **`force_shared_component_version`** -- opt-in (requires
  `release_version`) to force that same version onto every mono-repo
  component's chart *and* image tags too, instead of each resolving its
  own latest release tag independently (nightly's default, unchanged).
  Unlike the old workflow, this needs no separate `<component>/vX.Y.Z`
  git tags -- nightly already builds, packages, and pushes every
  component's chart and image itself, every run.

**Ref/commit selection** (the ticket's third requirement) needed no new
input. `nightly-build.yaml`'s checkout already uses `github.sha`, which
already tracks whatever branch/tag a `workflow_dispatch` run was
dispatched against via GitHub's own "Use workflow from" ref selector --
the same mechanism `publish-osac-installer-chart.yaml` already relies on
today (it has no custom ref input either). I deliberately did **not** add
a free-text arbitrary-commit input instead: several steps (image tagging
in `build`, promotion in `publish`) read the literal `GITHUB_SHA`
run-level variable directly, not a per-job `git rev-parse HEAD`. That
value is fixed for the whole run by whatever ref triggered it and stays
consistent with what's actually checked out *only* as long as the
checkout ref and the dispatch ref are the same thing -- decoupling them
would risk two different release attempts colliding on the same
provisional `sha-<short>` image tag. Documented this as the input's own
description text so a human dispatching a release sees it right in the
GitHub UI.

**Concurrency fix**: release-mode runs now get their own concurrency
group, keyed off the chosen version
(`nightly-build-release-<version>`), instead of sharing `nightly-build`'s
group with `cancel-in-progress: true`. Without this, a real release
dispatch could be silently cancelled mid-flight by the next scheduled
03:00 UTC nightly cron (or vice versa) -- exactly the kind of
hard-to-notice failure mode that shouldn't exist for something this
consequential.

**Small necessary side-fix**: `stamp_component_image_refs` never stamped
`aap.configAsCode.eeImage` (a real, schema-defined, template-consumed
field -- confirmed live in `osac-aap/charts/aap/templates/config-as-code-secret.yaml`).
`publish-osac-installer-chart.yaml` already stamps it;
`force_shared_component_version` needs to actually mirror that behavior
to be a real substitute. This applies uniformly (not gated on release
mode), so it also fixes nightly's own default runs, which have silently
left this field at its `""` placeholder in every published chart until
now. I deliberately did **not** replicate `publish-osac-installer-chart.yaml`'s
`.dbMigrate.image` stamp alongside it -- `dbMigrate` doesn't exist in the
umbrella chart's actual `values.yaml` or `values.schema.json` at all (only
in `values-example.yaml` and one CI test overlay, for a feature its own
comment says is "disabled until the 'migrate' subcommand is released").
Stamping it would auto-vivify an unschemed field into the packaged chart,
violating `osac-installer/AGENTS.md`'s "every value must have a matching
schema definition" invariant for a dead field nothing reads.

Also updated a few user-facing strings (Slack header/message, git tag
message, workflow run-name) to say "Release" instead of "Nightly Build"
when `release_version` is set, so a release run doesn't look
indistinguishable from routine nightly noise in the places people
actually look.

## Deliberately left as a human decision

Per the ticket's explicit sequencing, **`publish-osac-installer-chart.yaml`
is not touched or retired in this PR.** I'm not confident the new release
mode is fully proven without an actual production release dispatch, which
I cannot trigger myself. My recommendation: keep both workflows in place
until one real successful release dispatch (`release_version` set,
optionally `force_shared_component_version` too) has gone through
`nightly-build.yaml` end-to-end -- full build, all tests, e2e, publish,
and tag creation -- and produced a working `osac/vX.Y.Z` chart. Only then
retire `publish-osac-installer-chart.yaml` as a follow-up PR.

## What I validated

- **`validate_semver`**: valid/invalid semver strings including edge
  cases from the existing regex's own history (leading zeros, malformed
  prerelease identifiers, a `v`-prefixed string, a bare `1.2`).
- **Version-resolution logic**: extracted the actual `prepare` job's
  "Resolve nightly versions" step body and ran it against a scratch git
  repo with real component tags, covering all three modes -- normal
  nightly (independent per-component, unchanged), release mode with
  independent components (`release_version` set,
  `force_shared_component_version` false), and forced shared version
  (both set) -- confirming each produces exactly the expected
  `COMPONENT_VERSIONS` JSON. Also confirmed both validation guards
  (invalid semver; `force_shared_component_version` without
  `release_version`) fail loudly with the intended error and a non-zero
  exit.
- **`publish` job's "Generate release/nightly version" step**: same
  extraction technique, confirmed it produces the auto-suffixed nightly
  version when `release_version` is unset and the verbatim human-chosen
  version when it's set.
- **`stamp_component_image_refs`'s new `eeImage` stamp**: ran it against
  real copies of the umbrella and `osac-aap` subchart `values.yaml`,
  confirmed only the intended two lines changed, the umbrella file stays
  valid YAML, and -- in a full scratch mono-repo checkout with all 6
  components stamped to a shared `0.1.0` (simulating
  `force_shared_component_version`) -- `helm dependency build` +
  `helm lint` + `helm template` succeed against every
  `charts/osac/ci/*-values.yaml` and `values/*/instance.yaml` overlay,
  and the rendered `AAP_EE_IMAGE` secret value actually reflects the new
  stamp.
- **`ct lint`**: installed `ct` via `go install`, but it additionally
  requires `yamale` (Python) for schema validation, which I could not
  install in this environment (no working `pip`). Substituted `helm
  lint`/`helm template` across every overlay instead, same as the
  [OSAC-5178](https://redhat.atlassian.net/browse/OSAC-5178) PR.
- **Static checks**: `bash -n` on the changed shell; `pre-commit run`
  (yamllint --strict + repo-wide hooks) on all three changed files;
  `actionlint` against the whole modified workflow file -- zero new
  findings versus the pre-existing baseline (one unrelated, pre-existing
  `osac-ci` self-hosted-runner-label warning, confirmed present before my
  changes too). This also gives reasonable confidence the new `run-name`
  and `concurrency.group` expressions (`format()`, `&&`/`||`) are
  syntactically valid, since `actionlint` parses and type-checks
  expression syntax.

## What I could not validate (needs a real dispatch)

- The actual end-to-end `workflow_dispatch` behavior: whether GitHub
  correctly evaluates the new `run-name`/`concurrency.group` expressions
  at dispatch time, whether the two new inputs render and behave as
  expected in the real dispatch UI, and whether the full job graph
  (build -> tests -> e2e -> publish -> tag-and-notify) genuinely produces
  a working, installable `osac/v<release_version>` chart end-to-end.
- Real concurrency behavior: that a release-mode run and a concurrent
  nightly cron run truly don't cancel each other, and that two release
  dispatches for the same version do serialize/cancel as intended.
- Real Slack delivery of the updated header/message text.

## Explicitly out of scope (per the ticket)

- Nightly's default independent per-component versioning for
  scheduled/non-release runs -- unchanged.
- Local dev / PR-validation `file://` chart dependency resolution --
  untouched (already correctly scoped by [OSAC-5178](https://redhat.atlassian.net/browse/OSAC-5178)).
- Anything about a future Konflux/registry.redhat.io release path
  ([OSAC-4224](https://redhat.atlassian.net/browse/OSAC-4224)) -- this is only about the existing GHCR-based mechanism.
- Creating separate `<component>/vX.Y.Z` git tags or per-component
  GitHub Releases for `force_shared_component_version` -- deliberately
  not added; nightly's existing per-run build/package/push already covers
  this without them, and adding them would reintroduce coordination
  overhead this ticket is trying to remove.

## Summary

- **CI and release deployment:** `nightly-build.yaml` adds opt-in release inputs, SemVer validation, release-specific names, concurrency groups, tags, publishing, and Slack behavior.
- **Helm and image configuration:** `nightly-charts.sh` stamps `aap.configAsCode.eeImage` with the `bootstrap.image` reference.
- **Validation:** `validate_semver` validates release versions. Shared versioning requires `release_version`.
- **Verification:** Local validation passed. Real release dispatch, concurrency behavior, Slack delivery, and production release behavior remain unvalidated.

## Backward compatibility

Nightly runs remain unchanged when release inputs are absent. The new release behavior is opt-in. `publish-osac-installer-chart.yaml` remains unchanged.

## Risk classification

**risk:show** — The change affects CI release execution, version validation, concurrency, image stamping, tagging, and notifications. End-to-end release behavior is not validated. It is not **risk:ship** because production release execution remains untested. It is not **risk:ask** because the change is opt-in and does not modify runtime application behavior or persistent data.


eliorerz:
Holding this PR — the versioning design (force-shared-version vs. a manifest/pointer model) turned out to have real open questions (duplicate tags, version-ordering regressions when a component's own version has already advanced past the chosen release number). Stepping back to write up [OSAC-5183](https://redhat.atlassian.net/browse/OSAC-5183) as a proper enhancement proposal before continuing implementation. Will follow up once that's settled.

eliorerz:
Unholding — the versioning design question this was waiting on is now settled.

The [PRD/design EP](https://github.com/osac-project/enhancement-proposals/pull/283) merged, and the team weighed in in Slack: [Oved](https://redhat-internal.slack.com) and Eran both came down on independent per-component versions + the umbrella chart having its own separate release version (the manifest/pointer model) — matching both the PRD and design.md's primary proposal.

That's exactly this PR's *default* behavior (no `force_shared_component_version` set). The flag stays as an opt-in escape hatch, not the default — no rework needed.

Marking ready for review. Per the PR's own note, I'd still like one real `release_version`-set dispatch of `nightly-build.yaml` to go green end-to-end (build, tests, e2e, publish, tag) before we treat this as fully proven and follow up on retiring `publish-osac-installer-chart.yaml`.

eliorerz:
Went through all 4 findings against current code before touching anything:

**Finding 1 (missing signing before tag-and-notify) — confirmed, fixed by rebase, not new code.** This branch was based on a commit from before [OSAC-1148](https://redhat.atlassian.net/browse/OSAC-1148) (cosign signing) merged into main — `nightly-build.yaml` on `main` already signs images/charts inside the `publish` job, which already runs before `tag-and-notify`. Hand-writing new signing logic here would have duplicated/drifted from what already exists (the exact class of bug [OSAC-5178](https://redhat.atlassian.net/browse/OSAC-5178) fixed). Rebased onto current `upstream/main` instead — clean, no conflicts — so release mode now inherits signing for free.

**Finding 2 (concurrency-group case-sensitivity) — skipping, not confirmed as a real bug.** Semver treats prerelease identifiers as case-sensitive by spec (`1.0.0-RC1` and `1.0.0-rc1` are legitimately different versions), and the group key preserves the input verbatim via `format()` — no case-folding happens in this code. I have no evidence GitHub Actions concurrency groups fold case either. Worst case if they did: two genuinely different versions would serialize against each other unnecessarily — overly cautious, not a collision. No repro, no fix applied.

**Finding 3 (regex length) — valid, fixed.** Added a `${#version} -le 128` guard in `validate_semver` before the regex runs, cheap defense-in-depth against the prerelease clause's repeated-alternation groups.

**Finding 4 (build metadata breaking retagging) — confirmed real, fixed earlier than suggested.** `retag_component_image` (nightly-charts.sh) already rejects `+` via its `^[a-zA-Z0-9._-]+$` guard, so this would never silently produce a bad tag — but it would fail only after a full build/test/e2e run. Added an early check: `force_shared_component_version` + a `release_version` containing build metadata now fails immediately at input-validation time with a clear message, before any of that work happens.

Rebased, validated (YAML parse, bash syntax, functional checks on both new guards), and force-pushed to the same branch.

PR #902, opened 2026-09-11
OSAC-1148: Sign Helm chart OCI artifacts with cosign

eliorerz:
## Summary

Second slice of [OSAC-1148](https://redhat.atlassian.net/browse/OSAC-1148) (container images landed in #901). Signs every Helm chart this repo publishes to `oci://ghcr.io/osac-project/charts`, using the same keyless cosign approach as the container-image work: GitHub Actions OIDC via Fulcio/Rekor, no private key material to manage.

- New shared composite action `.github/actions/helm-push-and-sign` wraps `helm push` + digest capture + cosign sign, instead of duplicating that sequence across 10 call sites.
- Applied to all 9 chart-publish jobs in `publish-charts.yaml` (fulfillment-service, osac-operator, osac-operator-crds, osac-aap, bare-metal-fulfillment-operator, bare-metal-fulfillment-operator-crds, csi-driver, csi-backends, osac-metering) plus the umbrella chart in `publish-osac-installer-chart.yaml`.

## Design notes

- **Digest extraction**: `helm push`'s own stdout prints `Digest: sha256:...` after a successful push. I use that directly rather than reaching for a container-image-specific tool — `skopeo inspect` explicitly refuses non-container OCI artifacts (`unsupported image-specific operation on artifact with type "application/vnd.cncf.helm.config.v1+json"`, confirmed by testing against a local registry).
- **`id-token: write` placement**: unlike the container-image workflows (which split build and sign into separate jobs so a job also running on `pull_request` never holds the OIDC permission), every chart-publish job here gets `id-token: write` added directly, no job split. These jobs only ever run via `workflow_run` after a tag-triggered image build already succeeded — never on `pull_request`-controlled content — so the least-privilege concern that justified the earlier split doesn't apply.

## What I validated locally vs. what needs a real release

- Downloaded the exact Helm version this repo pins (`v3.21.4`, via `azure/setup-helm`) and confirmed its `helm push` stdout format matches what the new composite action parses.
- Stood up a local OCI registry (`podman run registry:2`) and ran the full push → digest-parse → `cosign sign` → `cosign verify` cycle against a real Helm chart OCI artifact end-to-end, using a local cosign key pair (to avoid needing real GitHub Actions OIDC) — confirmed cosign works identically against Helm charts as it does against container images.
- YAML-validated every changed file and ran this repo's full `pre-commit` hook set (yamllint, gitleaks, etc.) — all pass.

Cannot validate outside a real release: the actual GitHub Actions OIDC token exchange, and an end-to-end run of any of these 10 jobs (they only trigger on a real tag push after their component's image build succeeds, or via `workflow_dispatch` for the umbrella chart).

## Test plan

- [ ] On the next tagged release of any covered component, confirm the corresponding `publish-*-chart` job's new "Push and sign chart" step succeeds
- [ ] `cosign verify` the resulting chart OCI artifact against `ghcr.io/osac-project/charts/<chart-name>@sha256:<digest>` and confirm it validates

## Summary

- **CI and deployment:** Added the shared `helm-push-and-sign` action. It pushes Helm charts, extracts the OCI digest, and signs the digest-pinned artifact with Cosign.
- **Authentication:** Added GitHub Actions OIDC permissions for chart publishing jobs.
- **Chart publishing:** Updated nine component chart jobs and the umbrella chart workflow to publish signed OCI artifacts.
- **Release validation:** Added tag-to-commit verification for the fulfillment service release workflow.
- **Documentation:** Added Cosign verification instructions for component and umbrella charts.
- **Tests and validation:** Covered Helm output parsing, local push/sign/verify behavior, YAML validation, and pre-commit checks. GitHub OIDC exchange and release-triggered workflow execution remain untested.
- **API and data:** No application API, controller, database, or runtime changes.

## Backward compatibility

Existing unsigned charts remain unchanged. New chart releases require Cosign signing and OIDC permissions. Consumers that verify signatures must use the documented workflow identity constraints.

## Risk classification

**risk:show** — The change affects CI authentication and release artifact publication. It adds signing, digest parsing, and tag validation without changing application runtime behavior or stored data. It does not qualify as **risk:ship** because GitHub OIDC exchange and release-triggered workflow execution were not tested end to end. It does not qualify as **risk:ask** because local end-to-end verification and configuration checks were completed.

### DIFF
--- a/.github/actions/helm-push-and-sign/action.yaml
+++ b/.github/actions/helm-push-and-sign/action.yaml
@@ -1,0 +1,36 @@
+name: Push and sign Helm chart
+description: >-
+  Push a packaged Helm chart to an OCI registry, then sign the resulting OCI
+  artifact keylessly with cosign (see .github/actions/cosign-sign-image) --
+  the calling job must grant `permissions: id-token: write`. Assumes `helm`
+  is already on PATH and already logged into the target registry.
+inputs:
+  chart-file:
+    description: Path to the packaged chart .tgz to push, relative to the repo root.
+    required: true
+  chart-name:
+    description: Chart name -- the artifact pushed to and signed is <oci-repo>/<chart-name>.
+    required: true
+  oci-repo:
+    description: OCI repo to push under, without the oci:// prefix (e.g. ghcr.io/osac-project/charts).
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: Push chart to registry
+      id: push
+      shell: bash
+      run: |
+        output=$(helm push "${{ inputs.chart-file }}" "oci://${{ inputs.oci-repo }}" 2>&1)
+        echo "${output}"
+        digest=$(grep -oE '^Digest: sha256:[0-9a-f]+' <<<"${output}" | cut -d' ' -f2)
+        if [[ -z "${digest}" ]]; then
+          echo "::error::Could not parse digest from helm push output"
+          exit 1
+        fi
+        echo "digest=${digest}" >> "$GITHUB_OUTPUT"
+
+    - name: Sign chart with cosign
+      uses: ./.github/actions/cosign-sign-image
+      with:
+        image: ${{ inputs.oci-repo }}/${{ inputs.chart-name }}@${{ steps.push.outputs.digest }}
--- a/.github/scripts/fulfillment-service-verify-tag-matches-sha.sh
+++ b/.github/scripts/fulfillment-service-verify-tag-matches-sha.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+# Verifies that a git tag still resolves to the guarded commit SHA that was
+# validated by this workflow's guard job, protecting against a force-push/
+# retag race between the image build and any subsequent chart-publish or
+# release step.
+#
+# Required env vars: GH_TOKEN, REPO, TAG, GUARDED_SHA
+set -euo pipefail
+
+ref_json="$(gh api "repos/${REPO}/git/ref/tags/${TAG}" --jq '[.object.type, .object.sha] | @tsv')"
+read -r current_type current_sha <<< "$ref_json"
+if [[ "$current_type" == "tag" ]]; then
+  # Annotated tag: the ref's object.sha is the tag object, not the commit - peel it.
+  current_sha="$(gh api "repos/${REPO}/git/tags/${current_sha}" --jq .object.sha)"
+fi
+
+if [[ "$current_sha" != "$GUARDED_SHA" ]]; then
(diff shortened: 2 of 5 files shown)


eliorerz:
Added README.md verification instructions for Helm chart signatures too (was missing from the initial commit) — a "Verifying Helm chart signatures" section covering both the per-component charts (all signed under `publish-charts.yaml`'s identity, always `refs/heads/main` since `workflow_run` always executes the workflow file from the default branch) and the umbrella chart (`publish-osac-installer-chart.yaml`, `workflow_dispatch`-only, can run from either `main` or one of its own `osac/v*` tags).

Sanity-checked the new regex against a real published image (deliberately pointed at the wrong artifact) — cosign's error output confirms the SAN identity format it extracts matches the regex template exactly, just correctly rejecting due to the workflow-file mismatch.

eliorerz:
Verified this finding against the actual checkout mechanics before changing anything.

**The premise needs a correction**: these jobs don't check out "a tag" — `ref: \${{ needs.guard.outputs.sha }}\` resolves to \`github.event.workflow_run.head_sha\`, an already-resolved 40-char commit SHA captured at the moment the triggering workflow_run event fired. Git SHAs are content-addressed and can't be retargeted the way a tag name can, so there's no TOCTOU window on the checkout itself.

**But the underlying concern (immutable release tags) is legitimate, and 8 of the 9 flagged jobs already have it**: every `publish-*-chart` job except `publish-service-chart` already runs a "Verify tag still points at the guarded commit" step calling `.github/scripts/<component>-verify-tag-matches-sha.sh` — a pre-existing guard explicitly built for "a force-push/retag race between the image build and any subsequent chart-publish or release step" (this repo's own comment, predates this PR). That's the real mitigation for the scenario the finding describes.

**Found and fixed a genuine gap**: `publish-service-chart` (fulfillment-service) was the one job missing this check — confirmed pre-existing on `upstream/main`, not introduced by this PR, but worth closing since it's the same protection every sibling job already has. Added `fulfillment-service-verify-tag-matches-sha.sh` (identical to the other components' copies) and wired it in at the same point (right after checkout).

**Skipped**: "reference the action from a separately trusted SHA-pinned source instead of" — i.e. switching `uses: ./.github/actions/helm-push-and-sign` to a `uses: osac-project/osac/.github/actions/helm-push-and-sign@<pinned-sha>`-style external reference. This isn't something specific to the 9 new call sites; every local composite action in this repo (including `cosign-sign-image`, used ~20+ times across the container-image workflows from #901, some from pull_request-triggered checkouts which are *less* trusted than a tag) and every pre-existing `.github/scripts/*.sh` invocation in this very file resolve identically from whatever's checked out. Pinning only the 9 new call sites would be inconsistent security theater relative to the rest of the file and repo, and a real architectural change (treating same-repo local actions as if they were third-party) that's out of scope for this PR to introduce unilaterally.

PR #904, opened 2026-09-11
OSAC-5107: dispatch volume provisioning by provider

zszabo-rh:
## Summary

- Add provider-aware volume provisioning across `fulfillment-service` and `osac-operator`.
- Carry optional, immutable CSI-style topology segments from the private Volume API into the operator Volume CR and vendor-neutral provisioning request.
- Preserve provider selection and status through fulfillment reconciliation, operator feedback, and regenerated metering/CSI API clients.
- Route provisioning and deletion through a provider-keyed registry, register the VAST implementation, and fail explicitly for unsupported providers.

## Context

Volume tier resolution already selected the backend and protocol, but the operator used one vendor provisioner without retaining an explicit provider identity. That made provider-specific dispatch difficult and risked sending an unsupported provider through the VAST path. This change carries the provider selected by fulfillment-service through the contract and uses it as the operator dispatch key.

## Jira

https://redhat.atlassian.net/browse/OSAC-5107

### API and contracts

- Add `VolumeTopology.segments` and optional `VolumeSpec.topology` to the private volume contract.
- Add `VolumeStatus.provider` as the provider key selected during storage-tier resolution.
- Regenerate the fulfillment-service, osac-operator, osac-metering, and osac-csi-driver private API consumers.

### Fulfillment and reconciliation

- Resolve and stamp provider status when creating a volume, with a backend fallback for existing resolution paths.
- Propagate topology into the operator Volume CR and provider status back into CR status and feedback.
- Support topology field-mask updates while enforcing topology immutability after creation.

### Operator provisioning

- Replace the single provisioner with a provider-keyed `VendorProvisionerRegistry`.
- Register the VAST provisioner from `OSAC_VENDOR_CONTROLLERS` and pass provider, backend, protocol, and topology through the vendor request.
- Keep unconfigured provisioning non-fatal, while marking a volume `ProviderNotImplemented` instead of routing an unknown provider through VAST.
- Preserve legacy Volume CR behavior by falling back to `status.backend` when `status.provider` is absent.

## Backward compatibility and failure behavior

- Topology is optional and immutable; existing clients that omit it continue to work.
- Provider status is optional for older objects, with backend fallback during reconciliation and deletion.
- An empty vendor configuration leaves provisioning disabled as before.
- An unsupported provider fails closed with an explicit status condition and does not use another provider implementation.

## Test plan

- [x] `buf generate` and generated-artifact diff checks.
- [x] Fulfillment build, lint, and `ginkgo run -r internal` (99 suites passed).
- [x] Operator format, lint, build, controller tests (882 specs passed), manifest generation, and Helm CRD checks.
- [x] CSI driver build, lint, and test suites.
- [x] Metering schema, service, and adapter tests, lint, and Helm lint.
- [x] `pre-commit run --all-files`.
- [x] Configured performance, security, and over-engineering pre-flight reviews; no blocking findings.
- [ ] CI checks and live provider/cluster integration validation.

## Validation notes

The checkout does not define the documented `osac-operator` `test-kustomize` target; the available Helm CRD validation passed. The live VAST CSI create/delete path still requires an integration environment.

---

_This PR description was drafted with AI assistance ([create-pr](https://github.com/osac-project/osac-ai-skills/tree/main/skills/create-pr) v0.2.0). Review for accuracy._

## Summary

- **API surface:** Added optional immutable volume topology and provider fields to private protobufs, CRDs, and Go types. Regenerated clients and deep-copy code.
- **Controllers:** Added provider-keyed provisioning and deletion. Propagated provider and topology through fulfillment reconciliation, operator requests, feedback, metering, and CSI flows. Unsupported providers return `ProviderNotImplemented`.
- **Deployment and configuration:** Updated CRD schemas and `vendorControllers` configuration. Registered VAST by provider. Empty or invalid provider configuration does not block startup.
- **Backward compatibility:** Legacy volumes fall back from `status.provider` to `status.backend`. Topology cannot change after creation.
- **Tests and validation:** Added coverage for provider routing, topology propagation, legacy fallback, unsupported providers, and status synchronization. Build, lint, generated-artifact, manifest, CRD, and pre-commit validations passed. CI and live provider or cluster integration validation remain pending.
- **Unaffected areas:** No database or authentication changes.

## Risk classification

**risk:show** — The change affects volume provisioning, deletion, CRDs, provider routing, and API contracts. Local validation passed, but live provider and cluster integration validation remains pending.

This does not qualify as **risk:ask** because the supplied evidence reports no database migration, authentication change, destructive data operation, or known failing validation.


fulfillment-service/internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go
```diff
@@ -1014,6 +1014,7 @@ func newDAOTierResolver(
 
 		return &servers.TierResolution{
 			Backend:  backend.GetSpec().GetProvider(),
+			Provider: backend.GetSpec().GetProvider(),
```
akshaynadkarni, line 1018: Design question: status.backend vs status.provider

Tier resolution and Create both stamp backend and provider from the same source (StorageBackend.spec.provider), and they stay identical through the Volume CR, operator VendorCreateVolumeRequest, and CSI routing (status.backend → osac.backend).

I'm concerned about carrying two status fields with the same value long term - it's unclear which is authoritative for dispatch vs CSI routing vs inventory, and we'll keep asking "should endpointFor use Backend or Provider?" (today they're the same).

Since we're pre-1.0 with no customers, would it make sense to consolidate on provider as the single routing/dispatch key for the volume pipeline?

Concretely:
- Stamp only status.provider (deprecate or stop writing status.backend)
- Operator dispatch and VAST endpointFor use provider only (drop or deprecate VendorCreateVolumeRequest.Backend)
- CSI driver reads status.provider; consider osac.provider in volume context (with osac.backend alias during transition if needed)
- Remove TierResolution.Backend and resolvedVolumeProvider() backend fallback once provider is the sole field

I don't think we need to rename unrelated "backend" concepts (StorageBackend resource, tier backend_id associations, Secret.backend, etc.) — this is specifically about Volume status routing.

Happy to approve either way, but I'd prefer we fix this before LVMS (OSAC-4360) so new code only learns one field. Is that in scope for this PR or a fast follow-up?

akshaynadkarni (changes requested):
Please check my comment [here](https://github.com/osac-project/osac/pull/904/changes#r4009664659).

fulfillment-service/internal/servers/private_volumes_server.go
```diff
@@ -269,6 +275,17 @@ func applyVolumeUpdate(base, update *privatev1.Volume, mask *fieldmaskpb.FieldMa
...
+			if topology := update.GetSpec().GetTopology(); topology != nil {
+				base.GetSpec().SetTopology(proto.Clone(topology).(*privatev1.VolumeTopology))
+			} else {
+				base.GetSpec().SetTopology(nil)
+			}
+		case "spec.topology.segments":
+			if base.GetSpec().GetTopology() == nil {
+				base.GetSpec().SetTopology(&privatev1.VolumeTopology{})
+			}
+			base.GetSpec().GetTopology().SetSegments(update.GetSpec().GetTopology().GetSegments())
```
akshaynadkarni: Small consistency nit: the `spec.topology` mask path uses `proto.Clone`, but `spec.topology.segments` assigns the update's segments map by reference (`SetSegments(update.GetSpec().GetTopology().GetSegments())`). Is this intentional? 
For e.g. buildSpec() in the volume reconciler copies segment keys into a new map instead.

Consider cloning the segments map here too so we don't share the request map with the persisted object. Immutability validation still catches real topology changes, so this is defensive rather than a functional bug.


```suggestion
        case "spec.topology.segments":
	        if base.GetSpec().GetTopology() == nil {
		        base.GetSpec().SetTopology(&privatev1.VolumeTopology{})
	        }
	        segments := update.GetSpec().GetTopology().GetSegments()
	        cloned := make(map[string]string, len(segments))
	        for k, v := range segments {
		        cloned[k] = v
	        }
            base.GetSpec().GetTopology().SetSegments(cloned)
```

zszabo-rh: done!

fulfillment-service/internal/servers/private_volumes_server.go
```diff
@@ -31,14 +31,15 @@ import (
...
-// concrete backend and protocol. Backend is the StorageBackend's provider
+// concrete backend, provider, and protocol. Backend is the StorageBackend's provider
 // (e.g. "vast"), matching Volume.status.backend's documented contract and the
 // vendor routing keys ("vast", "pure", "ontap", ...) that osac-csi-driver's
 // Helm-templated --vendor-controllers/--vendor-sockets maps use -- those maps
 // can't be keyed by the StorageBackend's server-generated id, since that id
 // doesn't exist yet when the chart values are authored.
 type TierResolution struct {
 	Backend  string
+	Provider string
```
rgolangh, line 42: Provider and Backend are effectively the same thing. Akshay has spotted related issue on the status reporting.

fulfillment-service/internal/servers/private_volumes_server.go
```diff
@@ -180,6 +181,11 @@ func (s *PrivateVolumesServer) Create(ctx context.Context,
 	vol.GetStatus().SetVendorContext(nil)
 	vol.GetStatus().SetState(privatev1.VolumeState_VOLUME_STATE_CREATING)
 	vol.GetStatus().SetBackend(resolved.Backend)
+	provider := resolved.Provider
+	if provider == "" {
+		provider = resolved.Backend
```
rgolangh, line 186: provider and backend look the same

fulfillment-service/internal/cmd/service/start/grpcserver/start_grpc_server_cmd.go
zszabo-rh: Agreed! On the long-term `status.provider` should be the single routing key and `status.backend` should be retired. But since consolidating touches every multiple backend consumers, let's handle that in a follow-up PR. I'll create a ticket as my next task.

osac-operator/cmd/main.go
```diff
@@ -640,32 +639,67 @@ func setupVolumeControllers(mgr mcmanager.Manager, grpcConn *grpc.ClientConn) er
...
-			endpoints,
+		var perr error
+		provisioners, perr = newVendorProvisionerRegistry(
+			localMgr.GetAPIReader(), configNamespace, endpoints,
 		)
 		if perr != nil {
-			setupLog.Error(perr, "vendor provisioner init failed; volume provisioning disabled")
-		} else {
-			provisioner = p
+			setupLog.Error(perr, "vendor provisioner registry init failed; volume provisioning disabled")
```
rgolangh, line 647: vendorProvisioner(s)  - missing the new plural form

osac-operator/cmd/main.go
```diff
@@ -620,13 +618,14 @@ func setupVolumeControllers(mgr mcmanager.Manager, grpcConn *grpc.ClientConn) er
...
-	// unconfigured or misconfigured vendor backend can never take down the
-	// operator or the other controllers. Most setups (including LVMS/dev) have no
-	// vendor backend configured; their Volumes stay in Progressing until one is.
-	var provisioner controller.VendorProvisioner
+	// volume provisioning disabled (an empty registry) rather than crashing, so
+	// an unconfigured or misconfigured vendor backend can never take down the
+	// operator or the other controllers. Most setups (including LVMS/dev) have
+	// no vendor backend configured; their Volumes stay in Progressing until one
+	// is.
+	var provisioners controller.VendorProvisionerRegistry
```
rgolangh, line 628: again, a better name would be seomthing that would use the 'backend' in it. 
Like
'vendorControllersRegistry'
'csiBackendsRegistry'
'vendorBackendRegistry'


I think vendorControllersRegistry is the best fit and csiBackendsRegistry the 2nd

osac-operator/cmd/main.go
```diff
@@ -640,32 +639,67 @@ func setupVolumeControllers(mgr mcmanager.Manager, grpcConn *grpc.ClientConn) er
...
+
+	// Seed every configured provider as a nil entry so that Lookup returns
+	// ProviderNotImplementedError for unsupported providers instead of
+	// silently treating the registry as empty (→ VolumePhaseFailed, not
+	// Progressing).
+	for key := range endpoints {
+		registry[key] = nil
+	}
+
+	vastEndpoint, ok := endpoints["vast"]
```
rgolangh, line 682: Don't you want to add 'lvms' endpoint or is this in a sepearte PR?

osac-operator/config/crd/bases/osac.openshift.io_volumes.yaml
```diff
@@ -98,6 +98,17 @@ spec:
                 x-kubernetes-validations:
                 - message: storageTier is immutable
                   rule: self == oldSelf
+              topology:
+                description: Topology contains optional CSI-style placement segments.
```
rgolangh: Please also describe what will be the effect on the volume and scheduling of the workfload using it and what is the requirement from the target CSI driver. Same for segments. It is not clear at the moment

osac-operator/internal/controller/vendor_provisioner_registry.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// VendorProvisionerRegistry selects a provider-specific implementation by the
+// provider resolved by the fulfillment-service.
+type VendorProvisionerRegistry map[string]VendorProvisioner
+
+// Lookup returns the implementation registered for provider. A nil or absent
+// entry is treated as unimplemented so a partially populated registry cannot
+// result in a nil-interface panic during reconciliation.
+func (r VendorProvisionerRegistry) Lookup(provider string) (VendorProvisioner, error) {
```
rgolangh, line 49: I think this could just be a the behaviour inside the controller, where is will be clearer. Too much code for such a clear behaviour - any non exiswgint provider returns the new err value.

osac-operator/internal/controller/volume_controller.go
```diff
@@ -59,21 +59,23 @@ type VendorProvisioner interface {
...
+// VendorCreateVolumeRequest carries the provider-neutral parameters needed to
+// provision a volume. Provider, Backend, and Protocol are resolved by the
+// fulfillment-service before the Volume CR is created; the operator passes
+// them through to the selected implementation without re-resolving.
 // Tenant and Tier let the vendor implementation select the per-tenant
 // credentials and construct the per-tenant/per-tier vendor resource references
 // (e.g. the VAST subsystem/view name) without re-deriving them.
 type VendorCreateVolumeRequest struct {
 	Name       string
+	Provider   string
```
rgolangh, line 71: why do we need to specidfy a provider here? we have the backend.

osac-operator/internal/controller/volume_controller.go
```diff
@@ -96,12 +98,13 @@ type VendorCreateVolumeResponse struct {
 // authenticate the deprovision call.
 type VendorDeleteVolumeRequest struct {
 	VendorVolumeID string
+	Provider       string
```
rgolangh, line 101: same as above

rgolangh (dismissed):

fulfillment-service/internal/servers/private_volumes_server.go
zszabo-rh: I know, it will be handled in a follow-up task, I've already opened [OSAC-5311](https://redhat.atlassian.net/browse/OSAC-5311) for that

osac-operator/cmd/main.go
zszabo-rh: this is just the plumbing, it will be added in an upcoming task

osac-operator/cmd/main.go
zszabo-rh: I'm convicible, but honestly he current naming looks quite clear to me.. 

Isn't "Controllers" confusing a bit? VolumeReconciler is the controller and these are the vendor-specific provisioners it delegates to. We always referenced them as "provisioners" so far.

And also for "Backend", it's the term we're retiring (OSAC-5311).

osac-operator/internal/controller/vendor_provisioner_registry.go
zszabo-rh: You are totally right on this one but I'm still not sure, as his approach also has its pros. It separates
lookup+typed-error from reconciliation so tests can inject a registry without touching the controller and both paths (create+delete) can reuse the same Lookup+ProviderNotImplementedError.

fulfillment-service/internal/servers/private_volumes_server.go
zszabo-rh: same, will be handled in OSAC-5311

osac-operator/cmd/main.go
zszabo-rh: fixed

osac-operator/config/crd/bases/osac.openshift.io_volumes.yaml
zszabo-rh: fixed

osac-operator/internal/controller/volume_controller.go
zszabo-rh: OSAC-5311

osac-operator/internal/controller/volume_controller.go
zszabo-rh: OSAC-5311

akshaynadkarni (approved):
/lgtm

osac-operator/internal/controller/volume_controller.go
```diff
@@ -298,14 +313,20 @@ func (r *VolumeReconciler) handleDelete(ctx context.Context, vol *v1alpha1.Volum
...
-		if r.VendorProvisioner == nil {
+		if len(r.VendorProvisioners) == 0 {
 			return ctrl.Result{}, fmt.Errorf(
 				"volume %q has vendorVolumeID %q but no vendor provisioner is configured; "+
 					"refusing to remove finalizer to avoid leaking the backend volume",
 				vol.Name, vol.Status.VendorVolumeID)
 		}
-		err := r.VendorProvisioner.DeleteVolume(ctx, VendorDeleteVolumeRequest{
+		provider := resolvedVolumeProvider(vol)
+		provisioner, err := r.VendorProvisioners.Lookup(provider)
```
rgolangh: Lets settle on 'provisioner'  and not 'provider'

Volume provisioner is also part of the CSI teminology. Provider is not used anywhere.

Also regarding the code here - it is clear that 'provider' variable here is the provisionerName. I expect the code to be something like:

```
provisioner := r.VendorProvisioners.Lookup(vol.Status.Provisioner)

rgolangh:
dismiss

PR #924, opened 2026-09-12
NO-ISSUE: Fix auto-queue re-enqueue gap and hardcoded --rebase

eliorerz:
## Summary

- Found while diagnosing PR #923: it sat `mergeStateStatus=CLEAN` with auto-merge enabled but never actually entered the merge queue after the `ci-status-checks` ruleset was fixed to drop a stale required check. GitHub only enqueues a PR on specific triggers (new commit, review, label, or an explicit `enable-auto-merge` call) — not on a passive change to *why* a PR became mergeable. `requeue-eligible-prs` only scans PRs with `autoMergeRequest == null`, so a PR that already had auto-merge on but fell out of the queue this way was invisible to the one automation meant to catch stuck PRs. Added a step that scans CLEAN + mergeable + auto-merge-enabled PRs for a missing `mergeQueueEntry` and force-enqueues via the `enqueuePullRequest` mutation (the same call GitHub's own "Merge when ready" button makes).
- Both `gh pr merge --auto --rebase` call sites hardcode a merge strategy that conflicts with the `merge_queue` ruleset's configured method (SQUASH) once a PR already has auto-merge enabled — confirmed live: `gh pr merge 923 --auto --rebase` fails outright with "the merge strategy for main is set by the merge queue", while `gh pr merge 923 --auto` (no strategy) succeeds. This silently broke the re-enable path in `requeue-eligible-prs`. Dropped `--rebase` from both call sites.

## Test plan

- [x] YAML validated
- [x] `actionlint` clean
- [x] `enqueuePullRequest` mutation verified live against PR #923 (moved it from `mergeQueueEntry: null` to `QUEUED` position 1)
- [x] `gh pr merge --auto` (no strategy) verified live to succeed where `--auto --rebase` failed, on the same PR
- [ ] End-to-end: a future ruleset change stranding a PR gets picked up by the new step within one 10-minute cron cycle

## Affected areas

- **CI:** The scheduled workflow re-enqueues clean, mergeable, auto-merge-enabled pull requests that lack a `mergeQueueEntry`.
- **API surface:** The workflow uses GitHub’s `enqueuePullRequest` GraphQL mutation.
- **Merge behavior:** `gh pr merge --auto` no longer forces `--rebase`. The configured merge-queue strategy applies.
- **Reliability:** Transient `gh api` lookup failures affect only the current pull request.
- **Tests and validation:** YAML and `actionlint` validation passed. The enqueue mutation and merge command were tested against PR `#923`. End-to-end validation remains pending after the ruleset change.
- **Other areas:** No controller, database, authentication, deployment, or documentation changes.

## Compatibility and risk

Existing pull requests can enter the merge queue through the recovery path. The workflow rechecks blocking labels and human `CHANGES_REQUESTED` reviews before enqueueing. Merge behavior can change because the repository ruleset now controls the strategy instead of the hardcoded `--rebase` option. No public entity or external API contract changed.

## Risk classification

**risk:ship** — The change is limited to CI automation, includes focused validation, preserves blocking checks, and isolates transient lookup failures. It does not meet **risk:show** or **risk:ask** criteria because it does not change application runtime behavior, data storage, authentication, or public interfaces. End-to-end validation remains pending, but this does not raise the change to a higher risk class.

### DIFF
--- a/.github/workflows/auto-queue.yml
+++ b/.github/workflows/auto-queue.yml
@@ -144,7 +144,12 @@ jobs:
           GH_TOKEN: ${{ secrets.MERGE_QUEUE_TOKEN }}
           PR: ${{ github.event.pull_request.number }}
           REPO: ${{ github.repository }}
-        run: gh pr merge "$PR" --repo "$REPO" --auto --rebase
+        # No explicit strategy: the merge_queue ruleset on main already fixes
+        # the method (SQUASH). Passing --rebase here fails outright once a PR
+        # already has auto-merge enabled ("the merge strategy for main is set
+        # by the merge queue") -- harmless on first enable, but breaks the
+        # re-enable path used by requeue-eligible-prs below.
+        run: gh pr merge "$PR" --repo "$REPO" --auto
 
       - name: Disable auto-merge
         if: >-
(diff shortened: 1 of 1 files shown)


eliorerz:
Both findings verified against the actual repo config and are real — applied both, not blindly:

**Lines 255-262 (labels/review reapplication):** Confirmed via \`gh api repos/osac-project/osac/rulesets/20134747\` that the \`pull_request\` ruleset rule has \`required_approving_review_count: 0\`. That means \`mergeStateStatus: CLEAN\` does **not** encode anything about outstanding human review state or \`do-not-merge/*\` labels here — those are enforced only by this bot's own reactive \`Disable auto-merge\` step elsewhere in this file, which could in principle miss an event (token hiccup, a label added in the same instant the cron runs, etc). Fixed by reapplying the same required/blocking-label check and human-\`CHANGES_REQUESTED\` check used by the other two paths in this file before force-queueing, rather than relying solely on \`mergeable\`/\`mergeStateStatus\`.

**Lines 267-274 (failure handling under set -e):** Confirmed the \`already_queued\` lookup had no failure tolerance, unlike the enqueue call right below it. A single transient \`gh api\` failure for one PR would have aborted the whole batch via \`set -e\` instead of skipping just that PR. Fixed with the same \`if ! var=$(...); then ... continue; fi\` pattern.

Pushed as 60b7e0042. actionlint + YAML validated clean.

PR #926, opened 2026-09-12
NO-ISSUE: Cancel in-flight e2e work and dismiss orphan gates on PR close

eliorerz:
## Summary

Found while diagnosing why PR #550's e2e gate checks stayed `BLOCKED` for hours despite the real gate job succeeding: an `invalidate-e2e-gates` API placeholder (`in_progress`, unlock-time) was left permanently unresolved because nothing triggers cleanup unless a *new* push happens. `complete-stale-e2e-gates` (osac-test-infra #542, already merged) fixed the general case by running follow-up cleanup jobs right after each native gate job — but nothing handles the case where a PR closes (merged or not) while work is still in flight or an orphan is still sitting `in_progress`. That's a separate gap from #542, not a duplicate of it.

This adds a `pull_request_target: closed` handler that:
- Cancels every in-flight full-install run for the PR, via a new `CANCEL_ALL` mode on `cancel-stale-e2e-runs.sh` (the existing script explicitly skips the current head SHA by design, since it's built for the synchronize/obsolete-SHA path — this reuses it rather than duplicating it)
- Dismisses any remaining `in_progress` `e2e-*-gate` orphan checks, reusing `complete-stale-e2e-gates`'s existing `after-native-success: false` path as-is — no changes needed in `osac-test-infra`, that mode already does exactly this unconditionally

## Test plan

- [x] `bash -n` on the modified script
- [x] Verified the new `CANCEL_ALL` jq filter locally against mock run data: `false` mode still only matches obsolete SHAs (unchanged behavior), `true` mode matches all SHAs
- [ ] End-to-end: close/merge a PR with e2e still in flight and confirm the runs get cancelled and no orphan checks remain

## Summary

- **CI:** Added closed pull request cleanup for `main`.
- **CI:** Added a live-state guard before cleanup.
- **CI:** Added `CANCEL_ALL` mode to cancel matching full-install E2E runs for the exact pull request.
- **CI:** Dismisses remaining `in_progress` E2E gate checks.
- **Security:** Uses base-repository execution and scoped write permissions.
- **Validation:** Added YAML, shell syntax, and jq-filter validation. End-to-end testing remains pending.
- **API, controllers, database, authentication, deployment, and documentation:** No changes.

## Backward compatibility

No application or public API behavior changes. The change affects CI cleanup only. A reopened pull request is protected from delayed cleanup runs.

## Risk classification

**risk:ship** — The change is limited to CI automation. It uses exact pull request matching, checks the current pull request state, and uses scoped permissions. It does not affect production runtime behavior, data, authentication, or public APIs.

The change is close to **risk:show** because it changes visible CI behavior. It does not qualify because the impact is limited to pull request cleanup.

### DIFF
--- a/.github/scripts/cancel-stale-e2e-runs.sh
+++ b/.github/scripts/cancel-stale-e2e-runs.sh
@@ -21,6 +21,15 @@ if [[ -z "${REPO:-}" || -z "${HEAD_SHA:-}" || -z "${HEAD_BRANCH:-}" || -z "${HEA
   exit 1
 fi
 
+# CANCEL_ALL drops the "differs from HEAD_SHA" scoping below, so it leans on
+# repo+branch alone to identify "this PR's runs" -- not enough on its own if a
+# branch name is ever reused across PRs (delete-and-recreate, or a stale run
+# from a previous close/reopen cycle). Require an exact PR_NUMBER match too.
+if [[ "${CANCEL_ALL}" == "true" && -z "${PR_NUMBER:-}" ]]; then
+  echo "PR_NUMBER is required when CANCEL_ALL=true" >&2
+  exit 1
+fi
+
 # Return 0 when PR head still matches HEAD_SHA; 1 when moved or lookup failed.
 # Always 0 in CANCEL_ALL mode: a closed PR's head won't move, and there's no
 # newer push for this cleanup to defer to.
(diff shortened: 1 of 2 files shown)


eliorerz:
Two of three findings verified as real and fixed, one skipped:

**Line 82 (cancel-stale-e2e-runs.sh) — valid, fixed.** CANCEL_ALL drops the "differs from HEAD_SHA" scoping, leaving repo+branch as the only match criteria for which runs to cancel. Not sufficient on its own if a branch name is ever reused across PRs (delete-and-recreate, or a stale run surviving a prior close/reopen cycle). Now also requires an exact match against `PR_NUMBER` in each run's `pull_requests` list when `CANCEL_ALL` is set. Tested locally against mock run data: a same-branch run belonging to a different PR number is now correctly excluded, and the non-`CANCEL_ALL` path (used by the existing synchronize handler) is unaffected.

**Line 34 (e2e-cancel-on-pr-close.yml) — valid, fixed.** Added a `guard` job (`pull-requests: read`) that re-checks live PR state before either cleanup job runs, covering the case where a queued/delayed `pull_request_target` run fires after the PR was reopened. Both `cancel-in-flight-runs` and `dismiss-orphan-gates` now `needs: guard` and only run when it confirms the PR is still closed.

**Line 42 (pin to commit SHA) — skipped.** Checked the rest of the repo first: `e2e-cancel-stale-runs-on-push.yml` and all three `e2e-*-full-install.yml` files already reference this exact same `complete-stale-e2e-gates@main` action, unpinned, and those are already merged. This is an established convention here, not something this PR introduced — pinning only this new caller would just be inconsistent with the other four without a broader team decision to pin all of them.

Pushed as 4d860a755. YAML/actionlint/`bash -n` all clean, jq filter changes verified against mock data before pushing.

PR #935, opened 2026-09-13
NO-ISSUE: Fix complete-stale-e2e-gate-checks.sh re-blocking good PRs

eliorerz:
## Urgent -- currently re-blocking a good PR (#932)

Found live: an earlier \`MODE=dismiss\` run against #932's head SHA marked its \`e2e-*-gate\` orphan checks "cancelled" hours after the real native gate jobs had already completed "success". Since required-status-check evaluation uses the most recent check-run per name, the newly-cancelled orphan became the "latest" entry and re-blocked the PR. Confirmed via \`enqueuePullRequest\`'s own error:

\`\`\`
Pull request has failing required statuses and Pull request 3 of 26 required status checks are cancelled.
\`\`\`

Root cause: \`orphan_ids_for_gate\` only ever matched \`status=="in_progress"\`, so once dismiss mode had (wrongly) finalized an orphan as cancelled, neither mode could touch it again to fix it -- no way to undo the mistake.

## Fix

Reworks both modes around the native gate job's actual conclusion (success/skipped/pending/missing/failure) instead of a fixed target conclusion:
- \`MODE=complete\` mirrors the orphan onto the native result whenever it's \`success\` or \`skipped\` (never \`failure\` -- this script should never be what turns a genuinely broken PR green), and now matches orphans regardless of their current status, so it can correct a previously mis-cancelled one.
- \`MODE=dismiss\` only touches an orphan when the native gate has **not** posted a terminal result yet. Once a native result exists, cancelling the orphan is never safe -- that's exactly what caused this bug -- so dismiss now defers to complete instead.

## Test plan

- [x] \`bash -n\` clean
- [x] Verified the new \`native_gate_conclusion\`/\`orphan_ids_for_gate\` logic locally against a mock of #932's exact scenario (a success native result plus a later-cancelled orphan): correctly identifies the mis-cancelled orphan for correction under \`MODE=complete\`
- [ ] Run against #932 once merged to confirm it actually unblocks

## Summary

- **CI:** Updated `.github/scripts/complete-stale-e2e-gate-checks.sh`.
- `MODE=complete` mirrors only `success` or `skipped` from the native gate.
- Completion mode detects orphan checks in any status, including cancelled checks.
- `MODE=dismiss` cancels an orphan only when the native gate has no terminal result.
- The script refreshes check-run data before each orphan decision.
- Pagination and conclusion handling use reusable functions.

## Compatibility and verification

- No public API, controller, database, authentication, deployment, documentation, or test interface changed.
- Valid pull requests should no longer be re-blocked after orphan checks are cancelled.
- `bash -n` passes.
- Local testing reproduces the reported scenario.
- Verification against PR `#932` after merge remains pending.
- No Jira issue is referenced.

## Risk classification

**risk:ship** — The change is limited to a CI maintenance script, preserves native gate conclusions, and has bounded scope. It does not qualify for **risk:show** because it does not introduce a user-facing feature or notable operational change. It does not qualify for **risk:ask** because it has no unresolved security, data, or deployment risk.

### DIFF
--- a/.github/scripts/complete-stale-e2e-gate-checks.sh
+++ b/.github/scripts/complete-stale-e2e-gate-checks.sh
@@ -1,7 +1,22 @@
 #!/usr/bin/env bash
-# Manual cleanup for orphaned in_progress e2e-*-gate API checks on HEAD_SHA.
-# MODE=complete: mark orphans success when native gate job already succeeded.
-# MODE=dismiss: mark unlock-time orphans cancelled (mislabeled /runs/<id> checks).
+# Manual cleanup for orphaned e2e-*-gate API checks on HEAD_SHA -- either
+# still in_progress, or already (mis-)finalized by an earlier run of this
+# same script.
+#
+# MODE=complete: mirror the orphan onto whatever the native gate job's real,
+#   terminal conclusion actually is (success or skipped). Never touches a
+#   native failure: leaving the orphan as-is or matching a real failure both
+#   correctly keep the PR blocked, and this script should never be what
+#   turns a genuinely broken PR green.
+# MODE=dismiss: cancel unlock-time orphans, but ONLY when the native gate
+#   job has not posted a terminal result yet. Cancelling an orphan whose
+#   native gate already completed can make that cancellation the *latest*
+#   check-run for the name (required-status-check evaluation uses the most
+#   recent entry per name), silently re-blocking a PR that was actually
+#   fine -- confirmed this happening in practice: an earlier run of this
+#   script in dismiss mode did exactly that after the real gate had already
+#   succeeded hours earlier. Use MODE=complete instead once a native result
+#   exists.
 
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


eliorerz:
Valid and fixed. Confirmed on re-reading my own script: the per-orphan "re-check" inside the loop called \`native_gate_conclusion\`, but that read the same \`check_runs\` snapshot fetched once at the top of the script — it looked like a freshness guard but never actually queried GitHub again, so it couldn't catch anything that changed mid-run. Given this script exists specifically to fix a staleness-driven bug, worth doing properly.

Extracted the fetch into \`fetch_check_runs()\` and call it fresh immediately before each per-orphan conclusion decision, in both \`MODE=complete\` and \`MODE=dismiss\` (the same race applies symmetrically to both). As noted in the comment I added, this narrows but can't fully close the GET-to-PATCH gap for a single orphan — that's inherent to using two separate API calls, not something this script can eliminate.

Pushed as 2e55f0931. \`bash -n\` clean, re-verified \`native_gate_conclusion\`'s logic is unchanged (now takes the runs snapshot as an explicit arg instead of a global) against the same mock data used for the previous fix.

PR #956, opened 2026-09-14
OSAC-4214: preserve metadata on unmasked instance type updates

ibengal:
## Summary

Instance type updates without an update mask previously merged the requested spec but retained sparse client
metadata on the update object. This could erase existing labels, annotations, and finalizers.

This change passes the complete merged instance type to the generic update while preserving the client-
provided version for optimistic locking.

Fixes https://redhat.atlassian.net/browse/OSAC-4214

## Changes

- private_instance_types_server.go - Uses the complete merged object for unmasked updates while retaining the client version
- private_instance_types_server_test.go - Adds coverage for sparse metadata and omitted mutable metadata during
unmasked updates

## Test plan
- [x] `go test ./internal/servers`
- [x] Verified unmasked updates preserve existing labels, annotations, and finalizers when omitted by the
client.
- [x] Verified unmasked updates retain the client version for optimistic locking.

_Developed with AI assistance (Codex). Review for accuracy._

## Summary

- **API surface:** Updated unmasked instance-type updates to pass the fully merged instance type to the generic update operation. The client-provided metadata version remains in use for optimistic locking.
- **Data behavior:** Sparse requests now preserve existing labels, annotations, and finalizers while applying mutable specification changes.
- **Tests:** Added coverage for metadata preservation and version retention.
- **Other areas:** No changes affect controllers, database storage, authentication, deployment, CI, or documentation.

## Backward compatibility

This change is backward compatible. It corrects sparse update behavior and preserves metadata that was previously at risk of being removed.

## Risk classification

**risk:ship** — The change is small, isolated to instance-type update handling, preserves optimistic locking, adds focused tests, and does not change public declarations or security, database, deployment, or authentication behavior.

The change was close to **risk:show** because it alters update semantics, but it does not qualify because the new behavior preserves existing state and has focused test coverage.


fulfillment-service/internal/servers/private_instance_types_server.go
```diff
@@ -202,8 +202,10 @@ func (s *PrivateInstanceTypesServer) Update(ctx context.Context,
 		mask.Paths = append(mask.Paths, "spec.deprecation")
 	}
 
-	// Set the merged spec back into the request for the generic update:
-	request.GetObject().SetSpec(merged.GetSpec())
+	// set merged object back into request for the generic update
+	clientVersion := request.GetObject().GetMetadata().GetVersion()
+	request.SetObject(merged)
```
ygalblum, line 207: I understand this change so that the request hold the entire merged object and not just update the spec field. But, I don't understand why we need to separately get the version from the request and set it. Why isn't it in the merged one to begin with?

slintes: This rings a bell, I used this pattern for diskimages, because without some tests failed. Let my AI friend explain it:

>  The version can't just be "in the merged object to begin with" because the merged object's version comes from the
>  wrong source. There are two competing constraints that pull in opposite directions:
> 
>  1. Why SetObject(merged) and not SetSpec(merged.GetSpec()) (OSAC-3718):
>     GenericServer's no-mask Update path does tmpObject = requestObject and uses the request object as-is. If the client
>     sends a sparse object (e.g. missing metadata.name), the DB check_immutable_columns trigger sees the name changing
>     to empty and rejects the update. So the request must carry the complete metadata from the DB-sourced clone →
>     SetObject(merged).
> 
>  2. Why the version must then be restored (OSAC-3713):
>     merged was built from our own earlier s.generic.Get() fetch, so its metadata.version is the current DB version, not
>     the client's. Optimistic locking works by comparing the client-supplied version against a fresh DB fetch. If you
>     leave merged's version in place, that comparison is DB-version-vs-DB-version — always equal — so the lock is
>     silently bypassed and stale writes go through undetected.
> 
>  So: SetObject(merged) is required to survive the no-mask path with complete metadata, but it clobbers the client's
>  version with the DB's. Saving clientVersion before and restoring it after gives you both: complete metadata and a
>  working optimistic-lock check. That's why the version is handled separately rather than living in merged — the merged
>  object structurally has the wrong version by construction.


Does that make sense?

ygalblum: Thanks for the clarification

ygalblum (approved):
/approve
/lgtm

PR #960, opened 2026-09-14
OSAC-5240: implement BMaaS metering lifecycle and heartbeats

amito:
## Summary

This PR implements the BMaaS metering flow across the Epic 1 tasks:

- [OSAC-5243](https://redhat.atlassian.net/browse/OSAC-5243): map `BareMetalInstance` Watch payloads to stable resource and billing dimensions.
- [OSAC-5244](https://redhat.atlassian.net/browse/OSAC-5244): track independent allocation and consumption meter state and emit meter-specific lifecycle events.
- [OSAC-5245](https://redhat.atlassian.net/browse/OSAC-5245): generate state-aware allocation and consumption heartbeat events.

The branch also includes the local and Masayag review fixes for safe Watch handling, projection preservation, and initial BMaaS lifecycle publication.

## What changed

- Recognize and map BareMetalInstance Watch events with stable resource, tenant, project, state, catalog-item, and `bm_instance_type` dimensions.
- Reject missing or invalid immutable instance-type IDs before publishing incorrectly attributed billable events or advancing projections.
- Maintain independent allocation and consumption meter intervals across start, suspend, resume, stop, recovery, and deletion transitions.
- Use the authoritative deletion event timestamp when closing active meter intervals; skip malformed deletion events safely without deleting the projection.
- Publish `created.v1` plus allocation and consumption `started.v1` events when an initially observed BMaaS object is already `RUNNING`.
- Suppress heartbeats when the corresponding meter interval has not started, while preserving meter-specific dimensions, stable IDs, and elapsed durations.
- Preserve allocation billability and component-ever-started state across skipped or partial updates.
- Skip BMaaS missed-deletion cases without requiring a BMI List client.

## Commit organization

1. [OSAC-5243](https://redhat.atlassian.net/browse/OSAC-5243) — BMaaS Watch mapping, dimension plumbing, and contract coverage.
2. [OSAC-5244](https://redhat.atlassian.net/browse/OSAC-5244) — event timestamp prerequisite, transition matrix, persistent meter state, lifecycle decomposition, and transition coverage.
3. [OSAC-5245](https://redhat.atlassian.net/browse/OSAC-5245) — independent meter heartbeats, elapsed durations, shared fan-out, and reconciliation coverage.
4. [OSAC-5240](https://redhat.atlassian.net/browse/OSAC-5240) — local review fixes for safe BMaaS Watch and reconciliation handling.
5. [OSAC-5244](https://redhat.atlassian.net/browse/OSAC-5244) — follow-up fixes preserving meter state and publishing initial RUNNING lifecycle events.

## Validation

- `make test` from `osac-metering/` — passed: schema, metering-service, database-backed projection, Watch, heartbeat, reconciliation, and adapter suites.
- `make lint` from `osac-metering/` — passed: Helm, schema, metering-service, and adapter lint.
- Targeted BMaaS unit tests — passed.
- Pre-commit checks for changed files — all passed except the `osac-metering-golangci-lint` hook, which could not analyze Go files when invoked from the repository root; the component `make lint` passed independently.

_This PR description was drafted with AI assistance. Review for accuracy._

## Summary

- **API and event model:** Added BMaaS resource mapping, billing dimensions, lifecycle events, allocation and consumption meters, transition resolution, and duration-aware heartbeat events.
- **Watch and reconciliation:** Added BMaaS Watch handling, safe validation, duplicate-transition suppression, deletion handling, initial lifecycle publication, and synthetic heartbeat reuse.
- **Projection and database:** Added independent meter state persistence through the `component_ever_started` JSONB column. Existing databases receive an idempotent schema upgrade.
- **Configuration:** Added the `enableBMaaS` filter flag. BMaaS remains disabled when no enablement flags are set.
- **Tests:** Added coverage for BMaaS lifecycle transitions, invalid data, missing timestamps, independent meter state, reconciliation, database persistence, and configuration defaults.
- **Compatibility:** Existing deployments require the additive database migration. `watch.BuildFilter` now requires a third argument. BMaaS event processing is opt-in by default. Invalid instance-type IDs no longer produce billable events or advance projections.

Test and lint results are only partially supported: targeted BMaaS tests and component lint passed. The root Go lint hook could not analyze Go files. Review severity counts are unavailable.

## Risk classification

**risk:show** — applied because the change affects billing event generation, lifecycle state transitions, Watch processing, reconciliation, and persistent database state.

The change was close to **risk:ask** because it modifies multiple billing and persistence paths. No supplied labeling criteria define when that label is required, so a stricter classification cannot be confirmed from the available evidence.


osac-metering/metering-service/internal/watch/consumer.go
```diff
@@ -332,7 +335,11 @@ func (c *Consumer) publishLifecycleEvents(ctx context.Context, baseCE *cloudeven
 		return c.publishWithRetry(ctx, baseCE)
 	}
 
-	decomposed, err := events.BuildResourceEvents(mapper.ResourceType(), mapper.BillingDimensionsMap(), eventID, func(dims map[string]any, compEventID string) (cloudevents.Event, error) {
+	dims, err := mapper.BillingDimensionsMap()
+	if err != nil {
+		return err
+	}
```
masayag: Should this be removed? It overrides the initialization of dims from line 153

osac-metering/metering-service/internal/events/bare_metal_instance.go
```diff
@@ -0,0 +1,113 @@
...
+	if spec := m.instance.GetSpec(); spec != nil {
+		if template := spec.GetTemplate(); template != nil {
+			return NilIfEmpty(template.GetName())
+		}
+	}
+	return nil
+}
+
+func (m *bareMetalInstanceMapper) CurrentState() string {
+	state := privatev1.BareMetalInstanceState_BARE_METAL_INSTANCE_STATE_UNSPECIFIED
```
masayag, line 61: Don't we prefer using the the values for the BM instance status from the API instead? https://github.com/osac-project/osac/blob/e851ccf2bff1879117d75c35e9fa10b5cc59b2f3/proto/private/osac/private/v1/baremetal_instance_type.proto#L216

osac-metering/metering-service/internal/events/bare_metal_instance.go
```diff
@@ -0,0 +1,113 @@
...
+func (m *bareMetalInstanceMapper) IsBillable() bool {
+	return m.CurrentState() == BareMetalInstanceStateRunning
+}
+
+func (m *bareMetalInstanceMapper) BillingDimensionsMap() (map[string]any, error) {
+	return BareMetalInstanceBillingDimensions(m.instance)
+}
+
+func BareMetalInstanceBillingDimensions(bmi *privatev1.BareMetalInstance) (map[string]any, error) {
+	spec := bmi.GetSpec()
```
masayag, line 77: `meter_type` is expected to be one of the billing dimension (with values `allocation`/`consumption`). shouldn't they be included here as well, according to the BM instance state?

osac-metering/metering-service/internal/watch/consumer.go
amito: This is a different function scope - `dims` is a local variable in each scope.

osac-metering/metering-service/internal/events/bare_metal_instance.go
amito: Please clarify? This value is from the API already.

osac-metering/metering-service/internal/events/bare_metal_instance.go
amito: The current metering contract does not support `meter_type`.
Per the task separation, [OSAC-5244](https://redhat.atlassian.net/browse/OSAC-5244) will add support for consumption/allocation. If you think that the scope of the two should be covered here, I can unify the two to this PR.

osac-metering/metering-service/internal/watch/consumer.go
```diff
@@ -332,7 +335,11 @@ func (c *Consumer) publishLifecycleEvents(ctx context.Context, baseCE *cloudeven
...
+	decomposed, err := events.BuildResourceEvents(mapper.ResourceType(), dims, eventID, func(dims map[string]any, compEventID string) (cloudevents.Event, error) {
```
masayag: ah, line 333 was collapsed so I missed the function declaration. sorry.

osac-metering/metering-service/internal/events/bare_metal_instance.go
masayag: in this function we read the state that its type is the official one, but then we remove the prefix from it, so it becomes `Running` which is clear and concise, but de-factor creates a new values exposed by the metering system to its client. The alternative is using the values from the API directly (e.g. BareMetalInstanceStateRunning) without converting them.

osac-metering/metering-service/internal/events/bare_metal_instance.go
masayag: Yes, I think we can skip the intermediate step and have the meter-type introduced in this PR.
But this is optional from my perspective. I can live with a follow-up PR for this.

osac-metering/metering-service/internal/watch/consumer.go
```diff
@@ -346,6 +360,281 @@ func (c *Consumer) publishLifecycleEvents(ctx context.Context, baseCE *cloudeven
...
+	)
+
+	if event.GetType() == privatev1.EventType_EVENT_TYPE_OBJECT_CREATED {
+		created, err := events.MapWatchEvent(event, mapper, &events.StateContext{}, dims)
+		if err != nil {
+			return err
+		}
+		return c.publishAndUpsert(ctx, func() error {
+			return c.publishWithRetry(ctx, created)
+		}, projectionState, resourceID)
```
masayag: For an initial OBJECT_CREATED snapshot whose state is RUNNING, or another allocation-billable state, allocationEffect and consumptionEffect are start and buildBareMetalProjectionState opens the intervals, but this branch publishes only created.v1. No allocation or consumption started.v1 events are emitted, and ComponentEverStarted is already set, so the missing boundary cannot be repaired by a later transition. Publish the generic audit event plus the lifecycle meter events before upserting.

osac-metering/metering-service/cmd/metering-service/main.go
```diff
@@ -272,7 +272,7 @@ func run(ctx context.Context, logger logr.Logger, cfg *config) error {
 
 	eventsClient := privatev1.NewEventsClient(grpcConn)
 	consumer := watch.NewConsumer(eventsClient, publisher, store, logger)
-	consumer.Filter = watch.BuildFilter(cfg.enableVMaaS, cfg.enableCaaS)
+	consumer.Filter = watch.BuildFilter(cfg.enableVMaaS, cfg.enableCaaS, cfg.enableBMaaS)
```
masayag, line 286: BMaaS is added to the Watch filter here, but run still constructs Reconciler with only computeClient and clusterClient. loadFulfillmentState therefore never includes BMaaS; reconcileMissedDeletions treats every BMI projection as absent and deletes it, while emitting missed_deletion, at startup and every periodic reconcile. Please add BMaaS List reconciliation and billability handling, or exclude BMaaS from reconciliation until that exists.

osac-metering/metering-service/internal/watch/consumer.go
```diff
@@ -346,6 +360,281 @@ func (c *Consumer) publishLifecycleEvents(ctx context.Context, baseCE *cloudeven
...
+	}
+
+	if events.IsAllocationBillableState(state.CurrentState) {
+		if state.BillableSince == nil {
+			now := transitionTime.UTC()
+			state.BillableSince = &now
+		}
+	} else {
+		state.BillableSince = nil
+	}
```
masayag: The transition table explicitly returns BMaaSEffectSkip for PROVISIONING->DELETING and FAILED->DELETING, but this state-based branch creates BillableSince whenever the destination is allocation-billable. That opens an allocation interval without a start event, and violates the no-interval requirement for FAILED, causing later heartbeat or deletion durations to charge an unstarted interval. Update BillableSince based on allocationEffect, preserving only an existing interval on skip.

osac-metering/metering-service/internal/projection/postgres.go
```diff
@@ -103,6 +107,7 @@ func (s *PostgresStore) Upsert(ctx context.Context, state ResourceState) error {
 			fulfillment_version = EXCLUDED.fulfillment_version,
 			billing_dimensions = EXCLUDED.billing_dimensions,
 			component_billable_since = EXCLUDED.component_billable_since,
+			component_ever_started = EXCLUDED.component_ever_started,
```
masayag: component_ever_started is an ever flag, but this update replaces the stored JSON with EXCLUDED. The SELECT FOR UPDATE only checks fulfillment_version; an accepted same-or-newer-version upsert built from an older projection can erase true flags and make a later resume emit started.v1. Merge stored and incoming component flags with logical OR, and add a regression test.

masayag (changes requested):
Request changes. The inline comments identify four production issues: BMaaS projections are deleted by reconciliation, billable create snapshots omit meter-start events, allocation intervals open on skipped transitions, and component ever-started state is not monotonic under stale upserts.

osac-metering/metering-service/internal/watch/consumer.go
```diff
@@ -346,6 +372,304 @@ func (c *Consumer) publishLifecycleEvents(ctx context.Context, baseCE *cloudeven
...
+
+	if events.IsConsumptionBillableState(state.CurrentState) {
+		if _, ok := state.ComponentBillableSince[events.BMaaSMeterConsumption]; !ok {
+			now := transitionTime.UTC()
+			state.ComponentBillableSince[events.BMaaSMeterConsumption] = now
+		}
+	} else {
+		delete(state.ComponentBillableSince, events.BMaaSMeterConsumption)
+	}
```
masayag: Consumption interval updates still follow `IsConsumptionBillableState` instead of `consumptionEffect`. Allocation was switched to start / preserve / clear from the effect, including skip. Lifecycle emission is already effect-gated, so this branch can open `ComponentBillableSince["consumption"]` on `BMaaSEffectSkip` with no `started.v1` / `resumed.v1`. Heartbeats will then charge that interval.

Gate this the same way as allocation:

- `BMaaSEffectStart` / `BMaaSEffectResume`: open the consumption interval if missing
- `BMaaSEffectSuspend`: clear it
- `BMaaSEffectSkip` / default: preserve an existing interval; do not invent one

Add a Watch regression: RUNNING projection with `BillableSince` set, consumption key absent, and dims changed so `shouldSkipUpdate` does not swallow the event. Assert no consumption interval and no start event.

The Reconciler-created RUNNING projection path is not reachable on this tip (no BMI List client), but this Watch branch is still the landmine.

masayag (changes requested):
## Review (tip 3b093faf)

Prior change-request items are fixed on this tip:

- BMaaS projections are no longer deleted by reconciliation without a BMI List client
- CREATE on a billable RUNNING snapshot publishes `created.v1` plus allocation and consumption `started.v1`
- Allocation intervals follow `allocationEffect`, including FAILED/PROVISIONING → DELETING skip
- `component_ever_started` is merged monotonically on upsert

One remaining Watch bug should be fixed before merge. Consumption interval updates are still state-based and can open `ComponentBillableSince["consumption"]` on `BMaaSEffectSkip` with no matching start/resume event. Gate consumption the same way as allocation (inline comment).

Missed BMaaS deletion remains an overbill gap until a BMI List client exists. Keep skipping missed deletion in this PR; do not route a follow-up List path through `BuildResourceEvents` or `NextComponentBillableSince`.

osac-metering/metering-service/internal/projection/types.go
```diff
@@ -33,8 +33,12 @@ type ResourceState struct {
...
 	// resource types that don't decompose into components.
 	ComponentBillableSince map[string]time.Time
-	LastHeartbeatAt        *time.Time
-	TransitionTime         time.Time
-	FulfillmentVersion     int32
-	BillingDimensions      map[string]any
+	// ComponentEverStarted records whether each independently metered component
+	// has ever opened an interval. BMaaS uses it to distinguish started.v1 from
+	// resumed.v1 after consumption suspension.
+	ComponentEverStarted map[string]bool
```
masayag: ComponentEverStarted is carrying durable lifecycle history, but a generic map[string]bool makes the meter state implicit and has already required custom merge logic in PostgresStore.Upsert. Could we model each meter explicitly with active_since and first_started_at, or use a normalized (resource_id, meter_type) state row? Then first_started_at == NULL unambiguously selects started.v1 versus resumed.v1, while active_since controls heartbeat and closure. This also gives us a place for per-meter version and event identity, instead of coupling allocation to BillableSince and consumption to a JSON map. Is there a reason to prefer the current map over explicit per-meter state?

osac-metering/schema/resource.go
```diff
@@ -11,6 +11,7 @@ package schema
 
 // Resource type constants identify OSAC resource types in CloudEvent payloads.
 const (
-	ResourceTypeComputeInstance = "compute_instance"
-	ResourceTypeClusterOrder    = "cluster_order"
+	ResourceTypeComputeInstance   = "compute_instance"
+	ResourceTypeClusterOrder      = "cluster_order"
+	ResourceTypeBareMetalInstance = "bare_metal_instance"
```
masayag, line 18: This introduces a producer resource type, but the existing M360 adapter has no bare_metal_instance entry in adapters/cmd/m360-adapter/translate.go. The adapter consumes all topics by default, so each new BMaaS lifecycle, heartbeat, or correction event is rejected as an unknown resource_type and then DLQ’d or dropped as non-retryable. If M360 is part of the supported metering path, add the endpoint and contract with tests, or explicitly gate BMaaS from that consumer.

osac-metering/metering-service/internal/watch/consumer.go
```diff
@@ -346,6 +372,304 @@ func (c *Consumer) publishLifecycleEvents(ctx context.Context, baseCE *cloudeven
...
+			for i := range lifecycleEvents {
+				if err := c.publishWithRetry(ctx, &lifecycleEvents[i]); err != nil {
+					return err
+				}
+			}
+			return nil
+		}, projectionState, resourceID)
+	}
+
+	return c.publishAndUpsert(ctx, func() error {
```
masayag, line 545: The BMaaS handler publishes lifecycle events and then upserts the projection. A crash after Kafka publication but before the upsert causes the event to be published again on replay; deterministic IDs only suppress it while the adapter’s in-memory dedup cache is alive, and that cache expires/restarts. What guarantees provider-side idempotency across restarts? An outbox or durable event-id ledger is needed if duplicate delivery must not duplicate billing.

osac-metering/metering-service/internal/heartbeat/generator.go
```diff
@@ -134,30 +134,82 @@ func (g *Generator) publishResourceHeartbeats(ctx context.Context, hbEvents []cl
...
 	// e.g. a future in-tick retry — reproduces the same per-component
 	// CloudEvent IDs.
 	baseID := fmt.Sprintf("hb/%s/%d", state.ResourceID, now.Truncate(g.interval).Unix())
+	return BuildHeartbeatEvents(state, baseID, now, "osac-metering")
+}
+
+// BuildHeartbeatEvents builds the heartbeat event fan-out for one resource.
+// BMaaS has independent allocation and consumption meters; all other resource
+// types retain the existing resource decomposition behavior.
+func BuildHeartbeatEvents(state *projection.ResourceState, baseID string, now time.Time, source string) ([]cloudevents.Event, error) {
```
masayag, line 161: These new BMaaS heartbeats are routed to a different Kafka topic than lifecycle events. Kafka ordering is per topic, so a heartbeat generated from a pre-transition projection can arrive after a suspension or deletion. The adapter order tracker only records a metric and still submits the event, while heartbeat payloads do not carry transition_time for downstream ordering. Please define an ordering or interval-ledger contract, such as a shared topic/sequence or provider-side per-meter reconciliation.

omer-vishlitzky:
https://github.com/osac-project/osac/pull/998

we are getting rid of all the migrations

osac-metering/schema/resource.go
amito: Addressed via gating in a follow-up commit (will be pushed soon).

osac-metering/metering-service/internal/watch/consumer.go
amito: I believe this belongs in OSAC-983's outbox/replay architecture, or am I missing something?

osac-metering/metering-service/internal/heartbeat/generator.go
amito: Do you think this should be handled now, or by defining a contract after merging OSAC-983? (and only then enabling BMaaS)

osac-metering/metering-service/cmd/metering-service/main.go
```diff
@@ -284,7 +283,7 @@ func run(ctx context.Context, logger logr.Logger, cfg *config) error {
...
```
masayag, line 286: The new CLI default makes BMaaS opt-in only when no enablement environment variables are set, but the Helm serviceEnabled helper defaults every service tier to enabled when global.services is absent, and deployment.yaml therefore still emits ENABLE_BMAAS. Standard chart installs can still enter this branch with BMaaS enabled; with m360Adapter enabled, the existing M360 translator has no BMaaS endpoint. Please align the chart defaults with this opt-in behavior or enforce the compatibility gate explicitly.

osac-metering/metering-service/internal/projection/postgres.go
```diff
@@ -45,6 +46,9 @@ func (s *PostgresStore) Get(ctx context.Context, resourceID string) (*ResourceSt
 		}
 		return nil, fmt.Errorf("querying resource state %s: %w", resourceID, err)
 	}
+	if err := s.loadBMaaSMeterState(ctx, state); err != nil {
```
masayag: The normalized meter state is loaded in separate queries from the resource projection. A concurrent Watch update can therefore combine an older CurrentState/version with newer ActiveSince/FirstStartedAt values, or vice versa, causing incorrect lifecycle or heartbeat decisions. Load the resource and meter rows in one transaction/snapshot, or validate a shared version before returning the projection.

osac-metering/metering-service/internal/watch/consumer.go
```diff
@@ -180,11 +189,17 @@ func (c *Consumer) handleEvent(ctx context.Context, event *privatev1.Event) erro
...
 		}
+		if errors.Is(err, events.ErrDataQuality) && event.GetType() == privatev1.EventType_EVENT_TYPE_OBJECT_DELETED {
+			eventsSkipped.WithLabelValues("missing_event_timestamp").Inc()
+			c.logger.Info("skipping deleted event with missing timestamp",
+				"event_id", event.GetId(), "resource_id", resourceID)
+			return nil
+		}
 		return err
 	}
 	if projectionIsAhead(existing, version, currentState, dims) {
```
masayag, line 205: Lifecycle ordering currently uses fulfillment_version but does not reject an equal or earlier authoritative transition timestamp. A higher-version event can therefore regress TransitionTime or create a negative interval duration; same-state version updates can also overwrite the stored timestamp. Require strictly newer timestamps for state-changing events and deletion, while allowing equal timestamps only for metadata-only updates.

osac-metering/metering-service/internal/watch/consumer.go
```diff
@@ -400,6 +419,303 @@ func (c *Consumer) publishLifecycleEvents(ctx context.Context, baseCE *cloudeven
...
+			eventsSkipped.WithLabelValues("invalid_bmaas_transition").Inc()
+			c.logger.Info("skipping invalid BMaaS state transition",
+				"event_id", event.GetId(), "resource_id", resourceID,
+				"previous_state", previousState, "current_state", mapper.CurrentState())
+			return nil
+		}
+		return err
+	}
+
+	projectionState := c.buildBareMetalProjectionState(
```
masayag, line 504: A higher-version RUNNING-to-RUNNING event with changed BMaaS billing dimensions reaches this projection update without emitting a lifecycle boundary. Since spec.instance_type is immutable, overwriting the stored bm_instance_type silently relabels the already-active interval; subsequent heartbeats report the new type for time billed under the old type. Reject immutable-dimension drift as data quality or emit an explicit correction instead of updating the projection silently.

osac-metering/metering-service/internal/projection/postgres.go
```diff
@@ -122,6 +125,30 @@ func (s *PostgresStore) Upsert(ctx context.Context, state ResourceState) error {
 		return fmt.Errorf("upserting resource state %s: %w", state.ResourceID, err)
 	}
 
+	if state.ResourceType == schema.ResourceTypeBareMetalInstance {
```
masayag, line 138: BMaaS allocation state has two sources of truth: billable_since/generated is_billable drives ListBillable(), while metering_resource_meter_state.active_since drives BMaaS heartbeat generation. Upsert writes both, but no database constraint or read invariant keeps them synchronized. A divergent row can be selected as billable but emit no allocation heartbeat, or have an active meter that ListBillable never returns. Make the normalized allocation active_since authoritative for BMaaS, or enforce and validate the invariant.

osac-metering/metering-service/internal/watch/consumer.go
```diff
@@ -344,6 +381,29 @@ func projectionIsAhead(existing *projection.ResourceState, version int32, curren
...
+	currentState string,
+	dims map[string]any,
+	version int32,
+	transitionTime time.Time,
+) bool {
+	if existing == nil || version <= existing.FulfillmentVersion {
+		return false
+	}
+
+	stateChanging := existing.CurrentState != currentState || !events.DimensionsEqual(existing.BillingDimensions, dims)
```
masayag: stateChanging is true when billing dimensions change, even if the lifecycle state is unchanged. VMaaS/CaaS scaling updates commonly keep the same state_transition_time, so this guard can reject valid dimension changes. Use only existing.CurrentState != currentState for lifecycle timestamp ordering; keep BMaaS immutable-dimension validation separate.

osac-metering/charts/osac-metering/templates/deployment.yaml
```diff
@@ -140,7 +140,12 @@ spec:
             - name: ENABLE_VMAAS
               value: "true"
             {{- end }}
-            {{- if include "osac-metering.serviceEnabled" (list . "bmaas") }}
+            {{/* BMaaS Watch is opt-in until M360 accepts BMaaS (OSAC-5251). Default off. */}}
+            {{- $bmaas := dict }}
+            {{- if and .Values.global .Values.global.services }}
+            {{- $bmaas = index .Values.global.services "bmaas" | default dict }}
+            {{- end }}
+            {{- if and $bmaas (kindIs "map" $bmaas) (index $bmaas "enabled") }}
```
masayag: This undoes `7489f3c3f`. `ENABLE_BMAAS` follows `global.services.bmaas.enabled`, which is true in umbrella defaults, so default installs enable BMaaS Watch. The comment says opt-in / default off; that is not what renders. Leave `ENABLE_BMAAS` unset until OSAC-5251, or use a metering-specific flag that defaults false. Process-level opt-in in `enableAllIfNoneSet` is fine; Helm is what ships. With Watch on, M360 still has no `bare_metal_instance` translator (`NonRetryableError` on the shared lifecycle topic).

masayag (changes requested):
## Review (tip 2268a333a)

Prior Watch/projection items on this tip look fixed: snapshot JOIN, BMaaS dimension-drift skip, ListBillable via `allocation.active_since`, empty heartbeat checkpoint skip, and canonical proto states.

Two high-confidence issues remain:

- `transitionTimeIsStale` still treats dimension changes as lifecycle changes (`consumer.go:400`, already commented). Valid VMaaS/CaaS scales that reuse `state_transition_time` can be dropped.
- Helm re-derives `ENABLE_BMAAS` from `global.services.bmaas.enabled` (true in umbrella defaults), which undoes the opt-in from `7489f3c3f` while M360 still has no BMaaS translator.

osac-metering/metering-service/internal/watch/consumer.go
amito: Following up on this before the next iteration --
We are doing the same publish-then-upsert path as is done in VMaaS and CaaS.
The event IDs for BMaaS are deterministic; the in-memory cache does not give us a dedup guarantee. That being said, a transactional outbox is indeed in the OSAC-2506 design but not covered as part of the epic separation. I will add a separate task for it to make sure it is addressed.

masayag (approved):

PR #969, opened 2026-09-14
OSAC-5126: Add BareMetalWorkerReconciler for failure handling and retry

ori-amizur:
Implement bare-metal worker failure handling for ClusterOrder resources:

- Add condition constants: WorkerProvisioningFailed, WorkersFailed, FulfillmentServiceUnavailable with corresponding reason constants
- Add WorkerStatus struct (BMIName, BMINamespace, AttemptCount, NextRetryTime, LastFailureReason, LastFailureTime) and Workers field to ClusterOrderStatus
- Create BareMetalWorkerReconciler with:
  - 30-minute agent registration timeout that triggers BMI replacement
  - Escalating backoff between retries (reuses existing BackoffBaseDelay=2m, BackoffMaxDelay=30m doubling pattern from provisioning helpers)
  - IsTransientGRPCError() for Unavailable, DeadlineExceeded, ResourceExhausted, Aborted (transient errors NOT counted toward max retries)
  - FulfillmentServiceUnavailable condition on gRPC service errors
  - Terminal WorkersFailed condition after max retries exhausted
  - BMIProvider and FulfillmentClient interfaces for testability
- Add comprehensive Ginkgo unit tests covering: timeout detection, backoff calculation, gRPC error classification, terminal condition, BMI replacement flow, and fulfillment client reporting
- Run CRD/code generation and sync Helm CRDs

## Summary

- **API:** Added `WorkerStatus`, retry state, `AttemptStartTime`, and worker failure conditions.
- **Controllers:** Added BMI replacement, registration timeouts, transient gRPC handling, bounded backoff, fulfillment reporting, and terminal failure handling.
- **Deployment:** Updated generated and Helm CRDs with additive `status.workers` support.
- **Tests:** Added Ginkgo coverage for retries, timeouts, replacement, conditions, transient errors, timestamp handling, and client failures. Test execution results were not provided.
- **CI:** Jira validation passed for [OSAC-5126](https://redhat.atlassian.net/browse/OSAC-5126). The issue has no target version, while the target branch expects 5.1.0.
- **Compatibility:** Consumers of `ClusterOrder.status` must tolerate the new `workers` field. Orders with worker status now receive worker reconciliation and retry behavior. No database or authentication changes are included.

## Risk classification

**risk:show** — The PR changes production worker reconciliation and the public CRD status schema. It adds retries, BMI replacement, failure conditions, and fulfillment-service reporting. It does not qualify for **risk:ship** because it changes provisioning behavior. It does not qualify for **risk:ask** because the scope is defined and focused automated test coverage was added.


osac-operator/internal/controller/clusterorder_controller.go
```diff
@@ -316,6 +321,7 @@ func (r *ClusterOrderReconciler) patchStatusWithRetry(ctx context.Context, key c
 		latest.Status.DesiredConfigVersion = computed.DesiredConfigVersion
 		latest.Status.ApiEndpoint = computed.ApiEndpoint
 		latest.Status.IngressEndpoint = computed.IngressEndpoint
+		latest.Status.Workers = computed.Workers
```
vladikr, line 324: @ori-amizur I'm missing something here... who populates the Workers?! 
Is it in some follow up PR? I didn't find anything that populates this, it's always nil

osac-operator/internal/controller/baremetalworker_controller.go
```diff
@@ -0,0 +1,523 @@
...
+	backoff := ComputeWorkerBackoff(newAttemptCount)
+	now := r.now()
+	nextRetry := metav1.NewTime(now.Add(backoff))
+	failTime := metav1.NewTime(now)
+
+	worker.AttemptCount = newAttemptCount
+	worker.NextRetryTime = &nextRetry
+	worker.LastFailureReason = reason
+	worker.LastFailureTime = &failTime
+	worker.BMIName = newName
```
vladikr, line 561: okay... but that means that if the DeleteBMI succeeds but CreateBMI fails, we never get here! 
so the sattus still points at the old BMIName, right? 
the next reconciliation will get to replaceBMI but that old BMI is already gone...
I'm not sure what DeleteBMI does in this case (@carbonin fyi)

The fun part is that if DeleteBMI returns a a k8s error (NotFound or anything else) this `IsTransientGRPCError` returns false  so replaceBMI returns a hard error. 
the `ReconcileWorkers` aborts the whole worker list. CreateBMI never runs...
So ClusterOrder reconcile fails and retries forever on a name that no longer exists.

osac-operator/internal/controller/baremetalworker_controller.go
```diff
@@ -0,0 +1,523 @@
...
+		return ctrl.Result{}, false, nil
+	}
+
+	// Create replacement BMI before mutating retry state, so that a
+	// transient CreateBMI failure does not leave stale AttemptCount,
+	// NextRetryTime, or LastFailureReason in the worker status.
+	workerIndex := r.findWorkerIndex(instance, worker)
+	newName, newNamespace, err := r.BMIProvider.CreateBMI(ctx, instance, workerIndex)
+	if err != nil {
+		if IsTransientGRPCError(err) {
```
vladikr: but what if `IsTransientGRPCError` gets a NotFound or another client-go error?

osac-operator/internal/controller/baremetalworker_controller.go
```diff
@@ -0,0 +1,523 @@
...
+		}
+		return ctrl.Result{}, false, fmt.Errorf("deleting BMI %s/%s: %w", worker.BMINamespace, worker.BMIName, err)
+	}
+
+	// Compute new attempt count without mutating the worker yet, so that
+	// a transient CreateBMI failure does not persist stale retry state.
+	newAttemptCount := worker.AttemptCount + 1
+
+	// Check if max retries exhausted after increment
+	if newAttemptCount >= r.MaxRetries {
```
vladikr, line 484: @ori-amizur what was the intention here?
so if I have to do replaceBMI `MaxRetries` times (let's say MaxRetries is 5)
so on the last attempt, we delete the old BMI 
now `f newAttemptCount >= r.MaxRetries {` is true
and we return... then the next reconcile 
```
func (r *BareMetalWorkerReconciler) reconcileWorker(
	ctx context.Context, instance *v1alpha1.ClusterOrder, worker *v1alpha1.WorkerStatus, now time.Time,
) (workerReconcileResult, error) {

	// Check if max retries exhausted
	if worker.AttemptCount >= r.MaxRetries {
		return workerReconcileResult{exhausted: true}, nil   <----- we exit and there is no create again 
	}
```
but the status would still point at the old BMIName...

osac-operator/api/v1alpha1/clusterorder_types.go
```diff
@@ -212,6 +258,13 @@ type ClusterOrderStatus struct {
 	// +listType=map
 	// +listMapKey=name
 	NodeSets []NodeSetStatus `json:"nodeSets,omitempty"`
+
+	// Workers tracks per-worker provisioning state for failure handling and retry.
+	// Each entry corresponds to a bare-metal worker slot in the cluster.
+	// +kubebuilder:validation:Optional
+	// +listType=map
+	// +listMapKey=bmiName
```
vladikr: is it a good idea to key by a bmiName!? 
how stable is this name?

osac-operator/internal/controller/baremetalworker_controller.go
```diff
@@ -0,0 +1,523 @@
...
+			}
+		}
+	}
+
+	// Clear FulfillmentServiceUnavailable only after the entire reconciliation
+	// loop (including the fulfillment report) completed with zero transient
+	// gRPC errors. This prevents a single successful worker from prematurely
+	// clearing a condition that still applies to other workers.
+	if !hadTransientError {
+		instance.RemoveStatusCondition(v1alpha1.ConditionFulfillmentServiceUnavailable)
```
vladikr: how is this getting persisted? we removed the condition from the cached object and how do we persist it?

vladikr: looking at patchStatusWithRetry, there we get a fresh object via GET.
and then run 
```
for _, c := range computed.Conditions {
    apimeta.SetStatusCondition(&latest.Status.Conditions, c)
}
```
so it can only add/update the conditions that exist on the `latest` .. but it doesn't remove whatever we removed from the instance/computed

osac-operator/internal/controller/baremetalworker_controller.go
```diff
@@ -0,0 +1,589 @@
...
+	// when the entire reconciliation loop completes with zero transient
+	// errors. Setting to False ensures patchStatusWithRetry persists the
+	// cleared state via SetStatusCondition; removing would only modify
+	// the in-memory object without propagating to the API server.
+	if !hadTransientError {
+		instance.SetStatusCondition(
+			v1alpha1.ConditionFulfillmentServiceUnavailable,
+			metav1.ConditionFalse,
+			"No transient errors during reconciliation",
+			v1alpha1.ReasonAllWorkersHealthy,
```
vladikr: Maybe we should think of a different name for the reason, because we say on one hand that there are no transient errors, but it doesn't mean that all workers are healthy; we could have just replaced all of them.

osac-operator/internal/controller/baremetalworker_controller.go
```diff
@@ -0,0 +1,589 @@
...
+	return false
+}
+
+// IsTransientError returns true if the error is transient (should not count
+// toward the maximum retry budget) across both gRPC status codes and common
+// Kubernetes client errors. It covers:
+//   - gRPC: Unavailable, DeadlineExceeded, ResourceExhausted, Aborted
+//   - K8s: network timeouts, connection refused (net.Error with Timeout())
+//   - K8s: API server throttling (429 Too Many Requests)
+//   - K8s: server errors (5xx) indicating temporary unavailability
```
vladikr: I would say something like the following instead.
//   - k8s API: Timeout, ServerTimeout, ServiceUnavailable, TooManyRequests, InternalError
//   - network: any net.Error (timeout, connection refused, DNS), including wrapped errors
// NotFound, Conflict, and other client errors are not transient.

osac-operator/internal/controller/baremetalworker_controller.go
```diff
@@ -0,0 +1,589 @@
...
+			}, nil
+		}
+		return workerReconcileResult{}, fmt.Errorf("checking agent registration for %s/%s: %w",
+			worker.BMINamespace, worker.BMIName, err)
+	}
+
+	if regTime.IsZero() {
+		return r.handleUnregisteredAgent(ctx, instance, worker, now)
+	}
+
```
vladikr, line 399: I think there is an issue here. the agent can be registered but the BMI is not ready.
We could do something like that, but I don't know if that's the right timeout @rccrdpccl @carbonin wdyt?

```
if now.Sub(regTime) >= r.AgentRegistrationTimeout {
	result, transient, err := r.replaceBMI(ctx, instance, worker,
		v1alpha1.ReasonBMINotReadyTimeout,
		fmt.Sprintf("BMI did not become Ready within %s after agent registration", r.AgentRegistrationTimeout))
	if err != nil {
		return workerReconcileResult{}, err
	}
	return workerReconcileResult{requeueAfter: result.RequeueAfter, hadTransientError: transient}, nil
}

remaining := r.AgentRegistrationTimeout - now.Sub(regTime)
return workerReconcileResult{
	requeueAfter: minDuration(bootingWorkerRequeueInterval, remaining),
}, nil
```

or we can just add a TODO here to handle it in a follow-up when BMIProvider is wired 
I guess todo is better now.

osac-operator/internal/controller/baremetalworker_controller.go
```diff
@@ -0,0 +1,589 @@
...
+func (r *BareMetalWorkerReconciler) reconcileWorker(
+	ctx context.Context, instance *v1alpha1.ClusterOrder, worker *v1alpha1.WorkerStatus, now time.Time,
+) (workerReconcileResult, error) {
+	log := ctrllog.FromContext(ctx)
+
+	// Check if max retries exhausted
+	if worker.AttemptCount >= r.MaxRetries {
+		return workerReconcileResult{exhausted: true}, nil
+	}
+
```
vladikr, line 343: So each time we delete a failed bmi but didn't manage to create one, we remove the BMIName and when we come here, obviously the bmi is not ready... so we will never recreate it
I think we need to account for this case here. add something like that before the IsBMIReady chcek.
```
if worker.BMIName == "" {
		result, transient, err := r.createReplacementBMI(ctx, instance, worker,
			worker.LastFailureReason,
			"Retrying BMI creation after previous deletion")
		if err != nil {
			return workerReconcileResult{}, err
		}
		return workerReconcileResult{requeueAfter: result.RequeueAfter, hadTransientError: transient}, nil
	}
```

and we need to split the `createReplacementBMI` from the `replaceBMI`, something like 
```
func (r *BareMetalWorkerReconciler) createReplacementBMI(
	ctx context.Context,
	instance *v1alpha1.ClusterOrder,
	worker *v1alpha1.WorkerStatus,
	reason, message string,
) (ctrl.Result, bool, error) {
	oldBMIName := worker.BMIName

	newAttemptCount := worker.AttemptCount + 1
	workerIndex := r.findWorkerIndex(instance, worker)
	newName, newNamespace, err := r.BMIProvider.CreateBMI(ctx, instance, workerIndex)
	if err != nil {
		if IsTransientError(err) {
			r.setTransientGRPCCondition(instance, err)
			return ctrl.Result{RequeueAfter: provisioning.BackoffBaseDelay}, true, nil
		}
		return ctrl.Result{}, false, fmt.Errorf("creating replacement BMI: %w", err)
	}

	backoff := ComputeWorkerBackoff(newAttemptCount)
	now := r.now()
	nextRetry := metav1.NewTime(now.Add(backoff))
	failTime := metav1.NewTime(now)

	worker.AttemptCount = newAttemptCount
	worker.NextRetryTime = &nextRetry
	worker.LastFailureReason = reason
	worker.LastFailureTime = &failTime
	worker.BMIName = newName
	worker.BMINamespace = newNamespace
	worker.AttemptStartTime = nil

	instance.SetStatusCondition(
		v1alpha1.ConditionWorkerProvisioningFailed,
		metav1.ConditionTrue,
		fmt.Sprintf("Replaced %s with %s (attempt %d/%d, next retry after %s): %s",
			oldBMIName, newName, newAttemptCount, r.MaxRetries, backoff, message),
		v1alpha1.ReasonBMIReplacementTriggered,
	)
	log.Info("BMI replacement created", "oldBMI", oldBMIName, "newBMI", newName, "attempt", newAttemptCount)
	return ctrl.Result{RequeueAfter: backoff}, false, nil
}
``` 

so at the end of replaceBMI it would call `createReplacementBMI` after clearing the name.

vladikr (commented):
@ori-amizur thank you for making the changes! It's a huge PR...
I think it's mostly good, but there are 2 corner cases

osac-operator/internal/controller/baremetalworker_controller.go
ori-amizur: What is bad with the current definition ?

osac-operator/internal/controller/baremetalworker_controller.go
ori-amizur: The status is false.  How can we think that it is healthy ?

osac-operator/internal/controller/baremetalworker_controller.go
vladikr: @ori-amizur @redhat-chai-bot 
We should drop the `worker.AttemptCount > 0` - it's a bug... 
AttemptCount is still only incremented after a successful create inside replaceBMI
so the point of this code is to recreate the BMI if create didn't succeed during the replaceBMI
Empty name already means “delete happened, create did not.”

vladikr:
@ori-amizur everything looks great! Except `if worker.BMIName == "" && worker.AttemptCount > 0 {` we should just drop the `&& worker.AttemptCount > 0`  and all done :)

danmanor (approved):

rccrdpccl:
Sorry, I did not notice the pings earlier. This work is largely duplicate of aspects that would be handled by OSAC-2135.
I will try to merge the changes when we'll get to it :(

PR #980, opened 2026-09-15
NO-ISSUE: [DO NOT MERGE] Dummy intentional failures to exercise the AI diagnostic pipeline

eliorerz:
**DO NOT MERGE.** Throwaway PR to watch the new "⏳ Running" reservation
comment ([osac-test-infra#551](https://github.com/osac-project/osac-test-infra/pull/551) /
[osac#977](https://github.com/osac-project/osac/pull/977), both merged)
resolve into a real AI diagnosis end to end, covering both failure
classes the pipeline is meant to handle:

- **Build-stage failure (before provisioning)**: an unused `"unicode"`
  import in `osac-metering/adapters/cmd/echo-adapter/main.go` breaks
  `go build`. `echo-adapter` is built by VMaaS and CaaS but *not* by
  BMaaS (confirmed by diffing each suite's own component list in
  `.github/workflows/e2e-*-full-install.yml`), so this fails VMaaS's and
  CaaS's "Build and load component images" step before either ever boots
  a cluster.
- **Test-stage failure (after provisioning)**: an always-false assertion
  at the top of `test_baremetal_instance_lifecycle`, before any real
  resource creation, fails BMaaS during actual test execution -- BMaaS
  doesn't build `echo-adapter`, so its build stays intact and it reaches
  this point normally.

Both are obviously-deliberate, easily-revertible breaks (a clearly
commented unused import; an `assert False` with an explanatory message),
not real product regressions.

Closing this PR without merging once observed.

## Summary

- **CI/build:** Added an unused `unicode` import to `echo-adapter/main.go`, which causes the Go build to fail.
- **Tests:** Added an always-false assertion to `test_baremetal_instance_lifecycle`, which fails before bare-metal resource creation.
- **Affected areas:** CI and tests only. No API, controller, database, authentication, deployment, or documentation changes.
- **Compatibility:** No runtime compatibility impact is expected. The intentional failures make the PR non-mergeable.
- **Validation:** Test results and review finding counts are unavailable.

## Risk classification

**risk:ask** — based on the intentional build and test failures and the stated need to observe diagnostics before closing the PR. The supplied labeling criteria do not define an exact mapping. **risk:show** was not selected because the changes block validation rather than only showing a runtime behavior.


eliorerz:
Observation complete -- all three E2E suites correctly diagnosed the intentional dummy failures end-to-end (Reserve -> Running -> diagnosis), confirming the pipeline works as designed. Closing without merging (per the PR description).

PR #983, opened 2026-09-15
OSAC-5241: reconcile BMaaS usage

amito:
## Summary

Reconcile BMaaS usage safely across missed resources, version gaps, lifecycle transitions, stale heartbeats, and independent meter boundaries.

This PR is stacked on [PR #960](https://github.com/osac-project/osac/pull/960) (`OSAC-5240: implement BMaaS metering lifecycle and heartbeats`). GitHub currently shows `main` as the base because the stack branches live in the contributor fork; review this PR as the layer above #960.

Review changes are folded into the mapped commits:
- [OSAC-5247](https://redhat.atlassian.net/browse/OSAC-5247) load BMaaS state: replay source, unavailable-history handling, malformed dimensions.
- [OSAC-5247](https://redhat.atlassian.net/browse/OSAC-5247) reconcile meter boundaries: version gaps, transition boundaries, durations, idempotency.
- [OSAC-5248](https://redhat.atlassian.net/browse/OSAC-5248) stale heartbeats: lifecycle-state coverage.
- [OSAC-5241](https://redhat.atlassian.net/browse/OSAC-5241) harden reconciliation holds: hold observability, immutable dimensions, nil-client behavior.

## Validation

- `make -C osac-metering test`
- `make -C osac-metering lint`

## Summary

- Adds BMaaS reconciliation for missed resources, lifecycle transitions, version gaps, stale heartbeats, and independent allocation and consumption meter boundaries.
- Adds BMaaS presence tracking, replay-source handling, dimension validation, meter holds, and reconciliation observability.
- Adds PostgreSQL tombstones with `deleted_at`. Tombstones preserve first-use timestamps, clear active meter intervals, and block later updates.
- Makes deletion version-aware through `DeleteIfVersion`. Deleted projections become terminal.
- Updates watch and heartbeat flows to ignore stale or absent BMaaS resources.
- Expands reconciliation, projection, heartbeat, watch, and schema tests.

## API and compatibility

- `NewReconciler` now requires BMaaS and replay-source arguments.
- The projection `Store` interface replaces `Delete` with `DeleteIfVersion`.
- New exported BMaaS interfaces, types, and helpers are available.
- Implementations and test doubles that use the changed interfaces must be updated.
- Existing database tables receive the `deleted_at` column during initialization.

## Tests and validation

The stated validation commands are `make -C osac-metering test` and `make -C osac-metering lint`. Their execution results are not supplied.

## Risk classification

**risk:ask** — applied conservatively because the change modifies reconciliation, projection deletion, database state, and public interfaces. The labeling criteria were not supplied, so a more specific classification cannot be verified. The PR may qualify for **risk:show** if the labeling rules permit broad, well-tested internal behavior changes without a confirmed production failure.


osac-metering/metering-service/cmd/metering-service/main.go
```diff
@@ -236,13 +236,20 @@ func run(ctx context.Context, logger logr.Logger, cfg *config) error {
...
 		computeClient = privatev1.NewComputeInstancesClient(grpcConn)
 	}
 	if cfg.enableCaaS {
 		clusterClient = privatev1.NewClustersClient(grpcConn)
 	}
-	reconciler := reconciliation.NewReconciler(computeClient, clusterClient, store, publisher, logger, cfg.heartbeatInterval)
+	if cfg.enableBMaaS {
+		bareMetalClient = privatev1.NewBareMetalInstancesClient(grpcConn)
+	}
+	replaySource := reconciliation.NewUnavailableBMaaSReplaySource()
```
masayag, line 255: Production wiring always uses NewUnavailableBMaaSReplaySource. Any missed creation, version gap, or deletion whose boundary requires history will therefore remain held indefinitely; the tests use fake replay sources and do not exercise the deployed wiring. If the real replay dependency is intentionally deferred, please make it an explicit rollout prerequisite; otherwise wire the production replay implementation here.

osac-metering/metering-service/internal/reconciliation/reconciler.go
```diff
@@ -308,6 +873,47 @@ func (r *Reconciler) reconcileMissedDeletions(ctx context.Context, fulfillmentSt
...
+				intervals := bmaasIntervals(ps)
+				published, err := r.publishBMaaSCorrections(ctx, id, ps.TenantID, ps.ProjectID, MissedDeletion,
+					ps.CurrentState, "", ps.BillingDimensions, intervals,
+					allocationEffect, consumptionEffect,
+					ps.ComponentEverStarted[events.BMaaSMeterAllocation], ps.ComponentEverStarted[events.BMaaSMeterConsumption], boundaryTime)
+				if err != nil {
+					return corrections, err
+				}
+				corrections += boolToInt(published)
+				if err := r.store.Delete(ctx, id); err != nil {
```
masayag: With this reconciliation path running concurrently with the Watch consumer, the unconditional Delete can remove projection state that was newer than the snapshot used to decide missed deletion. The Watch deletion path has the same unversioned delete behavior. Removing the row also erases the version watermark, allowing a delayed update to recreate the resource after deletion. Use compare-and-delete plus a durable deletion tombstone/watermark.

osac-metering/metering-service/internal/reconciliation/reconciler.go
```diff
@@ -901,7 +1248,86 @@ func (r *Reconciler) loadVolumes(ctx context.Context, result map[string]fulfillm
...
+			}
+			dimensions, err := events.BareMetalInstanceBillingDimensions(bmi)
+			if err != nil {
+				if r.bmaasSkipped == nil {
+					r.bmaasSkipped = make(map[string]struct{})
+				}
+				r.bmaasSkipped[bmi.GetId()] = struct{}{}
+				r.holdBMaaS(bmi.GetId(), "missing_instance_type")
+				r.logger.Error(err, "skipping bare metal instance with invalid billing dimensions", "resource_id", bmi.GetId())
+				continue
```
masayag: [P1] Preserve metering for resources whose source dimensions are invalid

This `continue` excludes the BMI from both `result` and `listedIDs`, but the loader still succeeds and publishes a new presence snapshot. An existing billable projection is then filtered from normal heartbeats, and stale-heartbeat reconciliation skips it because it is absent from `fulfillmentState`; if the bad source row persists, it receives no heartbeats indefinitely. Please fail the snapshot load before replacing presence, or track source presence independently from dimension conversion.

osac-metering/metering-service/internal/events/bmaas_transitions.go
```diff
@@ -140,6 +140,23 @@ func ResolveConsumptionTransition(from, to string) (string, error) {
...
+// resume.
+func BMaaSEffectEventType(effect string, everStarted bool) string {
+	switch effect {
+	case BMaaSEffectStart:
+		return ResolveLifecycleStartEvent(everStarted)
+	case BMaaSEffectResume:
+		return ResolveLifecycleStartEvent(everStarted)
+	case BMaaSEffectSuspend:
+		return EventSuspended
+	default:
```
masayag: [P2] Reject unknown BMaaS effects

The default also matches unknown strings, making a typo or new effect indistinguishable from the deliberate skip case. `buildBMaaSCorrectionEvents` can then emit no event for this meter, after which reconciliation still upserts the new state. Handle `BMaaSEffectSkip` explicitly and return an error for any other value.

osac-metering/metering-service/cmd/metering-service/main.go
```diff
@@ -248,6 +252,7 @@ func run(ctx context.Context, logger logr.Logger, cfg *config) error {
 	logger.Info("service ready")
 
 	hbGen := heartbeat.NewGenerator(store, publisher, logger, cfg.heartbeatInterval)
+	hbGen.SetBMaaSPresence(bmaasPresence)
```
omer-vishlitzky: set is redundant here, can simply pass it to the constructor

osac-metering/metering-service/cmd/metering-service/main.go
```diff
@@ -232,6 +235,7 @@ func run(ctx context.Context, logger logr.Logger, cfg *config) error {
 	if err != nil {
 		return fmt.Errorf("loading external IP pool families: %w", err)
 	}
+	reconciler.SetBMaaSPresence(bmaasPresence)
```
omer-vishlitzky: set is redundant here, can simply pass it to the constructor

PR #994, opened 2026-09-15
OSAC-5324: defer ClusterOrder readiness until HostedCluster is ready

tzvatot:
## [OSAC-5324](https://redhat.atlassian.net/browse/OSAC-5324): ClusterOrder readiness follows HostedCluster lifecycle

**Jira:** https://redhat.atlassian.net/browse/OSAC-5324

### Summary

Decouple AAP provisioning-job completion from the terminal ClusterOrder Ready phase. The order now remains progressing while HostedCluster and worker NodePool readiness are observed, preserving lifecycle transition events before the final Ready event.

### Changes

- Keep ClusterOrder progressing after a successful provisioning job.
- Require HostedCluster control-plane/API readiness and ClusterVersionSucceeding.
- Require `AllMachinesReady=True` for every requested worker NodePool.
- Add regression coverage for readiness ordering, recovery, and event sequencing.

### Testing

- Operator Go tests: passed.
- Controller transition-event tests: passed.
- `go vet`: passed.
- `golangci-lint`: passed.
- Helm lint: passed.
- CRD synchronization check: passed.
- Kind integration tests and CaaS E2E remain environment-blocked; details are in `05-validation-report.md`.

### Acceptance Criteria

- [x] Provisioning success does not prematurely mark ClusterOrder Ready.
- [x] HostedCluster stage transitions remain observable before terminal readiness.
- [x] Ready is emitted once for the terminal transition.
- [x] Unit coverage covers success, progression, and recovery paths.
- [ ] CaaS lifecycle E2E requires the deployed OSAC/AAP test environment.

## Summary

- **Controller:** Defers `ClusterOrder` readiness until provisioning succeeds and live HostedCluster and requested worker NodePools are ready.
- **Transition events:** Records skipped provisioning stages in order before `Ready`.
- **Tests:** Adds coverage for readiness ordering, worker readiness, recovery, status persistence, and transition events.
- **API surface:** No public API changes.
- **Database, auth, deployment, documentation:** No changes reported.
- **Compatibility:** Orders can remain `Progressing` longer because readiness now uses current cluster conditions.

## Validation

Operator tests, transition-event tests, `go vet`, `golangci-lint`, Helm lint, and CRD synchronization passed. Kind integration and CaaS E2E tests remain blocked by the required environment.

## Risk classification

Risk label: **unavailable**. The supplied context does not provide the labeling criteria or an applied label. Therefore, classification as `risk:ship`, `risk:show`, or `risk:ask` cannot be confirmed.


osac-operator/internal/controller/clusterorder_controller.go
```diff
@@ -739,6 +747,56 @@ func hostedClusterIsReady(hc *hypershiftv1beta1.HostedCluster) bool {
...
+	}
+	if len(instance.Spec.NodeRequests) > 0 && len(nodePools) == 0 {
+		return false
+	}
+
+	for i := range nodePools {
+		if !nodePoolIsReady(&nodePools[i]) {
+			return false
+		}
+	}
```
vladikr, line 803: @tzvatot, I didn't understand here what exactly the satisfying condition here... 
is it really enough to have one ready pool regardless of the NodeRequest? 

It also looks different from the handleNodePool logic here https://github.com/osac-project/osac/blob/main/osac-operator/internal/controller/clusterorder_controller.go#L679

I mean, none of this actually checks that the requested capacity exists.

osac-operator/internal/controller/clusterorder_controller.go
```diff
@@ -739,6 +747,56 @@ func hostedClusterIsReady(hc *hypershiftv1beta1.HostedCluster) bool {
...
+	return allMachinesReady && poolReady
+}
+
+func provisioningJobSucceeded(instance *v1alpha1.ClusterOrder) bool {
+	job := provisioning.FindLatestJobByType(instance.Status.ProvisioningJobs, v1alpha1.JobTypeProvision)
+	return job != nil && job.State == v1alpha1.JobStateSucceeded
+}
+
+func finalizeReadyIfProvisioned(instance *v1alpha1.ClusterOrder, hc *hypershiftv1beta1.HostedCluster,
```
vladikr, line 826: It's really difficult to follow :(

Looking at this it doesn't look to me like a readiness check. 
It withholds `Ready` so the next reconcile can emit `WorkersJoining` before `recordTransitionEvents` sees Reason=AsExpected. 

the return value is unused, and there is nothing that explicitly requeues that follow-up pass.
There is no `RequeueAfter` when Ready is withheld only because this was the first observation.
What is the way to have a follow up path? There won't be one, only if there was a ClusterOrder status patch to enqueue this again.
The new test calls `Reconcile` twice, so it doesn't really prove the follow-up
happens on its own.

I dont think this two pass protocol is worth the complexity.

I would suggest to derive the `Ready` from live HostedCluster/NodePool/job state. 
The event recorder should emit every stage we crossed when we go Ready in one observation

vladikr:
I would like to suggest splitting this into three parts.

1. Derive Ready from the current objects that are in front of us, not from the previous reconcile.

We should look at the live HostedCluster, NodePools, and provision job.

If the provision job already succeeded, the HostedCluster control plane / API /
ClusterVersion are ready, and the requested worker NodePools are ready, so we cab set
Phase=Ready and Progressing=False.

Otherwise we need to keep Phase=Progressing, and set the Progressing reason from the
current HostedCluster conditions (PreparingInfrastructure,
ControlPlaneStarting, or WorkersJoining).

I wouldn't use a saved flag like `controlPlaneWasAvailable` to decide Ready.
That flag is “what we thought last time,” not “what the cluster is now.”

2. When we skip a stage, still emit the event for it.

A cluster does not become Ready in one jump.

The progression should be a squence:

PreparingInfrastructure -> ControlPlaneStarting -> WorkersJoining -> Ready

In the general case that is spread across many reconciles, each time the HostedCluster
changes we reconcile, save the new stage and emit that stage’s
event, then wait again. After a while workers finish joining, we
reconcile again, and only then we set Ready.

The weird case is when one reconcile already sees the finished cluster.

It when the control plane is up and workers are already ready. 
If we set Phase=Ready immediately, `recordTransitionEvents` only compares the start and the end
of that reconcile. 
Start may still be ControlPlaneStarting, end is Ready.
So WorkersJoining never gets to the status, and that event is never emitted.

This PR avoids that by refusing Ready on the first look, and hoping a
second reconcile runs. 

I would keep Ready on the first look, and fix the event recorder instead.
Give it the ordered list above. 
When the old stage is `ControlPlaneStarting` and the new state is Ready, 
we should emit WorkersJoining and then Ready. 
When the stages changes one at a time, we emit each one as it happens. 
Same function either way and we don't need thta extra reconcile.

3. `OnSuccess` should not set phase nor the Progressing reason.

`OnSuccess` means the AAP provision job finished. 
That is not the same thing as “the cluster is Ready” and it is also not a useful stage name.

Right now it sets Phase=Progressing and Reason=Progressing. 
That is a generic label. We already have better ones from the HostedCluster:
PreparingInfrastructure, ControlPlaneStarting, WorkersJoining.

If `OnSuccess` writes Reason=Progressing on the same pass,`handleHostedCluster` can lose that detailed stage. 
I think we should let `OnSuccess` only record that the job succeeded.
And the HostedCluster/NodePool observer decide on the Phase and the Progressing reason.

osac-operator/internal/controller/clusterorder_controller.go
tzvatot: Good catch. The previous check only verified that at least one NodePool existed and that all pools were ready. I updated the readiness check to compare the requested node count with the aggregate observed NodePool.Status.Replicas, while still requiring every pool to report both AllMachinesReady=True and Ready=True. I also added coverage for insufficient capacity.

osac-operator/internal/controller/clusterorder_controller.go
tzvatot: Agreed. I removed the controlPlaneWasAvailable two-pass protocol. Ready is now derived directly from the current provisioning job, HostedCluster, and NodePool observations, so a single reconcile can finalize the order. The lifecycle test now verifies this one-pass behavior.

tzvatot:
> I would suggest to derive the Ready from live HostedCluster/NodePool/job state.\n>\n> The event recorder should emit every stage we crossed when we go Ready in one observation.\n>\n> provisioningCallbacks.OnSuccess() should not set Phase nor Progressing reason.\n\nAgreed. I removed the two-pass readiness protocol and now derive Ready from the current provisioning job, HostedCluster, and NodePool observations. A single reconcile emits skipped lifecycle stages in order before Ready, and OnSuccess() no longer changes the phase or progressing reason. The updated tests cover the one-pass lifecycle, ordered events, and insufficient worker capacity.

tzvatot:
Closing in favor of [https://github.com/osac-project/osac/pull/1028](https://github.com/osac-project/osac/pull/1028) and followup PRs.

PR #1003, opened 2026-09-15
OSAC-5337: Split release into a reusable build/publish workflow

eliorerz:
## Summary

Implements [OSAC-5337](https://redhat.atlassian.net/browse/OSAC-5337): splits the actual
build/test/publish/tag pipeline out of `nightly-build.yaml` into a new `workflow_call`-only
reusable workflow, `osac-build-and-publish.yaml`, parameterized by `mode: nightly | release`.
`nightly-build.yaml` shrinks to its schedule trigger; a new `osac-release.yaml` is the
`workflow_dispatch`-only entry point for cutting a real release. One implementation, two thin
callers -- no drift between "how nightly builds a component" and "how a release builds a
component" (the same class of bug [OSAC-5178](https://redhat.atlassian.net/browse/OSAC-5178) fixed for a hand-copied regex).

This corrects the real defect the first `osac v0.0.10` release dispatch exposed: every component
was rebuilt from `HEAD` and tagged `-nightly.<suffix>` on *every* release, regardless of whether
it had changed since its last real release -- so a "release" never gave any component a real,
traceable `<component>/vX.Y.Z` tag identifying what actually shipped.

## What changed

- **`osac-build-and-publish.yaml`** (new): the actual pipeline (`prepare` → `build` → tests →
`publish` → `tag-and-notify` → `cleanup`), unchanged in structure from `nightly-build.yaml`
except for the version-resolution algorithm below.
- **`nightly-build.yaml`**: shrinks to its schedule + manual `skip_e2e` trigger, calls the
reusable workflow with `mode: nightly`. Behavior is unchanged.
- **`osac-release.yaml`** (new): `workflow_dispatch`-only, with `release_version` (required) and
an optional `component_versions` input, calls the reusable workflow with `mode: release`.
- **`lib.sh`**: adds `real_tag_exists`, a small helper checking whether a `<prefix>/v<version>`
tag already exists.

**Unified per-component version-resolution algorithm** (one algorithm, both modes):

1. No explicit version requested (release mode default): pin the component's current latest real
`<component>/vX.Y.Z` tag verbatim -- no rebuild, no retag.
2. Explicit version requested (`component_versions`, e.g. `osac-operator=0.0.13`) and it already
exists: pin it too -- avoids redundant rebuilds.
3. Explicit version requested and it doesn't exist yet: build fresh from `HEAD`, run the full
existing test gate, and *only on success*, tag it as a real, permanent git tag and publish at
that version.

Nightly mode is the trivial case of the same algorithm: its computed target
(`<base>-nightly.<date>.<sha>.<run>.<attempt>`) is guaranteed to never already exist, so it always
builds -- the only other mode-specific difference is that a successful nightly build never gets
promoted to a permanent tag.

This deliberately avoids a separate "bump this component first, then release" step: cutting a
release and bumping a couple of components that need last-minute fixes is one dispatch, not three
manual steps across two mechanisms.

**Pinned components need zero build/test/package work.** `build`'s "Resolve component version"
step now skips a component when it's pinned (not just when it has no resolved version at all,
the pre-existing case). `publish`'s image-promotion and sub-chart packaging steps skip pinned
components too -- their chart/image already exist in GHCR at the resolved version; the umbrella's
existing OCI-dependency-pull mechanism ([OSAC-5178](https://redhat.atlassian.net/browse/OSAC-5178)) fetches them directly rather than anything
being repackaged from this `HEAD` checkout, which may not even match what was actually published
at that historical tag.

**Real per-component tags are atomic with the rest of the pipeline.** `tag-and-notify` creates a
permanent `<component>/vX.Y.Z` tag for every component that needed building (release mode only)
*after* `publish` has already succeeded -- which itself required the whole test gate to pass. A
component only earns a permanent tag once everything requested in that run passed together.

## What I validated

- YAML parses cleanly for all three workflow files; `bash -n` syntax-checked every embedded `run:`
block across all three (with `${{ }}` expressions stripped first).
- Ran the actual "Resolve component versions" script standalone against this repo's real git tags
for: nightly mode: release mode with no overrides (all pinned): an override to an
already-existing tag (pins, no rebuild): an override to a not-yet-existing version (builds and
tags): multiple simultaneous overrides, mixed existing/new: and every error path (unknown
component name, malformed `component=version` entry, invalid semver, `component_versions` set in
nightly mode, `release_version` missing in release mode). All behaved as designed -- see commit
for the fuller reasoning.
- Cross-checked every `needs.prepare.outputs.*`/`needs.publish.outputs.*` reference in the
reusable workflow against what those jobs actually declare in `outputs:` -- no mismatches.
- Confirmed no stale references remain anywhere in the repo to the removed
`force_shared_component_version` input or the renamed "Resolve nightly versions" step.

## What I could not validate

I cannot trigger a real `workflow_dispatch` run of either `nightly-build.yaml` or
`osac-release.yaml` from here. The actual GitHub Actions execution -- job-to-job data flow via
`needs.*.outputs`, the nested `workflow_call`/reusable-workflow permission propagation, and the
cross-repo `e2e-*-full-install.yml@main` calls two levels deep -- needs a real dispatch to
confirm end-to-end.

## Deliberately out of scope (see ticket's Open Questions)

- Retiring `publish-osac-installer-chart.yaml` -- still needs one proven real `osac-release.yaml`
dispatch first, per OSAC-5183's original sequencing.
- Atomicity across *osac-ui's* bump path -- osac-ui has its own separate repo/release process and
isn't exposed through `component_versions`; release mode always pins its latest tag.

## Summary

- **CI/deployment:** Added reusable `osac-build-and-publish.yaml` for nightly and release builds, tests, publishing, tagging, notifications, and cleanup.
- **Release workflow:** Added manual `osac-release.yaml` with `release_version`, `component_versions`, and `skip_e2e` inputs.
- **Nightly workflow:** Reduced `nightly-build.yaml` to a scheduled or manual caller.
- **Release safety:** Added semver validation, duplicate release checks, component tag resolution, and post-publish permanent tagging.
- **Shell API:** Added `real_tag_exists(prefix, version)` to `osac-installer/scripts/lib.sh`.
- **Validation:** Checked YAML and shell syntax, version resolution, error paths, workflow outputs, and stale references. End-to-end GitHub Actions dispatches were not run.
- **Compatibility:** Nightly behavior remains unchanged. Release automation is new. Existing component versions skip rebuild, testing, image promotion, and chart packaging.
- **Other areas:** No controller, database, authentication, or documentation changes were identified.

## Review status

Current review severity counts are unavailable because no review findings were supplied.

## Risk classification

**risk:show** — applied because the PR changes CI release and publication workflows, including artifact promotion, chart publication, tagging, validation, and failure handling. Static and scenario validation was performed, but end-to-end workflow dispatches were not run.

The supplied labeling instructions do not define criteria for `risk:ship`, `risk:show`, or `risk:ask`. Therefore, a comparison with a nearby classification cannot be verified.


eliorerz:
Verified all 4 findings against current code before touching anything:

**Finding 1 (missing duplicate-release guard) — confirmed, fixed.** There was no check preventing a release dispatch from re-publishing an already-existing `osac/vX.Y.Z` tag, and my own `osac-release.yaml` comment falsely claimed such a check existed. Added `real_tag_exists osac "${RELEASE_VERSION}"` right after semver validation in "Resolve component versions," and rewrote the concurrency comment to describe the real mechanism instead of an aspirational one. Tested: a real existing tag (`osac/v0.0.9`) is now rejected with a clear error; a genuinely new version still resolves normally.

**Finding 2 (`UI_SUB_VERSION` guard swallowing exit status) — confirmed, fixed.** `[[ -n "${UI_SUB_VERSION}" ]] && stamp_ci_overlay_if_present ...` as the last line of a `set -euo pipefail` script meant an empty (valid) `UI_SUB_VERSION` would fail the entire step. Pre-existing in the original file, worth fixing since it's in this diff. Replaced with an explicit `if` block.

**Finding 3 (`resolve_release_tag` sourcing tags from `HEAD`-reachable `git describe`) — skipping, stale citation.** The cited lines (1338-1351) are my "Create per-component release tags" step, unrelated to `resolve_release_tag`, which actually lives at `lib.sh:151` — unmodified by this PR and unchanged from what every nightly build has relied on in production already. Not touching pre-existing, unmodified code based on a misdirected line reference.

**Finding 4 (`A && B || C` overlay-stamping anti-pattern) — confirmed, fixed.** If the first `stamp_ci_overlay_if_present` call failed, `||` would run the second one too, potentially masking a real failure or double-stamping. Fixed all three real instances (operator, BMF, CSI) with explicit `if`/`else`. Left metering's `&&`/`||` alone — it only sets a variable (`METERING_TAG=`), no duplicate side effect risk there.

Re-validated everything: YAML parse, bash syntax on every embedded script block, and re-ran the full functional test suite from the previous round (nightly, release default-pin, override-to-existing, override-to-new, multi-override, all error paths) plus a new one for the duplicate-release guard — all pass. Amended and force-pushed.

eliorerz:
Verified both findings against current code:

**Finding 1 (override ignored when a component has never been tagged) — confirmed real, fixed.** The loop required a successful `resolve_release_tag` *before* ever checking for an override, so `component_versions` could never bootstrap a component's first-ever release -- it would just silently skip with a warning, override or not. Reordered so an override is checked first (release mode only; nightly always requires a real prior tag, unchanged); a non-overridden component with no tag still skips exactly as before.

Tested against a scratch clone with a component's only tag deleted (simulating "never released"):
- With an override requesting its first release: correctly builds and tags it, everything else stays pinned as usual.
- With no override: still skips cleanly with the same warning as before -- no regression.

Also re-ran the full existing scenario suite (nightly, release default-pin, override-to-existing, override-to-new, duplicate-release rejection) -- all unchanged and passing.

**Finding 2 (`@main` instead of a pinned SHA for the vmaas/caas/bmaas e2e reusable-workflow calls) — skipping, out of scope here.** Checked: `osac-project/osac-test-infra@main` is used identically in 20+ other places across this repo's own `.github/workflows/` (e2e-caas-full-install.yml, e2e-bmaas-full-install.yml, e2e-vmaas-full-install.yml, slash-command.yml, and more) -- this is a deliberate, pre-existing, repo-wide convention, not something introduced by this PR or unique to these 3 lines. Pinning just these three would create inconsistency with the other 20+ references rather than fix anything this PR is responsible for. A repo-wide SHA-pinning policy change belongs in its own dedicated PR/ticket, not bundled into this one.

Re-validated: YAML parse, bash syntax on every embedded script, and the full functional test suite (including the new bootstrap scenario). Amended and force-pushed.

PR #1012, opened 2026-09-16
NO-ISSUE: retire publish-osac-installer-chart.yaml, fix stale signing docs

eliorerz:
## Summary

Retires `publish-osac-installer-chart.yaml` -- superseded by `osac-release.yaml` ([OSAC-5337](https://redhat.atlassian.net/browse/OSAC-5337)), now
proven via a real successful `osac v0.0.10` release (all 6 mono-repo components bumped and
tagged for real: `osac-operator/v0.0.13`, `fulfillment-service/v0.0.99`, `osac-aap/v0.0.14`,
`bare-metal-fulfillment-operator/v0.0.13`, `osac-metering/v0.0.2`, `osac-csi-driver/v0.0.2`).
This is the sequencing both [OSAC-5183](https://redhat.atlassian.net/browse/OSAC-5183) and [OSAC-5337](https://redhat.atlassian.net/browse/OSAC-5337) always intended ("retire once
`osac-release.yaml` has one proven real dispatch") -- that dispatch has now happened.

Also fixes README documentation that was already factually wrong as a result of [OSAC-5337](https://redhat.atlassian.net/browse/OSAC-5337)'s
refactor, confirmed empirically against the real `v0.0.10` signatures.

## What changed

- **Deleted** `.github/workflows/publish-osac-installer-chart.yaml`.
- **Deleted** `check_chart_published` / `check_umbrella_mono_repo_charts_published` from
`nightly-charts.sh` -- both were only ever called from the now-deleted workflow (the old
force-shared-version pre-flight check). Confirmed dead code: no other callers anywhere in the
repo.
- **Fixed README.md's cosign verification commands.** Both the image and chart sections
documented `publish-osac-installer-chart.yaml` / `nightly-build.yaml` as the alternate signer
identities for nightly/release-produced artifacts. Since [OSAC-5337](https://redhat.atlassian.net/browse/OSAC-5337), the actual signing happens
inside the shared `osac-build-and-publish.yaml` reusable workflow both `nightly-build.yaml` and
`osac-release.yaml` call into -- so that's the one identity that now covers both, for every
artifact. The umbrella chart's dedicated verification block simplifies from an either/or
(`publish-osac-installer-chart.yaml` OR `nightly-build.yaml`) to a single identity, since there's
now only one publisher.
- Updated two stale comments in `lib.sh`/`nightly-charts.sh` that referenced the deleted
workflow by name.

## What I verified

Not just "the regex looks right" -- pulled a real `cosign` client and checked against the actual
live signatures from today's `osac v0.0.10` release:

- The **old, still-documented** commands fail: `cosign verify` against the real umbrella chart
signature with the old `(publish-osac-installer-chart.yaml|nightly-build.yaml)` regex returns "no
matching attestations", and cosign's own error message reveals the real identity is
`osac-build-and-publish.yaml@refs/heads/main`.
- Same result for a real image (`osac-csi-driver:0.0.2`) -- old regex rejects it, real identity is
the same `osac-build-and-publish.yaml@refs/heads/main`.
- The **new, corrected** commands in this PR pass against both the real chart and the real image
signature.
- Confirmed no other workflow references `publish-osac-installer-chart.yaml` (e.g. via
`workflow_run`) before deleting it.
- Confirmed `check_chart_published`/`check_umbrella_mono_repo_charts_published` have zero other
callers repo-wide before removing them; confirmed the generic `http_json`/`http_retry` helpers
they used are still needed elsewhere (`check_osac_ui_image`) and left untouched.
- `bash -n` on both modified shell scripts.

## Summary

- **CI/deployment:** Removed the obsolete `publish-osac-installer-chart.yaml` workflow. `osac-release.yaml` now provides the replacement release path.
- **Scripts:** Removed unused chart publication checks from `nightly-charts.sh`. Updated stale workflow references and comments.
- **Documentation:** Updated Cosign verification commands for current shared-workflow signer identities. Documented `main` and manually dispatched non-`main` ref requirements.
- **Other areas:** No API, controller, database, authentication, or runtime changes.
- **Compatibility:** Manual dispatches of the removed workflow are no longer available. No other compatibility impact is supported by the evidence.
- **Verification:** Reported checks include live image and chart signature checks, repository-wide reference checks, and `bash -n` validation.
- **Review findings:** Severity counts are unavailable because current review findings were not supplied.

## Risk classification

The applied risk label is unavailable. The supplied evidence does not identify an applied `risk:ship`, `risk:show`, or `risk:ask` label or provide the labeling criteria needed to classify the change. A near classification is also unavailable.


eliorerz:
Verified the finding against current code: real gap. The `workflow_dispatch`-ref caveat was stated once before both chart code blocks, but the umbrella-specific block ("only ever has one signer... verify with just: `refs/heads/main`") read as a standalone, unqualified recipe — someone jumping straight to it for a chart published by a manually-dispatched `osac-release.yaml` run against a non-`main` ref would use the wrong regex.

Fixed: split the umbrella section into the common case (nightly/main-dispatched release) and an explicit follow-up sentence covering a non-`main` dispatch, so the block is self-contained instead of relying on a caveat from three paragraphs earlier. Amended and force-pushed.

eliorerz:
Verified the finding — it was somewhat garbled (mixed "image verification" wording with a `publish-charts.yaml`-only concern), but there was a real, valid issue underneath in two places:

1. **Images section**: the "Verify an image" examples never showed the `osac-build-and-publish.yaml` alternate identity at all, even though the prose right above claims it's valid for any image in the table. Added a second worked example.

2. **Charts section**: the general verify command's regex hardcodes `refs/heads/main` for *both* alternatives, but `publish-charts.yaml` and `osac-build-and-publish.yaml` don't have the same ref behavior — `publish-charts.yaml` only ever runs via `workflow_run` off `main` (fixed, permanent), while `osac-build-and-publish.yaml`'s ref follows whatever ref a manual dispatch used. Added a note making that distinction explicit, so only the `osac-build-and-publish.yaml` half needs adjusting for a non-`main` dispatch.

**Something the verification process itself caught, worth flagging:** my first attempt used `osac-operator` as the new image worked example, copying the pattern of the existing `build-image.yaml` example right above it. It didn't verify — turned out `osac-operator`'s specific release-run image was a false negative caused by a real, interesting coincidence: a routine `push to main` landed at the *exact same commit* our release dispatched against, moments earlier, and independently triggered `build-image.yaml`'s own build. Since the build is reproducible, both runs produced a byte-identical digest, and only one signature was discoverable for it. Switched the example to `metering-echo-adapter` (no such collision) and empirically confirmed it verifies. Added a note explaining this collision case can happen, so a reader who hits it with a real image isn't confused by a "no matching attestations" error on a legitimately-signed artifact.

Re-verified all four commands (image × 2 identities, chart × 2 identities) against real live signatures before pushing. Amended and force-pushed.

eliorerz:
Verified — real, valid gap. Rewrote the regex so each alternative has its own ref constraint directly, instead of one shared `refs/heads/main` plus a prose instruction to hand-edit it: `publish-charts\.yaml@refs/heads/main` (fixed, since it only ever runs via `workflow_run` off `main`) vs `osac-build-and-publish\.yaml@refs/.+` (any ref, since nightly/release can run from a different branch or tag). Empirically re-verified: passes against the real umbrella chart signature, and correctly still rejects an unrelated identity (an image signed by `build-image.yaml`) — confirms it isn't overly permissive. Amended and force-pushed.

eliorerz:
Verified — and this one caught a real overreach in my own earlier wording. I'd claimed "only the signature from whichever workflow actually ran is discoverable... not both," based on a cosign verify test that returned one match for an OR'd identity regex. Went back and queried Rekor's public index directly for the exact digest in question — the entry it found is a DSSE/bundle-format entry (the newer sigstore bundle style, same as what we found for chart signing earlier), not a simple hashedrekord. That format isn't reliably indexed by container digest the way I was assuming, which means I can't actually prove a second signature doesn't exist -- my own test had a real blind spot. Adopted the safer, more defensible framing: both identities may validly sign the same digest; check both before concluding an artifact isn't signed. Amended and force-pushed.

eliorerz:
Verified — real, valid gap. The sentence conflated nightly's "always rebuild everyone" behavior with release's "only rebuild what's selected" behavior (OSAC-5337's pin-by-default model), incorrectly implying both unconditionally rebuild every image in the table. Rewrote to distinguish them explicitly: nightly always rebuilds everything; osac-release.yaml only rebuilds components actually named in that release's component_versions (and not already at the requested version) — everything else stays pinned to its existing image, untouched and not resigned. Kept the shared-identity point, since it's still correct whenever a rebuild does happen through either path. Amended and force-pushed.

PR #1014, opened 2026-09-16
NO-ISSUE: build and publish the OSAC CLI alongside fulfillment-service

eliorerz:
## Summary

Folds a `linux/amd64` build of the `osac` CLI into `fulfillment-service`'s own build step in
`osac-build-and-publish.yaml`, so every nightly run and every release that actually rebuilds
`fulfillment-service` produces a fresh CLI artifact automatically -- for tests and local
development, not as a separate mechanism that needs to be kept in sync.

## Why

`publish-binaries.yaml` (the official, cross-platform, goreleaser-based CLI release) only fires
on a real `fulfillment-service/vX.Y.Z` tag push. That's correct for real releases -- and already
happens automatically whenever `fulfillment-service` is released, whether standalone or bundled
via `osac-release.yaml`'s `component_versions` (the tag push fires it either way, no extra work
needed there). But nightly never creates a real tag, so nightly builds have never had a
correspondingly fresh CLI artifact of their own.

## What changed

- New steps in the `build` job's `fulfillment-service` path (gated on `matrix.name ==
'fulfillment-service' && steps.resolve.outputs.skip != 'true'`, so a pinned/not-rebuilt
fulfillment-service produces no new artifact either):
  - `Set up Go for OSAC CLI build` -- reuses the existing `./.github/actions/setup-go` composite
  action.
  - `Build OSAC CLI (linux/amd64)` -- plain `go build` with the same version `ldflags` the
  official goreleaser release uses.
  - `Sign OSAC CLI with cosign` -- `cosign sign-blob --bundle`, the same keyless flow
  `fulfillment-service/.goreleaser.yaml`'s own `signs:` config already uses.
  - `Upload OSAC CLI artifact` -- a workflow artifact (14-day retention, matching the existing
  nightly chart artifact), not a GitHub Release. Avoids cluttering the Releases page with
  ephemeral nightly entries; nightly stuff stays throwaway, consistent with everything else this
  pipeline produces.
- `publish-binaries.yaml` is untouched -- still the official cross-platform release path for real
tags.

## What I verified

- YAML parses cleanly; `bash -n` on every embedded script in the modified workflow.
- Confirmed `fulfillment-service/internal/version.id` (the `ldflags` target) and `cmd/osac`
actually exist.
- Actually ran the exact `go build` invocation locally with the same flags -- compiles cleanly,
and the resulting binary correctly reports its embedded version (`osac version` → the injected
value).
- Confirmed `cosign sign-blob --bundle` is the right flag via `cosign --help`, matching the
existing goreleaser config's own usage.

## Changes

- **CI:** Adds an OSAC CLI build to `.github/workflows/osac-build-and-publish.yaml`.
- The build targets `linux/amd64` and embeds the resolved version.
- Cosign signs the binary with a keyless Sigstore bundle.
- The workflow uploads the binary and signature for 14 days.
- `publish-binaries.yaml` remains unchanged.

## API surface

No public API changes.

## Other areas

No changes are reported for controllers, database, authentication, deployment, tests, or documentation.

## Compatibility

No runtime or release-tag behavior changes are expected. The change adds CI artifacts only.

## Risk classification

A risk label cannot be determined. The supplied evidence does not include the labeling criteria or an applied label. Review findings and test results are also unavailable.


eliorerz:
Superseded by #1015 -- baking the CLI directly into the fulfillment-service image is strictly better than this workflow-artifact approach: one artifact instead of two, no separate consumer to wire up (this PR's artifact had none), and CI can extract it from whatever fulfillment-service image it's already using rather than depending on nightly/release having recently run. Closing this in favor of #1015 + osac-test-infra#557.

PR #1024, opened 2026-09-16
OSAC-5357: Fix release pipeline tagging components without publishing them

eliorerz:
## Summary

The release pipeline's "Tag and notify" job (`osac-build-and-publish.yaml`) could create a real `<component>/vX.Y.Z` release tag without the corresponding image/GitHub release ever getting published, while still reporting the overall run as successful.

**Root cause:** the tag was pushed using the default `GITHUB_TOKEN`. GitHub does not fire push-triggered workflows for a ref created by `GITHUB_TOKEN` (anti-recursion by design), so each component's own tag-triggered publish workflow (`publish-image.yaml`, `publish-binaries.yaml`, `publish-proto.yaml`, and the equivalent per-component workflows for every other component) silently never ran.

Confirmed live in run [35026075538](https://github.com/osac-project/osac/actions/runs/35026075538): tag `fulfillment-service/v0.0.99` was created and the run reported `success`, but no `ghcr.io/osac-project/fulfillment-service:v0.0.99` image and no GitHub Release exist for it.

**Audit:** this gap is not specific to `fulfillment-service` — it affects every `component_versions` component (`osac-operator`, `osac-aap`, `bare-metal-fulfillment-operator`, `osac-metering`, `osac-csi-driver`), since all of them publish exclusively via a `push: tags:` trigger on their own component-scoped tag.

## Changes

- `.github/workflows/osac-build-and-publish.yaml`: "Tag and notify" now pushes tags authenticated as `secrets.OSAC_BOT_PAT` instead of the default `GITHUB_TOKEN`, so the push behaves like a real actor and correctly fires downstream publish workflows (same PAT already used in `pr-dashboard.yml` for this exact class of `GITHUB_TOKEN` limitation).
- New step "Verify component publish workflows actually ran": after tagging, waits for and requires every downstream publish workflow mapped to each tagged component to have started and succeeded — failing the run loudly if one doesn't. Job timeout bumped 15→90 minutes to allow for real image/binary builds.
- `osac-installer/scripts/nightly-charts.sh`: added the component→workflow mapping and the poll/wait helpers backing the verification step.
- `.github/AGENTS.md`: documented the new release-safety invariant.

## Out of scope (flagged, not fixed here)

This was the first-ever release through this pipeline, and it ran with `skip_e2e=true` — nothing would have caught this gap even with a working publish, since e2e was skipped. Worth a separate discussion on whether `skip_e2e` should be restricted for real release runs.

## Test plan

- [x] `pre-commit run --all-files` (yamllint --strict, gitleaks, etc.) passes
- [x] YAML syntax validated (`python3 -c "import yaml; yaml.safe_load(...)"`)
- [x] Shell syntax validated (`bash -n nightly-charts.sh`)
- [x] New helper functions (`_component_publish_workflows`, `wait_for_component_publish_workflow_run`, `verify_component_publish`) unit-tested locally against a mocked `gh` CLI: success path, never-started timeout, and failed-run path all behave correctly (loud `::error::` + non-zero exit on failure)
- [ ] End-to-end verification: dispatch `osac-release.yaml` bumping a component version and confirm the resulting tag gets a real published image and GitHub Release (recommend running this from this branch before merge, since it exercises real CI infra)

## Summary

- **CI and deployment:** Release tags now use `secrets.OSAC_BOT_PAT` so downstream publish workflows can start.
- **CI verification:** Component mappings and polling helpers require each mapped publish workflow to start and succeed.
- **CI timeout:** Verification timeout increases from 15 to 90 minutes.
- **Security:** Job permissions use `contents: read`.
- **Validation:** Unknown components now fail verification.
- **Documentation:** Release safety rules require actor-created tags and publish-workflow verification.
- **API surface:** No public API changes. Internal shell helpers were added.
- **Tests:** YAML parsing, shell syntax, and mock tests passed. Real end-to-end verification remains pending.
- **Other areas:** No controller, database, or application changes.

## Backward compatibility

Release jobs now require `secrets.OSAC_BOT_PAT`. Releases fail when a mapped publish workflow is missing or unsuccessful. `skip_e2e` behavior is unchanged.

`fulfillment-service` releases remain blocked by the known `publish-proto.yaml` failure caused by the missing Buf repository reported in OSAC-5216.

## Risk classification

No risk label or labeling criteria were supplied. The applied label is unavailable. The change is close to `risk:ask` because end-to-end verification is pending and `fulfillment-service` has a known publish failure, but the supplied evidence does not establish the criteria for that label.


eliorerz:
Verified this against the actual codebase before reviewing -- most of it checks out:

- `secrets.OSAC_BOT_PAT` genuinely exists and is used identically in `pr-dashboard.yml` -- not a made-up secret name.
- `needs.prepare.outputs.components-to-tag` / `component-versions` genuinely exist with those exact names/shapes on `prepare`, already consumed the same way elsewhere in this file.
- `_gha_sanitize_for_message` is a real, pre-existing helper in `nightly-charts.sh`, reused correctly, not invented.
- All six mapped workflow filenames in `_component_publish_workflows` exist with exactly the `<component>/v*` tag patterns claimed.
- The core fix (checkout with a PAT instead of the default `GITHUB_TOKEN` for the tag push) is correct and directly addresses the real root cause: I independently confirmed a `GITHUB_TOKEN`-authored tag push doesn't fire `push`-triggered workflows, while a manual `git push` with real credentials does (verified live with the `fulfillment-service/v0.0.100` workaround tag).

**One confirmed blocker, though:** `_component_publish_workflows` requires `publish-proto.yaml` to succeed for `fulfillment-service`, via `verify_component_publish` -> `wait_for_component_publish_workflow_run`, which hard-fails the whole job if that run's conclusion isn't `success` (no partial-success tolerance). But `publish-proto.yaml` has failed on **every single** `fulfillment-service/v*` tag push since at least v0.0.92 (checked v0.0.92 through today's v0.0.100 -- 10/10 failures), due to an already-open, unrelated bug: [OSAC-5216](https://redhat.atlassian.net/browse/OSAC-5216) (`buf push` to `buf.build/osac-project/tests` 404s every time -- that repo was never created on the Buf registry).

As written, this means merging this PR would make *every future release that bumps fulfillment-service* fail at this new verification step, unconditionally, until [OSAC-5216](https://redhat.atlassian.net/browse/OSAC-5216) is fixed separately -- trading today's silent-success bug for a guaranteed-failure one. Given fulfillment-service is the component that's actually being released in practice, this would immediately block the very next real release.

Suggest either:
1. Fix [OSAC-5216](https://redhat.atlassian.net/browse/OSAC-5216) first and merge this after, or
2. Don't hard-require `publish-proto.yaml` in the required set yet -- treat it as best-effort/warn-only for now, with an explicit comment linking [OSAC-5216](https://redhat.atlassian.net/browse/OSAC-5216), and make it a hard requirement once that's resolved.

Everything else here looks solid and ready once that's addressed.

eliorerz:
Verified both CodeRabbit findings against the actual code -- both confirmed real, fixed.

**`contents: write` unused (least privilege) -- confirmed, fixed.** Checked: the checkout step overrides `token:` with `secrets.OSAC_BOT_PAT` and leaves `persist-credentials` at its default (`true`), so that PAT -- not `GITHUB_TOKEN` -- is what authenticates every `git push` in this job (both "Create umbrella tag" and "Create per-component release tags"). Checked the rest of the job too (chart-manifest download, Slack notify) -- nothing else writes to repo contents via `GITHUB_TOKEN`. Dropped `contents: write` to `contents: read`.

**`verify_component_publish` silently succeeds for an unknown component -- confirmed real, fixed.** `_component_publish_workflows`'s `return 1` for an unrecognized component was being lost: it was piped into the loop via `done < <(_component_publish_workflows "${component}")`, and an unknown component produces empty stdout (only writes to stderr) -- so the `while read` loop just iterates zero times and the function falls through to an implicit `return 0`. Reproduced this locally before fixing: calling with a bogus component name printed the `::error::` line but returned exit status 0. Fixed by capturing `_component_publish_workflows`'s output into a variable first and checking its exit status explicitly, before feeding the captured string to the loop via a herestring instead of process substitution.

Tested 3 scenarios with a mocked `wait_for_component_publish_workflow_run`: unknown component (now correctly returns 1, zero downstream calls made), a known multi-workflow component like `fulfillment-service` (still correctly calls all 3 mapped workflows, returns 0), and a known component whose downstream run fails (still correctly returns 1). All behave as intended.

Validated: YAML parse on the workflow file, `bash -n` on the script, functional tests above. Pushed as a new commit.

PR #1025, opened 2026-09-16
OSAC-5152: Register cudn_evpn k8s manager via Helm values

bkopilov:
## [OSAC-5152](https://redhat.atlassian.net/browse/OSAC-5152): Register cudn_evpn k8s manager via an explicit E2E profile

**Jira:** [OSAC-5152](https://redhat.atlassian.net/browse/OSAC-5152)
**Epic:** [OSAC-5142](https://redhat.atlassian.net/browse/OSAC-5142) — K8s Manager Registration and Infrastructure

### Summary

Registers the `cudn_evpn` k8s manager through a complete, explicitly named `cudn-evpn-netris-test` OpenShift profile. The base chart and normal profiles remain unchanged.

### Review resolutions

- Removed the small `charts/osac/ci/cudn-evpn-values.yaml` fragment.
- Added `osac-installer/values/cudn-evpn-netris-test/` with complete infrastructure and instance values.
- The profile registers both `netris` as the fabric manager and `cudn_evpn` as the k8s manager.
- The profile sets `networkClass.fabricManager: netris` and `networkClass.k8sManager: cudn_evpn`, so the E2E deployment actually selects EVPN.
- Included the standard Netris AAP instance-group configuration; passwords, SSH keys, and site-specific secrets are supplied through a private values override.
- Added Makefile profile validation and documented `make install PLATFORM=openshift PROFILE=cudn-evpn-netris-test NS=osac`.
- Normalized underscores to hyphens in ConfigMap metadata names while preserving `data.name: cudn_evpn`.

### Scope and dependencies

The profile is opt-in and does not change the base chart or normal profiles. It assumes that Phase 1 IPv4 EVPN/BGP/VTEP prerequisites and the `cudn_evpn` implementation are already available; this PR only registers and selects the managers. Discovery and readiness consumers are explicit in [osac-project/osac#1038](https://github.com/osac-project/osac/pull/1038) and [osac-project/osac#1049](https://github.com/osac-project/osac/pull/1049).

### Testing

- JSON schema validation passed.
- Profile-specific Helm lint passed.
- Profile infrastructure and instance charts rendered successfully.
- Rendered profile contains both manager registrations, normalized metadata `osac-network-k8s-manager-cudn-evpn`, `data.name: cudn_evpn`, and the `netris`/`cudn_evpn` NetworkClass selection.
- The existing CaaS profile renders without `cudn_evpn`.
- YAML lint and `git diff --check` passed.
- Repository-wide `make helm-validate` remains blocked by the pre-existing default service hostname schema requirement when linting without profile hostname overrides.

### Acceptance Criteria

- [x] The existing network-manager schema validates the registration shape without duplicating the runtime value.
- [x] A complete named E2E profile contains normal installer values and explicit EVPN/Netris configuration.
- [x] The profile preserves both manager registrations and selects them on the default NetworkClass.
- [x] The ConfigMap metadata name is Kubernetes-valid and normalized to `cudn-evpn`, while `data.name` remains `cudn_evpn`.
- [x] Standard profiles are not implicitly opted into EVPN registration.

## Summary

- **Deployment:** Added the OpenShift-only `cudn-evpn-netris-test` profile for VMaaS and BMaaS. It registers and selects `netris` and `cudn_evpn` through the default `NetworkClass`.
- **Chart behavior:** ConfigMap metadata names now replace underscores with hyphens. The manager identity remains `cudn_evpn` in `data.name`.
- **Infrastructure:** Added values for operators, databases, Vault-compatible storage, AAP, Metal3, LVMS, and related services.
- **Documentation:** Documented installation, profile selection, private overrides, and prerequisites. External EVPN/BGP/VTEP infrastructure, an EVPN `FRRConfiguration`, the `cudn_evpn` implementation, and BareMetalHost inventory are required.
- **CI and validation:** Added Makefile profile validation. Schema, profile-specific Helm lint and rendering, YAML lint, and diff checks passed. Repository-wide `make helm-validate` remains blocked by an existing default service-hostname schema requirement.
- **API, controllers, database, and auth:** No direct API or controller code changed. The profile enables networking provisioning and configures database, Vault-compatible storage, Keycloak, CA issuer, Kafka, and AAP-related services.

## Compatibility

Base values and standard profiles remain unchanged. EVPN registration is opt-in through the new profile. Existing installations should not change unless they select this profile. Generated ConfigMap names now normalize underscores to hyphens.

## Risk classification

No applied risk label or risk-labeling criteria were supplied. The classification is unavailable. A comparison with another risk class cannot be established.


osac-installer/charts/osac/values.yaml
```diff
@@ -364,7 +364,11 @@ validation:
 #     role: k8s
 #     description: "CUDN LocalNet k8s manager"
 #     capabilities: "ipv4,ipv6,dualStack"
-networkManagers: []
+networkManagers:
+  - name: cudn_evpn
+    role: k8s
+    description: "OVN EVPN k8s manager for VM-to-fabric bridging (Phase 1: IPv4-only)"
+    capabilities: "ipv4"
```
danmanor: you need to add it to the schema not to the values

bkopilov (commented):
@danmanor Could you clarify what you mean by adding it to the schema? The networkManagers schema definition already exists in values.schema.json (lines 1485-1519) with the correct structure.

Did you mean:
1. Update the schema's `default: []` to include cudn_evpn as an example?
2. Move the cudn_evpn entry from values.yaml to only the schema default?
3. Something else?

The current approach follows the pattern where values.yaml contains the actual default values and the schema defines their structure/validation. Happy to adjust if there's a different convention for this project.

osac-installer/charts/osac/values.yaml
bkopilov: I've updated the schema's `default` array to include the cudn_evpn entry (commit d1548537f). This makes the schema self-documenting while keeping the entry in values.yaml as well (since Helm uses values.yaml as the actual default source).

If you meant something else, please let me know and I'll adjust. Thanks for the review!

danmanor:
Please address the following before merging:

1. Remove the networkManagers entry from the default array in values.schema.json. The schema should describe and validate values; the runtime default belongs in values.yaml, which should remain the single source of truth. Keeping the same default in both files creates unnecessary duplication and drift risk.

2. Normalize the ConfigMap metadata name. The installer template currently uses the raw manager name for metadata.name, so cudn_evpn renders as ...-cudn_evpn, which is not a valid Kubernetes resource name. Please use cudn-evpn for the resource name while preserving data.name: cudn_evpn as the manager identity. The operator chart already has this underscore-to-hyphen normalization precedent.

3. Please confirm the intended scope. Adding this to the base installer values exposes cudn_evpn in every profile, although Phase 1 requires manually prepared EVPN/BGP/VTEP infrastructure and the design says it should not be the default NetworkClass. It would be safer as an explicit profile or overlay opt-in. Also note that the BMaaS profile overrides the networkManagers array, so it would need to list both netris and cudn_evpn if that combination is intended.

The registration itself is the right mechanism: the operator discovers the labeled ConfigMap and resolves cudn_evpn by data.name. However, this PR only advertises the manager; it does not provide the corresponding EVPN implementation, so the implementation PR or dependency should be explicit.

bkopilov:
Addressed in commit `3d7fba2b0`:

- Removed the `cudn_evpn` entry from the schema default and kept the base `values.yaml` registration empty.
- Added `charts/osac/ci/cudn-evpn-values.yaml` as an explicit opt-in overlay for environments with prepared EVPN/BGP/VTEP infrastructure.
- Normalized ConfigMap metadata names with underscore-to-hyphen replacement while preserving `data.name: cudn_evpn`.
- Kept BMaaS `networkManagers` netris-only; it is not implicitly opted into EVPN.

This PR only advertises the manager. Discovery and readiness consumers are [#1038](https://github.com/osac-project/osac/pull/1038) and [#1049](https://github.com/osac-project/osac/pull/1049); the EVPN dataplane implementation remains a separate follow-up. Focused schema, Helm overlay render, YAML lint, and diff checks pass.

danmanor:
Thanks for addressing the schema, ConfigMap name, and default-scope concerns.

One remaining point about `charts/osac/ci/cudn-evpn-values.yaml`: if this is intended to drive an E2E deployment, I think it should become a complete, explicitly named profile such as `cudn-evpn-netris-test`, rather than only a small values fragment under `charts/osac/ci/`.

The current fragment only registers `cudn_evpn`. An actual E2E environment also needs:

- the normal instance and infrastructure values required by the installer;
- both manager registrations: `netris` as the fabric manager and `cudn_evpn` as the k8s manager;
- `networkClass.fabricManager: netris` and `networkClass.k8sManager: cudn_evpn`, otherwise the test will not select EVPN;
- the EVPN prerequisites and Netris configuration needed by the implementation;
- Makefile or CI wiring that actually installs this profile.

This is especially important because `networkManagers` is a list. A later values file replaces the earlier list, so applying the current EVPN-only overlay after a profile that contains `netris` would remove the Netris registration.

A dedicated `cudn-evpn-netris-test` profile can select EVPN only for this E2E environment while leaving the base chart and normal profiles unchanged. If this file is intended only as a Helm render fixture, please state that explicitly and keep the real E2E profile separate.

bkopilov:
Addressed in commit `261ab312a`:

- Removed `charts/osac/ci/cudn-evpn-values.yaml` and added the complete `osac-installer/values/cudn-evpn-netris-test/` profile with both `infra.yaml` and `instance.yaml`.
- The profile registers both `netris` as `role: fabric` and `cudn_evpn` as `role: k8s`.
- It sets `networkClass.fabricManager: netris` and `networkClass.k8sManager: cudn_evpn`.
- Added the standard Netris AAP configuration and documented private overrides for passwords, SSH keys, and site-specific values.
- Added Makefile support and documented the real install command: `make install PLATFORM=openshift PROFILE=cudn-evpn-netris-test NS=osac`.
- The base chart and normal profiles remain EVPN-free; the profile still assumes the external EVPN/BGP/VTEP prerequisites and the `cudn_evpn` implementation are available.

Validation passed for the new profile: Helm lint, infra/instance rendering, manager/NetworkClass assertions, schema parsing, YAML lint, and diff checks. `make helm-validate` reaches the existing default-hostname schema failure before profile validation.

bkopilov:
Follow-up in commit `0544f1699`: the dedicated profile now explicitly targets VMaaS. It enables VMaaS and CSI, disables CaaS/BMaaS, uses the VMaaS infrastructure profile with CNV, and retains the `netris` fabric + `cudn_evpn` k8s manager pair on the default NetworkClass. Normal profiles remain unchanged.

bkopilov:
Correction in commit `ce74fe93a`: VMaaS and BMaaS are now enabled together in `cudn-evpn-netris-test` (CaaS remains disabled). The profile enables CSI, restores the BMaaS/Metal3 backend configuration, and keeps the shared `netris` + `cudn_evpn` NetworkClass so VMs and bare-metal instances can attach to the same VPC. The target cluster must provide the Metal3/BareMetalHost inventory and EVPN/BGP/VTEP prerequisites.

bkopilov:
Follow-up in commit `e4184f2ee`: documented that this setup uses an EVPN `FRRConfiguration`. The profile now explicitly lists the FRR operator/CRD and `FRRConfiguration` as external prerequisites and states that the installer does not create them; it still registers/selects `netris` + `cudn_evpn` for the shared VMaaS/BMaaS NetworkClass.

danmanor:
Clarification on the registration path: I would like to keep `cudn_evpn` registered through the installer-level `networkManagers` list in this PR.

```yaml
networkManagers:
  - name: cudn_evpn
    role: k8s
    capabilities: "ipv4"
```

This is functionally sufficient for registration. The installer renders the labeled ConfigMap, and the operator discovers it by the label and resolves the manager using `data.name`. It does not require a duplicate `operator.networkManagers` entry. The current `k8s_only` registration path in the operator chart is inconsistent, and we can clean that up separately.

Please keep `cudn_evpn` unchanged as the logical manager name in the values, ConfigMap `data.name`, and `NetworkClass.k8sManager`. The underscore-to-dash conversion should apply only to the Kubernetes `metadata.name`, producing `...-cudn-evpn`, because Kubernetes names cannot contain underscores.

For the actual E2E profile, the installer-level list should contain both `netris` and `cudn_evpn`, and the profile should select them through the NetworkClass. Also, `operator.controllers.networkingProvisioning` needs to be enabled; otherwise the profile registers the managers but does not dispatch the AAP provisioning jobs.

bkopilov:
Addressed in commit `cd3b48c6e`:

- Kept `cudn_evpn` registration at the installer-level `networkManagers` list; no duplicate `operator.networkManagers` entry was added.
- Preserved `cudn_evpn` as the logical name in values, ConfigMap `data.name`, and `NetworkClass.k8sManager`. Only the Kubernetes `metadata.name` is normalized to `...-cudn-evpn`.
- The explicit E2E profile contains both `netris` and `cudn_evpn` and selects them through the NetworkClass.
- Enabled `operator.controllers.networkingProvisioning` in the EVPN profile so the operator dispatches the AAP networking provisioning jobs.

Focused YAML, Helm lint, rendering, and registration assertions pass.

danmanor:
I’m okay with keeping `cudn_evpn` as the logical manager ID for this PR. The underscore-to-dash conversion should apply only to the Kubernetes ConfigMap `metadata.name`; `data.name` and `NetworkClass.k8sManager` must remain `cudn_evpn`. This keeps the existing manager contract intact. A project-wide rename to `cudn-evpn` can be handled separately if needed.

Please also set `bmf.env.enableNetworkingProvisioning: "true"` since this profile includes BMaaS networking.

bkopilov:
Addressed in commit `12620f77b`:

- Kept `cudn_evpn` unchanged as the logical manager ID in the installer values, ConfigMap `data.name`, and `NetworkClass.k8sManager`.
- The installer template continues to normalize only the ConfigMap `metadata.name`, producing `...-cudn-evpn` for Kubernetes naming compliance.
- Set `bmf.env.enableNetworkingProvisioning: "true"` in the `cudn-evpn-netris-test` profile so BMaaS networking provisioning is enabled.

Validation passed with `git diff --check` and strict YAML lint for the updated profile.

danmanor (approved):

PR #1031, opened 2026-09-16
OSAC-5357: pass secrets to osac-build-and-publish.yaml via secrets: inherit

eliorerz:
## Summary

Follow-up to #1024. Reusable workflows (`workflow_call`) never inherit the caller's secrets automatically -- `osac-build-and-publish.yaml`'s "Tag and notify" job checks out the repo with `token: ${{ secrets.OSAC_BOT_PAT }}` unconditionally (both nightly and release mode), but neither `osac-release.yaml` nor `nightly-build.yaml` declared `secrets: inherit` on the job that calls it. That secret resolved to empty at runtime, and checkout failed outright with `Input required and not supplied: token`.

## How this was found

Manually dispatched a real verification release (`osac/v0.0.11`, `component_versions=fulfillment-service=0.0.101`, `skip_e2e=true`) to confirm #1024's fix actually works end-to-end. Hit this failure in "Tag and notify": https://github.com/osac-project/osac/actions/runs/35094409539/job/104793700923

No tag was created by the failed run -- checkout fails before any tagging step runs, so this was caught safely before publishing anything. But since `nightly-build.yaml` calls the same reusable workflow the same way, and the checkout step's PAT requirement isn't gated by mode, the next scheduled nightly build (03:00 UTC) would have hit the identical failure on its umbrella tag step.

## Fix

Added `secrets: inherit` to the job in both `osac-release.yaml` and `nightly-build.yaml` that calls `osac-build-and-publish.yaml`.

## Test plan

- [x] YAML syntax validated on both files
- [ ] Re-dispatch the same verification release after this merges, confirm "Tag and notify" actually succeeds and the new "Verify component publish workflows actually ran" step behaves as intended

## Changes

- **CI/deployment:** Release and nightly workflows now pass only `OSAC_BOT_PAT` to `osac-build-and-publish.yaml`.
- **Secret handling:** The reusable workflow declares `OSAC_BOT_PAT` as a required secret.
- This fixes the empty-token failure in the “Tag and notify” checkout.
- No API, controller, database, documentation, or test-code changes were identified.

## Compatibility and verification

- The change replaces broad secret inheritance with explicit secret mapping.
- Callers must provide `OSAC_BOT_PAT`.
- YAML syntax was validated.
- End-to-end verification remains pending.
- Review finding counts are unavailable from the supplied evidence.

## Risk classification

- **risk:ship:** Applied because the change is limited to CI secret propagation, uses explicit secret access, and passed YAML syntax validation.
- The formal labeling criteria were not supplied, so the classification criteria cannot be verified against them.
- **risk:show** and **risk:ask** were not selected. No evidence indicates a product runtime change or an unresolved high-impact issue.


eliorerz:

`secrets: inherit` was functionally correct but broader than necessary: it passes the caller's entire secret set into the reusable workflow, when `osac-build-and-publish.yaml` only ever needs one (`OSAC_BOT_PAT`, for the tag push in "Tag and notify" -- every other secret reference in the file is `secrets.GITHUB_TOKEN`, which is always automatically available in every job regardless of inheritance, so nothing else depends on inherited secrets). Checked this before changing anything, to make sure removing `inherit` wouldn't silently break some other secret usage.

Declared `OSAC_BOT_PAT` as a required `workflow_call` secret on `osac-build-and-publish.yaml`, and both `osac-release.yaml` and `nightly-build.yaml` now pass it explicitly instead of inheriting everything.

Validated: YAML parse on all three modified files.

PR #1035, opened 2026-09-16
OSAC-3141: implement block storage metering

omer-vishlitzky:
## Summary

Implement [OSAC-3141](https://redhat.atlassian.net/browse/OSAC-3141) block storage metering on top of merged PR #1023 ([OSAC-5325](https://redhat.atlassian.net/browse/OSAC-5325) project attribution).

- Add authoritative Volume lifecycle timestamps through proto, CRD, operator, and feedback.
- Add strict Volume lifecycle mapping, block-volume eligibility, deletion-request boundaries, Watch registration, projection/reconciliation loading, heartbeat support, and storage adapter routing.
- Add focused mapper, transition, Watch, CSI, and reconciliation coverage.

PR #1023 is already merged into `main`; this branch contains only the two [OSAC-3141](https://redhat.atlassian.net/browse/OSAC-3141) commits.

## Validation

- `make test` in `osac-metering`
- Proto lint
- `make lint` in metering-service, adapters, schema, and osac-csi-driver
- CSI and fulfillment Volume unit tests
- Compile-only tests across operator, fulfillment-service, and CSI driver
- Operator CRDs regenerated

The operator envtest suite could not start locally because `bin/k8s/1.31.0-linux-amd64/etcd` was unavailable.

### API surface
- Adds Volume lifecycle timestamps and provisioned capacity to the private proto.
- Adds the `volume` resource type and lifecycle usage fields.
- Changes Volume capacity validation from immutable to increase-only.
- Updates mapper interfaces and adds volume metering helpers.

### Controllers and metering
- Adds Volume lifecycle reconciliation, billability checks, transition validation, and deletion boundaries.
- Adds Volume Watch filtering, retry handling, projection loading, scaling events, and heartbeat usage.
- Routes Volume events to `/storage/event`.
- Validates vendor responses and preserves vendor identity.
- Persists `Deleting` and `Deleted` phases with authoritative timestamps.

### Deployment and generated artifacts
- Regenerates the Volume CRDs.
- Updates operator and fulfillment-service integration points.

### Tests and validation
- Adds coverage for lifecycle mapping, transitions, Watch behavior, scaling, heartbeats, reconciliation, CSI behavior, and update invariants.
- Reported validation includes metering tests, proto linting, affected-component linting, CSI and fulfillment Volume tests, compile-only tests, and CRD regeneration.
- The operator envtest suite did not run because the required etcd binary was unavailable.

### Backward-compatibility
- Existing mapper implementations must accept the new `CloudEventType` parameter.
- Volume updates can now increase capacity after provisioning.
- Volume events now use the storage endpoint instead of being rejected.
- Fulfillment updates reject unsupported parent masks and vendor volume ID changes.
- Consumers and reconciliation clients must support Volume payloads and lifecycle fields.

## Risk classification

`risk:ask` applied because the labeling instructions and their criteria were not supplied. The available evidence shows broad cross-component changes across APIs, controllers, metering, reconciliation, and CRDs, but it does not define the required thresholds for `risk:ship`, `risk:show`, or `risk:ask`. A nearby classification cannot be determined without those criteria.


omer-vishlitzky:
@coderabbitai help

omer-vishlitzky:
@coderabbitai full review

osac-metering/metering-service/internal/events/volume.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	if eventType == privatev1.EventType_EVENT_TYPE_OBJECT_UPDATED &&
+		(result == eventBillableStart || result == EventSuspended) &&
+		!volumeHasBlockIdentity(m.volume) && m.CurrentState() == VolumeStateAvailable {
+		return "", ErrSkipTransition
+	}
+	if eventType == privatev1.EventType_EVENT_TYPE_OBJECT_UPDATED &&
+		result == EventSuspended && previousState == VolumeStateAvailable &&
+		!volumeHasBlockIdentity(m.volume) {
+		return "", ErrSkipTransition
```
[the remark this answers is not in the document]
omer-vishlitzky: Fixed in commit `2cf5a60a4`. The Watch mapping now carries the projection `PreviousBillable` state into Volume transition resolution and emits `suspended.v1` when an AVAILABLE Volume loses block eligibility. Added `TestVolumeEligibilityLossSuspendsThePreviousInterval`.

osac-operator/internal/controller/volume_controller.go
```diff
@@ -285,7 +294,11 @@ func (r *VolumeReconciler) handleDelete(ctx context.Context, vol *v1alpha1.Volum
 	log := ctrllog.FromContext(ctx)
 	log.Info("deleting volume")
 
-	vol.Status.Phase = v1alpha1.VolumePhaseDeleting
+	if vol.Status.Phase != v1alpha1.VolumePhaseDeleting {
+		vol.Status.Phase = v1alpha1.VolumePhaseDeleting
+		now := metav1.Now()
+		vol.Status.StateTransitionTime = &now
```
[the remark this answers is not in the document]
omer-vishlitzky: Fixed in commit `2cf5a60a4`. `handleDelete` now persists the Deleting phase and `StateTransitionTime` with `Status().Update` before removing the finalizer. The metering boundary remains the durable deletion timestamp.

omer-vishlitzky:
Addressed the reconciliation finding in `2cf5a60a4` as fail-fast behavior: if a Volume changes billability without an authoritative lifecycle transition, reconciliation returns an error instead of silently advancing the projection version. This follows the project constraint against inventing timestamps or fallback correction boundaries. Added a regression test covering an AVAILABLE Volume whose block identity changes.

masayag (changes requested):
## Request changes

PR #1035 partially implements OSAC-3141, but it is not ready to merge against the approved design:
https://github.com/osac-project/enhancement-proposals/blob/main/enhancements/OSAC-3141-metering-storage/design.md

### P1

1. **Deletion reconciliation returns an error after successful cleanup**
   `osac-operator/internal/controller/volume_controller.go:171-179, 333-343`
   `handleDelete` persists status, removes the finalizer, and then `Reconcile` attempts a second status update because it still sees a status diff. PR CI fails three deletion tests with UID precondition errors. Make status-update ownership exclusive to the deletion path and add regression coverage.

2. **Volume corrections are silently discarded**
   `osac-metering/metering-service/internal/reconciliation/reconciler.go:95-100, 190-194`
   `volume` is absent from `correctionResourceTypes`; `publishCorrections` returns successfully without publishing. Projection state is still advanced or deleted, so missed Volume events permanently lose billing corrections.

3. **The canonical GiB-second usage contract is not implemented**
   `osac-metering/schema/lifecycle.go:15-30`
   `osac-metering/metering-service/internal/heartbeat/generator.go:213-221`
   Volume events contain `size_gib` and `duration_seconds`, but not the design-required `usage` interval/cumulative object, fixed-point quantity, unit, precision, or explicit boundaries.

4. **Volume metering is enabled unconditionally**
   `osac-metering/metering-service/internal/watch/consumer.go:48-57`
   `osac-metering/metering-service/cmd/metering-service/main.go:252-255, 287-290`
   There is no resource registry or volume feature gate. Volumes are always watched, reconciled, heartbeated, and routed before the required lifecycle, correction, and adapter prerequisites are complete.

5. **Resize metering is not deliverable**
   `osac-operator/api/v1alpha1/volume_types.go:32-36`
   `fulfillment-service/internal/servers/private_volumes_server.go:278-293`
   `size_gib` remains immutable and the fulfillment API rejects changes. The approved design requires successful expansion to split usage intervals. The existing scaling branch only handles a hypothetical update and has no resize tests.

6. **The operator can report AVAILABLE without a valid block identity**
   `osac-operator/internal/controller/volume_controller.go:271-278`
   Empty vendor IDs or unsupported protocols are accepted before setting `Ready`; feedback then maps `Ready` to `AVAILABLE`. Validate the response and enforce write-once vendor identity.

7. **The M360 storage route bypasses its design gate**
   `osac-metering/adapters/cmd/m360-adapter/translate.go:23-30`
   The approved design assigns storage routing to OSAC-4285 and says it must not be enabled before that gate passes. This route is unconditional and has no storage adapter contract coverage.

8. **The required DELETED lifecycle state is not implemented in the CRD/operator path**
   `osac-operator/api/v1alpha1/volume_types.go:71-79`
   `osac-operator/internal/controller/volume_feedback_controller.go:128-137`
   The design requires vendor cleanup, persisted `DELETED`, feedback propagation, and finalizer removal as separate steps. The implementation only reports `DELETING` and removes the finalizer.

### P2

9. **Feedback can erase an authoritative transition timestamp**
   `osac-operator/internal/controller/volume_feedback_controller.go:140-143`
   A missing CR timestamp clears the remote persisted timestamp. During upgrades or races this can erase the billing boundary. Preserve the remote value when the CR has no timestamp.

### Validation

- Metering event, watch, reconciliation, and heartbeat tests pass locally.
- Proto compilation and generated-code checks pass.
- PR CI reports operator unit tests failed: 882 passed, 3 failed.
- Full local metering tests were blocked by unavailable Podman/PostgreSQL.
- Local operator envtest was blocked by the missing Darwin etcd asset.

masayag (commented):
Follow-up on the current head `03f09164dff1bc7867b3ad3adb2326d08baac206`:

- The deletion reconciliation issue is addressed by the latest `volume_controller.go` changes.
- The replacement operator unit-test run now passes.
- The remaining findings in the changes-requested review still apply: Volume corrections are silently discarded, the canonical GiB-second usage/correction contract is absent, volume metering is ungated, resize is not deliverable, AVAILABLE is not gated on valid block identity, the M360 storage route bypasses its design gate, the DELETED lifecycle state is missing, and feedback can clear an authoritative timestamp.

masayag (changes requested):
## Round 2 review: request changes

Reviewed current head `64ac0ec2b52fda1571da1293cc09d58195f42b61` against the approved OSAC-3141 design.

### Resolved since the previous review

- The deletion status-update regression is fixed. `Deleting` and `Deleted` are now persisted in separate reconciles before finalizer removal.
- Volume corrections are now registered.
- Canonical lifecycle usage fields and Volume heartbeat usage were added.
- `DELETED` lifecycle mapping, vendor-response validation, vendor identity validation, and parent status-mask rejection were added.
- The previous timestamp-clearing behavior was corrected.

### Remaining P1 findings

1. **Existing billable Volumes without `provisioned_size_gib` now fail metering**
   `osac-metering/metering-service/internal/events/volume.go:128-145`
   For `AVAILABLE`, `DELETING`, and `DELETED` block Volumes, the mapper replaces `spec.size_gib` with `status.provisioned_size_gib`. Existing CRs created before this field was added have the status value `0`; they therefore fail `ValidateBillingDimensions` and can make startup reconciliation fail. `VolumeReconciler.handleUpdate` returns immediately for an existing `Ready` Volume and does not backfill the field (`osac-operator/internal/controller/volume_controller.go:223-224`). Add an upgrade/backfill path or a safe compatibility fallback, with a pre-existing-volume test.

2. **Resize is still not implemented end to end**
   `osac-operator/api/v1alpha1/volume_types.go:32-36`
   `fulfillment-service/internal/servers/private_volumes_server.go:292-305`
   Both the CRD and fulfillment API still reject every `size_gib` change as immutable, and the operator has no expansion path. This contradicts the PR description and approved design, which require increase-only expansion with an effective timestamp and split usage intervals. The new scaling/usage helpers are unreachable through the supported API path.

3. **Missed resize reconciliation retroactively bills the new capacity**
   `osac-metering/metering-service/internal/reconciliation/reconciler.go:352-365, 826-849`
   When reconciliation sees a same-state Volume dimension change, it replaces `BillingDimensions` but leaves `BillableSince` unchanged and uses the lifecycle state timestamp as the boundary. Subsequent cumulative heartbeats calculate the entire interval with the new capacity. The correction is emitted with no affected interval. Reconciliation must use the committed capacity-change timestamp, close the old slice, and open the new one.

4. **The correction contract is still incomplete**
   `osac-metering/metering-service/internal/reconciliation/correction.go:34-53, 168-180`
   Corrections still lack the design-required `correction_id`, `source_event_id`, adjustment records, stable adjustment/provider identities, and usable affected intervals. `publishCorrections` passes `nil` for the interval. Volume corrections may now be published, but downstream consumers cannot apply them accurately or idempotently.

5. **Volume metering remains unconditionally enabled**
   `osac-metering/metering-service/internal/watch/consumer.go:48-57`
   `osac-metering/metering-service/cmd/metering-service/main.go:249-255, 287-290`
   There is still no resource registry or Volume feature gate. Startup reconciliation, Watch, projection, and heartbeat processing always activate Volume metering, bypassing the design's staged prerequisite and disable/re-enable behavior.

6. **The M360 storage route remains enabled before its design gate**
   `osac-metering/adapters/cmd/m360-adapter/translate.go:23-30`
   The approved design assigns the storage route to OSAC-4285 and explicitly defers routing until that integration gate passes. The route is unconditional in the current adapter and has no deployment gate or correction/usage contract enforcement.

### Validation

- Local targeted metering event/watch/reconciliation/heartbeat tests pass.
- Proto and schema tests pass.
- Operator packages compile.
- PR CI component tests and integration tests pass.
- `check-labels` is currently failing only because the required `lgtm` label is missing.

osac-metering/metering-service/internal/events/volume.go
```diff
@@ -0,0 +1,270 @@
...
+
+func VolumeTransitionTime(volume *privatev1.Volume, event *privatev1.Event) (time.Time, error) {
+	if event.GetType() == privatev1.EventType_EVENT_TYPE_OBJECT_UPDATED && volume.GetMetadata().GetDeletionTimestamp() != nil {
+		return volume.GetMetadata().GetDeletionTimestamp().AsTime(), nil
+	}
+	return ResolveTransitionTime(
+		event.GetType(),
+		event.GetTimestamp(),
+		volume.GetMetadata().GetCreationTimestamp(),
+		volume.GetStatus().GetStateTransitionTime(),
```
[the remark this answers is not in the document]
omer-vishlitzky: Resize provisioning is intentionally out of scope for OSAC-3141. This PR meters resize events when the storage/CSI path provides an authoritative committed-capacity boundary; it does not implement expansion. For missed reconciliation without that boundary, the metering path now fails closed and leaves projection state unchanged rather than retroactively billing the new capacity.

osac-csi-driver/pkg/driver/controller.go
```diff
@@ -178,6 +178,41 @@ func (c *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVolu
...
+	if req.GetVolumeId() == "" {
+		return nil, status.Error(codes.InvalidArgument, "volume ID is required")
+	}
+	if req.GetCapacityRange() == nil {
+		return nil, status.Error(codes.InvalidArgument, "positive required capacity is required")
+	}
+	capacityRange := req.GetCapacityRange()
+	requestedBytes := capacityRange.GetRequiredBytes()
+	if requestedBytes <= 0 {
+		return nil, status.Error(codes.InvalidArgument, "positive required capacity is required")
```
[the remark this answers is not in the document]
omer-vishlitzky: This PR does not implement CSI expansion or limit-only expansion semantics. Resize provisioning is owned by the storage/CSI work; OSAC-3141 only consumes authoritative resize events for metering.

osac-metering/metering-service/internal/watch/consumer.go
```diff
@@ -138,12 +138,30 @@ func (c *Consumer) consumeStream(ctx context.Context) (int, error) {
...
+			return nil
+		}
+		c.logger.Error(err, "event handling failed, retrying current event",
+			"event_id", event.GetId(), "delay", delay)
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-time.After(delay):
+		}
+		delay = min(delay*2, c.MaxDelay)
```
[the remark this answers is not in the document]
omer-vishlitzky: The current code does not retry handler errors forever: a mapping/data-quality error returns from consumeStream and the consumer reconnects. We intentionally do not quarantine and continue because the current Watch API has no acknowledgement or cursor contract; doing so would silently drop the unprocessed event. The upcoming durable Watch mechanism should define that behavior.

omer-vishlitzky:
Follow-up on the review disposition:

### Addressed

- **Missed Volume resize reconciliation:** reconciliation now fails closed when Volume billing dimensions change without an authoritative capacity boundary. It does not publish an incorrect correction or mutate projection state. Regression coverage verifies the existing projection remains unchanged.
- **[OSAC-4285](https://redhat.atlassian.net/browse/OSAC-4285) storage integration:** the existing `/storage/event` route now flattens the canonical Volume usage object into the M360 fields (`usage_semantics`, interval boundaries, quantity, unit, and precision), with adapter coverage.
- The earlier lifecycle findings around eligibility loss, deletion ordering, `DELETED`, vendor identity, parent status masks, and timestamp preservation were addressed in prior commits.

### Intentionally not addressed

- **Resize provisioning/CSI expansion:** [OSAC-3141](https://redhat.atlassian.net/browse/OSAC-3141) provides metering for resize events; it does not implement volume expansion. The CRD, fulfillment API, operator, and CSI expansion implementation remain owned by the storage/CSI resize work.
- **Canonical correction contract:** this is intentionally deferred. PR 959 explicitly states that networking corrections are not extended and are deferred to [OSAC-983](https://redhat.atlassian.net/browse/OSAC-983)/shared Part 1 durable event distribution. Its correction producer still uses the same existing scaffolding. Storage follows the same boundary rather than creating a resource-specific correction contract that would be replaced by the upcoming [OSAC-983](https://redhat.atlassian.net/browse/OSAC-983) work.
- **Legacy/pre-existing Volume backfill:** this is a greenfield deployment. We do not add migration or fallback behavior for Volumes lacking authoritative `provisioned_size_gib`; malformed authoritative state fails loudly.
- **Per-resource Volume feature gates:** selective metering is not supported. All resource meters are enabled together and prerequisite failures are surfaced rather than silently disabling Volume.
- **Watch quarantine/continue-after-error:** the current consumer fails fast and reconnects. The Watch API has no acknowledgement or cursor contract, so quarantining an unprocessed event and continuing would silently drop it.
- Unrelated vendor CSI overflow/TLS suggestions remain outside this metering change and are not part of this PR.

masayag (changes requested):
## Round 3 review: request changes

Reviewed current head `8e86b225a9f9aa6952f61fc5f214a88587dbf9f9`.

### Resolved in this revision

- M360 now flattens the canonical Volume usage fields.
- Reconciliation no longer mutates a projection or emits an invalid correction for an unobserved Volume capacity change.
- Prior lifecycle, deletion, identity, timestamp, and parent-mask findings remain addressed.

### Remaining blockers

1. **Legacy billable Volumes can still fail startup reconciliation**
   `osac-metering/metering-service/internal/events/volume.go:128-145`
   Existing `AVAILABLE`/`DELETING`/`DELETED` block Volumes created before `status.provisioned_size_gib` was introduced have that field as zero. The mapper replaces the valid `spec.size_gib` with zero and fails validation. `VolumeReconciler.handleUpdate` does not backfill existing `Ready` Volumes. The current code therefore needs an upgrade/backfill path, a safe compatibility fallback, or an explicit enforced greenfield deployment precondition and migration documentation.

2. **The correction contract required by the approved design remains intentionally deferred**
   `osac-metering/metering-service/internal/reconciliation/correction.go:34-53, 168-180`
   Corrections still lack `correction_id`, `source_event_id`, adjustment records, stable provider identities, and usable affected intervals; reconciliation passes `nil` for the interval. The approved OSAC-3141 design makes the correction/read-model contract a graduation prerequisite. Deferring it to OSAC-983 requires revising the approved design and OSAC-3141 graduation criteria; otherwise this PR cannot claim complete block-storage metering.

3. **Resize scope is inconsistent with the approved design and PR description**
   `osac-operator/api/v1alpha1/volume_types.go:32-36`
   `fulfillment-service/internal/servers/private_volumes_server.go:292-305`
   The current code still rejects every `size_gib` update, while the PR metadata says capacity validation is now increase-only and the approved design includes resize interval boundaries. If CSI/resize provisioning is intentionally owned elsewhere, update the design, PRD, and Jira scope to make OSAC-3141 metering-only; otherwise the implementation is incomplete.

4. **A single invalid Volume can block all metering reconciliation**
   `osac-metering/metering-service/internal/reconciliation/reconciler.go:350-355`
   The fail-closed Volume error propagates out of the entire reconciliation pass. At startup, `main.go:261-264` turns any reconciliation error into a service-exit failure. One Volume with an unobserved capacity boundary therefore prevents VM, CaaS, networking, and heartbeat recovery from starting. Isolate the Volume failure or gate only the Volume meter instead of taking down the whole metering service.

5. **Volume metering still bypasses the design’s resource registry/feature gate**
   `osac-metering/metering-service/internal/watch/consumer.go:48-57`
   `osac-metering/metering-service/cmd/metering-service/main.go:249-255, 287-290`
   Volumes are always loaded, watched, reconciled, and heartbeated. The approved design requires per-resource registry/gate behavior for staged rollout and disable/re-enable handling. The stated decision that selective metering is unsupported is a design divergence that must be documented and approved before merging.

6. **The M360 storage route is still unconditional**
   `osac-metering/adapters/cmd/m360-adapter/translate.go:23-30`
   The approved design assigns storage routing to the OSAC-4285 integration gate. Flattening the payload does not establish that gate. Either provide evidence that OSAC-4285’s storage contract is complete, or prevent `/storage/event` routing until it is.

### Validation

- Latest targeted metering event/watch/reconciliation/heartbeat tests pass.
- Proto, schema, and compile-only operator checks pass.
- PR component unit tests and pre-commit pass.
- `check-labels` remains failed because the required `lgtm` label is missing.

osac-metering/metering-service/internal/watch/consumer.go
```diff
@@ -53,7 +53,7 @@ func BuildFilter(vmaas, caas bool) string {
 	if caas {
 		parts = append(parts, "has(event.cluster)")
 	}
-	parts = append(parts, "has(event.external_ip)", "has(event.nat_gateway)")
+	parts = append(parts, "has(event.external_ip)", "has(event.nat_gateway)", "has(event.volume)")
```
masayag: **P1 — Volume metering is enabled with no registry/gate**

`has(event.volume)` is unconditional, unlike CaaS/VMaaS which honor `ENABLE_*`.

The approved design requires a shared `MeterResourceRegistry` and forbids Volume events until that registry, the correction consumer, and the lifecycle gates are in place. Disabling Volume must freeze this meter without taking down the process or other resource types.

Keep Volume off by default. Seed existing tokens from `ENABLE_*`, make `METERING_RESOURCES` additive, and reload Watch/loaders/heartbeat/adapter routes together.

osac-metering/metering-service/cmd/metering-service/main.go
```diff
@@ -249,8 +249,10 @@ func run(ctx context.Context, logger logr.Logger, cfg *config) error {
 	externalIPClient := privatev1.NewExternalIPsClient(grpcConn)
 	natGatewayClient := privatev1.NewNATGatewaysClient(grpcConn)
 	externalIPPoolClient := privatev1.NewExternalIPPoolsClient(grpcConn)
+	volumeClient := privatev1.NewVolumesClient(grpcConn)
 	reconciler := reconciliation.NewReconciler(computeClient, clusterClient, store, publisher, logger, cfg.heartbeatInterval)
 	reconciler.SetNetworkingClients(externalIPClient, natGatewayClient, externalIPPoolClient, cfg.deploymentID)
+	reconciler.SetVolumeClient(volumeClient)
```
masayag: **P1 — Volume is always reconciled**

`SetVolumeClient` always runs, then startup `Reconcile` is fatal on any error. Combined with the unconditional Watch filter, Volume cannot be staged or frozen independently of CaaS/VMaaS.

Wire this through the design's resource registry. If Volume is not enabled, do not construct the client, do not load volumes, and do not fail the process on Volume mapping errors.

osac-metering/adapters/cmd/m360-adapter/translate.go
```diff
@@ -26,6 +26,7 @@ var resourceTypeEndpoints = map[string]string{
 	schema.ResourceTypeClusterOrder:    "/caas/event",
 	schema.ResourceTypeExternalIP:      "/networking/event",
 	schema.ResourceTypeNATGateway:      "/networking/event",
+	schema.ResourceTypeVolume:          "/storage/event",
```
masayag, line 29: **P1 — `/storage/event` ships before OSAC-4285**

The approved design assigns this route to OSAC-4285 and says it must not be enabled before that gate. Core storage metering does not need M360 routing. Flattening canonical usage into M360 fields does not satisfy the gate.

Drop this mapping. Reject or ignore Volume events in the M360 adapter until OSAC-4285 confirms the storage contract.

osac-metering/metering-service/internal/events/volume.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+func VolumeBillingDimensions(volume *privatev1.Volume) (map[string]any, error) {
+	state := VolumeCurrentState(volume)
+	if _, ok := volumeStates[state]; !ok {
+		return nil, fmt.Errorf("%w: volume %s has unknown state %q", ErrDataQuality, volume.GetId(), state)
+	}
+	sizeGiB := volume.GetSpec().GetSizeGib()
+	if _, committed := volumeCommittedCapacityStates[state]; committed && volumeHasBlockIdentity(volume) {
+		sizeGiB = volume.GetStatus().GetProvisionedSizeGib()
```
masayag, line 145: **P1 — Pre-existing Ready volumes can fail all metering at startup**

For `AVAILABLE`/`DELETING`/`DELETED` block volumes this replaces `spec.size_gib` with `status.provisioned_size_gib`. That field is new and omitted when 0; `ValidateBillingDimensions` then rejects `size_gib <= 0`.

`VolumeReconciler.handleUpdate` returns immediately for `Ready`, so volumes created before this PR never get `ProvisionedSizeGiB`. `loadVolumes` returns on the first mapping error, and startup reconciliation is fatal — one such Volume can prevent VM, CaaS, and networking recovery from starting.

Fall back to `spec.size_gib` when provisioned size is unset, backfill Ready volumes, and isolate Volume mapping failures so they cannot take down the process.

osac-metering/metering-service/internal/reconciliation/reconciler.go
```diff
@@ -336,6 +350,9 @@ func (r *Reconciler) reconcileFulfillmentResources(ctx context.Context, fulfillm
 				}
 			}
 		} else if !events.DimensionsEqual(ps.BillingDimensions, fs.billingDimensions) {
+			if events.IsVolumeResourceType(fs.resourceType) {
```
masayag, line 345: **P1 — Unobserved Volume resize fails the entire reconciliation pass**

Leaving the projection unchanged is correct versus silently rebilling a new size. Returning this error from the whole pass is not: startup `Reconcile` treats any error as fatal, so one Volume blocks every other resource type.

Freeze this Volume row and continue. Do not fail-closed the metering service on a single Volume.

osac-metering/metering-service/internal/reconciliation/reconciler.go
```diff
@@ -781,12 +826,59 @@ func (r *Reconciler) loadNATGateways(ctx context.Context, result map[string]fulf
...
+		limit := int32(defaultPageSize)
+		resp, err := r.volumeClient.List(ctx, &privatev1.VolumesListRequest{Offset: &offset, Limit: &limit})
+		if err != nil {
+			return fmt.Errorf("listing volumes (offset=%d): %w", offset, err)
+		}
+		items := resp.GetItems()
+		for _, volume := range items {
+			dimensions, dimErr := events.VolumeBillingDimensions(volume)
+			if dimErr != nil {
+				return fmt.Errorf("mapping volume %s: %w", volume.GetId(), dimErr)
```
masayag, line 819: **P1 — First Volume mapping error aborts the whole load**

A single Volume with unknown state or zero committed capacity fails `loadVolumes`, which fails `Reconcile`, which fails process startup.

Skip/quarantine the bad Volume (and metric/log it). Do not make Volume data quality a process-wide fatal.

osac-operator/api/v1alpha1/volume_types.go
```diff
@@ -125,6 +126,15 @@ type VolumeStatus struct {
...
 	VendorContext map[string]string `json:"vendorContext,omitempty"`
+
+	// StateTransitionTime is the authoritative time at which the lifecycle
+	// phase last changed.
+	// +kubebuilder:validation:Optional
+	StateTransitionTime *metav1.Time `json:"stateTransitionTime,omitempty"`
+
+	// ProvisionedSizeGiB is the capacity last committed on the vendor array.
+	// +kubebuilder:validation:Optional
+	ProvisionedSizeGiB int64 `json:"provisionedSizeGiB,omitempty"`
```
masayag, line 161: **P1 — Resize is not deliverable**

`provisionedSizeGiB` is the committed-capacity field metering wants, but `spec.sizeGiB` is still `self == oldSelf` (immutable). Fulfillment still returns `spec.size_gib is immutable`, and Ready volumes have no expansion path.

The design and PRD require successful expansion to emit `updated.v1` and split usage. Watch has a synthetic resize test the API cannot produce. The PR summary claiming increase-only validation is false.

Implement increase-only spec + CRD + operator commit of this field, or revise the design/Jira scope and stop claiming resize in this PR.

fulfillment-service/internal/servers/private_volumes_server.go
```diff
@@ -256,28 +269,27 @@ func cloneVolume(v *privatev1.Volume) *privatev1.Volume {
...
 			// Unknown paths are handled by the generic update layer.
 		}
 	}
+	return nil
 }
 
-// validateVolumeImmutability checks that immutable spec fields have not been changed.
-// storage_tier, size_gib, and access_mode are immutable after creation because they are
-// provisioned directly into the vendor CSI call and cannot be modified post-creation.
+// validateVolumeImmutability checks fields that are immutable after creation.
```
masayag: **P1 — Size updates are still rejected**

This function still treats `spec.size_gib` as immutable. The PR summary claiming increase-only validation is false; proto comments still say the spec is fully immutable.

The design and PRD require successful expansion to emit `updated.v1` and split usage. Watch has a synthetic resize test the API cannot produce.

If resize is in OSAC-3141: change validation to increase-only and persist committed `provisioned_size_gib` on success. If it is not: revise the design/Jira scope and stop claiming resize in this PR.

osac-metering/metering-service/internal/events/volume.go
```diff
@@ -0,0 +1,305 @@
...
+	{StateEmpty, VolumeStateDeleted}:     {Skip: true},
+
+	{VolumeStateUnspecified, VolumeStateUnspecified}: {Skip: true},
+	{VolumeStateUnspecified, VolumeStateCreating}:    {Skip: true},
+	{VolumeStateUnspecified, VolumeStateAvailable}:   {EventType: eventBillableStart},
+	{VolumeStateUnspecified, VolumeStateFailed}:      {Skip: true},
+	{VolumeStateUnspecified, VolumeStateDeleting}:    {Skip: true},
+	{VolumeStateUnspecified, VolumeStateDeleted}:     {Skip: true},
+
+	{VolumeStateCreating, VolumeStateUnspecified}: {Skip: true},
```
masayag: **P2 — `CREATING → UNSPECIFIED` is skip, not reject**

The design marks this pair `reject` (stale/invalid regression). Skip updates projection instead of failing the event. `volume_transitions_test.go` only checks error vs skip, not event type.

Omit reject rows from the table so lookup fails, and assert started/updated/suspended/skip/error per ordered pair.

osac-operator/internal/controller/volume_feedback_controller.go
```diff
@@ -117,31 +117,50 @@ func (r *VolumeFeedbackReconciler) Reconcile(ctx context.Context, request ctrl.R
...
 // syncVolumePhase converts the CRD phase to the proto state enum.
 //
 // CRD Phase      -> Proto State
 // Progressing    -> CREATING   (volume is being provisioned on vendor array)
 // Ready          -> AVAILABLE  (vendor provisioned, ready for use)
 // Failed         -> FAILED     (vendor provisioning failed)
 // Deleting       -> DELETING   (volume is being deprovisioned)
-func syncVolumePhase(ctx context.Context, obj *v1alpha1.Volume, remote *privatev1.Volume) {
+// Deleted        -> DELETED    (vendor cleanup completed)
+func syncVolumePhase(obj *v1alpha1.Volume, remote *privatev1.Volume) error {
```
masayag, line 163: **P2 — Feedback still publishes `AVAILABLE` without a vendor ID**

New provisions are gated in `handleProvisioning`, but `syncVolumePhase` still maps `Ready → AVAILABLE` with no ID check. A Ready CR without an ID (legacy, partial status, or bug) still reports `AVAILABLE`.

Refuse to publish `AVAILABLE` unless `vendorVolumeID` is non-empty.

osac-operator/internal/controller/volume_controller.go
```diff
@@ -257,15 +262,29 @@ func (r *VolumeReconciler) handleProvisioning(ctx context.Context, vol *v1alpha1
...
 		vol.Status.Phase = v1alpha1.VolumePhaseFailed
+		now := metav1.Now()
+		vol.Status.StateTransitionTime = &now
 		setVendorProvisionedCondition(&vol.Status.Conditions, metav1.ConditionFalse, "ProvisioningFailed", err.Error())
 		return ctrl.Result{}, nil
 	}
+	if resp.VendorVolumeID == "" {
+		return r.failProvisioning(vol, "vendor CreateVolume returned an empty volume ID")
+	}
+	if resp.Protocol != string(v1alpha1.VolumeProtocolBlock) {
```
masayag, line 293: **P2 — Non-block vendor protocol is treated as a provisioning failure**

The design says NFS is **non-billable**, not unprovisionable. The CRD still allows `NFS`. If this controller serves NFS volumes, they now fail before Ready — a provisioning regression vs graduation's non-disruption requirement.

If block-only is intentional, tighten the CRD/protocol enum and tests. Otherwise provision NFS and leave it non-billable in metering.

masayag (changes requested):
## Request changes: design alignment review of `8e86b225a`

Reviewed against the approved design:
https://github.com/osac-project/enhancement-proposals/blob/main/enhancements/OSAC-3141-metering-storage/design.md

This HEAD is real progress on Volume lifecycle, but it does **not** deliver OSAC-3141 and should not ship as complete block-storage metering.

### Delivered

- Object-aware billability (`BLOCK` + vendor ID + `AVAILABLE` + no deletion timestamp) on Watch, loader, and heartbeat
- `OBJECT_CREATED` forced to `CREATING`; mapper rejects any other created state
- Deletion close uses `metadata.deletion_timestamp`, including a same-version Watch boundary
- Volume start/close/heartbeat payloads carry a canonical `usage` object
- Write-once `vendor_volume_id`; operator rejects empty IDs and ID replacement before `Ready`
- `VolumePhaseDeleted` with separate deleting → vendor delete → `Deleted` → later finalizer removal
- Feedback no longer clears a remote timestamp when the CR timestamp is missing
- List `limit=500`, no-progress failure, and `Get` confirmation before missed-deletion
- Project from `Metadata.project` (OSAC-5325 already on `main`)

### Remaining blockers

1. Volume metering is enabled with no registry/gate.
2. M360 `/storage/event` ships before OSAC-4285.
3. Volume events are on while the Part 1 correction contract is still scaffolding (`affected_interval=nil`, no adjustments).
4. Resize is not deliverable: CRD and fulfillment still treat `size_gib` as immutable. The PR text claiming increase-only validation is false.
5. A pre-existing Ready Volume with unset `provisioned_size_gib` can fail all metering at startup.

Inline comments cover these plus the Important follow-ups (transition table, feedback AVAILABLE without vendor ID, NFS treated as a provisioning failure, recon blast radius).

### Smallest honest merge slice

Keep Volume **off** until a registry/gate exists; remove `/storage/event`; fall back/backfill `provisioned_size_gib`; stop claiming resize unless the write path lands; rewrite `review fix` to a proper `OSAC-3141` commit.

**Not ready to merge.**

osac-metering/metering-service/internal/events/network.go
```diff
@@ -51,6 +51,13 @@ var requiredBillingDimensions = map[string][]string{
 		"tenant_id",
 		"project_id",
 	},
+	schema.ResourceTypeVolume: {
```
masayag: @omer-vishlitzky why is that added here? they are storage-related, not network-related.

osac-metering/adapters/cmd/m360-adapter/translate.go
omer-vishlitzky: OSAC-4285 is still in progress, but we are keeping the producer-side `/storage/event` route as an explicit integration exception. The canonical Volume usage fields are flattened and covered by adapter tests; the owning gate and remaining end-to-end completion are documented follow-up work.

osac-metering/metering-service/cmd/metering-service/main.go
omer-vishlitzky: This is the same whole-pass fail-fast reconciliation model already used for VMaaS/CaaS. We are not introducing a separate Volume-only failure policy or silently continuing with partial metering. The pending revision keeps the explicit fail-fast behavior.

osac-metering/metering-service/internal/watch/consumer.go
omer-vishlitzky: Responding with the agreed product decision: selective metering is unsupported. All resource meters are enabled together; we will not add a Volume-only registry gate in this PR. The design/scope documentation will record this intentional divergence from staged per-resource gating.

osac-metering/metering-service/internal/events/volume.go
omer-vishlitzky: No legacy backfill or fallback will be added. This is a greenfield deployment and authoritative `provisioned_size_gib` is a required contract for billable Volume states; invalid state fails loudly.

osac-operator/api/v1alpha1/volume_types.go
omer-vishlitzky: Resize provisioning is intentionally outside OSAC-3141. This PR meters resize events once storage/CSI provides the authoritative committed-capacity boundary; the storage team owns the increase-only API/operator/CSI implementation.

osac-operator/internal/controller/volume_feedback_controller.go
omer-vishlitzky: Addressed in the pending revision: feedback now refuses to map `Ready` to `AVAILABLE` when `vendor_volume_id` is empty, with regression coverage.

osac-metering/metering-service/internal/reconciliation/reconciler.go
omer-vishlitzky: The existing DLQ is adapter-consumer-only and requires an original Kafka ConsumerMessage. Reconciliation has no such message and no reconciliation quarantine contract. Invalid authoritative Volume state therefore remains an explicit reconciliation error rather than being silently quarantined.

fulfillment-service/internal/servers/private_volumes_server.go
omer-vishlitzky: The fulfillment size update path is intentionally not changed here. OSAC-3141 is metering-only for resize; storage/CSI owns making size updates possible and committing the authoritative capacity boundary.

osac-metering/metering-service/internal/events/volume.go
omer-vishlitzky: Addressed in the pending revision: `CREATING -> UNSPECIFIED` is removed from the transition table, so the explicit resolver rejects it. The transition test now asserts this invalid pair.

osac-metering/metering-service/internal/events/network.go
omer-vishlitzky: Addressed in the pending revision: Volume-specific dimension requirements and validation now live with Volume code. Networking files retain networking-only dimension requirements and validation; only genuinely shared dispatch/comparison helpers remain shared.

osac-operator/internal/controller/volume_controller.go
omer-vishlitzky: Metering intentionally meters block storage only. NFS remains a valid non-billable protocol; provisioning support for NFS is outside this metering change and does not require a CRD change here.

osac-metering/metering-service/internal/reconciliation/reconciler.go
omer-vishlitzky: The projection remains unchanged, which avoids incorrect rebilling. The process-wide fail-fast behavior is intentional and matches the existing VMaaS/CaaS reconciliation model; we are not introducing selective Volume isolation.

omer-vishlitzky:
Review disposition and implementation follow-up:

### Addressed in the pending revision

- `CREATING -> UNSPECIFIED` is now an explicit invalid transition, with test coverage.
- Feedback refuses to report `AVAILABLE` when `vendor_volume_id` is empty.
- Volume dimension validation now lives with Volume code; networking validation remains in networking code, with only shared dispatch/comparison helpers shared.
- Missed resize reconciliation remains fail-closed: it publishes no invalid correction and leaves projection unchanged.
- M360 storage usage flattening is implemented and covered.

### Intentional responses

- **No per-resource registry/gate:** selective metering is unsupported; all meters are enabled together. This differs from the staged-gate design and will be documented as an intentional product decision.
- **Whole-pass reconciliation failure:** this is the existing VMaaS/CaaS fail-fast model. We are not introducing a separate Volume-only failure policy.
- **Legacy `provisioned_size_gib` fallback/backfill:** this is greenfield. No migration or compatibility fallback is added.
- **Reconciliation DLQ:** the existing DLQ is adapter-consumer-only and handles Kafka messages after publication. It is not a reconciliation quarantine mechanism.
- **Resize provisioning:** [OSAC-3141](https://redhat.atlassian.net/browse/OSAC-3141) meters authoritative resize events; storage/CSI owns resize implementation.
- **NFS:** metering is block-only; NFS remains a valid non-billable protocol. Provisioning support is outside this metering change.
- **Correction contract:** deliberately deferred to OSAC-983/Part 1, matching PR 959, which explicitly deferred the same networking correction work. EP/PRD/Jira graduation scope must be revised accordingly.
- **M360 route:** [OSAC-4285](https://redhat.atlassian.net/browse/OSAC-4285) is still in progress, but we are keeping the producer-side `/storage/event` route as an explicit integration exception and documenting that the owning gate is incomplete.

Validation for the pending changes passes in metering service, adapters, schema, operator compilation, and targeted pre-commit checks.

omer-vishlitzky:
## Follow-up Review Disposition

The follow-up work separates Watch correctness from storage-owned lifecycle work and intentionally deferred reconciliation/integration work.

### Addressed

- Removed the deletion-timestamp fallback. A `DELETING` Volume without `metadata.deletion_timestamp` now fails with a data-quality error instead of using `state_transition_time`.
- Refactored Watch orchestration into preparation, stale-event handling, mapping, commit, and policy helpers.
- Removed the generic networking `MapperContext` from the events API.
- Added a resource-specific Watch mapper factory for ExternalIP pool enrichment and NAT deployment identity.
- Removed construction of unconfigured network mappers from the generic mapper factory.
- Moved usage generation behind `ResourceMapper`, rather than hardcoding Volume usage in generic event mapping.
- Added shared resource identity validation to scaling events, matching normal lifecycle events.
- Projection update failures in the Watch skip path now fail the event and reconnect instead of being logged and acknowledged.
- Added coverage for ExternalIP Watch mapping through pool-cache lookup, mapper dependencies, scaling identity validation, deletion timestamp requirements, and projection update failures.
- Updated all-resource constructor and Watch filter tests after removing selective service enablement.

### Intentionally Not Addressed

- **NFS:** This is block-storage-only. Protocol is immutable and NFS is outside the product contract. No NFS billing path is added.
- **Vendor-ID eligibility loss:** A Volume cannot legitimately change vendor identity mid-life. Write-once enforcement belongs to fulfillment/storage. The Watch path does not invent a normal billing transition for this impossible state.
- **Fulfillment write-once/output-only enforcement:** This is owned by the storage/fulfillment API boundary, not metering. It remains a separate storage concern.
- **Resize provisioning:** This PR meters authoritative resize events; it does not implement the Volume API, operator, or CSI expansion path. That remains storage/CSI scope.
- **Resize event dimension semantics:** The producer closes the old capacity slice with old dimensions and explicit usage. The arithmetic remains correct when the subsequent new-capacity interval is observed. Whether `updated.v1` must expose the new dimensions is an event-contract decision and should be resolved with the downstream consumer before changing this path.
- **M360 canonical usage validation:** Skipped as a separate adapter/integration concern. The producer emits canonical usage for usage-bearing Volume events; M360 storage contract ownership remains with OSAC-4285.
- **M360 storage route gate:** The producer-side `/storage/event` route remains an explicit integration exception pending OSAC-4285.
- **Correction identity/read-model contract:** Deferred to Part 1/OSAC-983 durable event and correction work. Reconciliation errors are intentionally fail-fast but are not used to redesign the soon-to-be-replaced delivery mechanism.
- **Per-resource feature gates:** Selective metering is intentionally unsupported. All resource meters are enabled together.
- **Reconciliation isolation:** A Volume reconciliation failure can still fail the reconciliation pass. This matches the existing fail-fast model and is acceptable because Watch accuracy is the priority and the reconciliation mechanism is temporary.
- **Watch quarantine/continue-after-error:** The consumer still fails fast and reconnects. The current Watch API has no acknowledgement or cursor contract; continuing would silently drop the failed event.
- **Legacy committed-capacity backfill:** Not added. This is a greenfield deployment with no migration or compatibility fallback.
- **Unrelated CSI vendor TLS/overflow findings:** Outside [OSAC-3141](https://redhat.atlassian.net/browse/OSAC-3141) metering scope.

### Validation

The follow-up work passes local `make test`, `make lint`, and `go test ./...` for `metering-service`.

osac-metering/metering-service/internal/events/dimensions.go
```diff
@@ -9,7 +9,21 @@ in compliance with the License. You may obtain a copy of the License at
...
+)
+
+// ValidateBillingDimensions rejects incomplete billing dimensions before an
+// event reaches Kafka. An empty project_id identifies the tenant default
+// project; all other string dimensions must be non-empty.
+func ValidateBillingDimensions(resourceType string, dimensions map[string]any) error {
+	if resourceType == schema.ResourceTypeVolume {
+		return validateVolumeBillingDimensions(dimensions)
+	}
+	return validateNetworkingBillingDimensions(resourceType, dimensions)
```
masayag, line 25: no need to check explicitly if this is netowrking resource type?
also, couldn't there be more than a single network resource type (e.g. External IP or NAT gateway?)
why is that restricted only for volume and network?

osac-metering/metering-service/internal/heartbeat/generator.go
```diff
@@ -180,6 +180,17 @@ func (g *Generator) buildHeartbeatEvent(state *projection.ResourceState, eventID
 		BillingDimensions: dims,
 		SchemaVersion:     schema.SchemaVersion,
 	}
+	if events.IsVolumeResourceType(state.ResourceType) {
```
masayag, line 183: I wouldn't expect to find this here.
When there is a need to augment the event with specific resource type data, that should be an abstraction for each resource type.
Why volume gets a special care here?

osac-metering/metering-service/internal/reconciliation/reconciler.go
```diff
@@ -201,15 +204,15 @@ func (r *Reconciler) reconcileFulfillmentResources(ctx context.Context, fulfillm
 
 	for id, fs := range fulfillmentState {
 		ps, exists := projMap[id]
-		if events.IsNetworkingResourceType(fs.resourceType) && fs.transitionTime.IsZero() {
-			sourceBillable, billErr := isBillableForType(fs.resourceType, fs.state)
+		if (events.IsNetworkingResourceType(fs.resourceType) || events.IsVolumeResourceType(fs.resourceType)) && fs.transitionTime.IsZero() {
```
masayag, line 207: should `events.IsNetworkingResourceType(fs.resourceType) || events.IsVolumeResourceType(fs.resourceType)` be extract to a function, such as `IsSupportedResourceType(fs.resourceType))` to avoid an `if` statement that gets bigger and bigger?

osac-metering/metering-service/internal/reconciliation/reconciler.go
```diff
@@ -277,6 +280,9 @@ func (r *Reconciler) reconcileFulfillmentResources(ctx context.Context, fulfillm
 		if fs.version > ps.FulfillmentVersion &&
 			ps.CurrentState == fs.state &&
 			events.DimensionsEqual(ps.BillingDimensions, fs.billingDimensions) {
+			if events.IsVolumeResourceType(fs.resourceType) && ps.IsBillable != fs.isBillable {
```
masayag, line 283: Can this happen only for storage or can it be generalized?

osac-metering/metering-service/internal/events/volume.go
```diff
@@ -0,0 +1,319 @@
...
+func (m *volumeMapper) CatalogItemID() *string { return nil }
+func (m *volumeMapper) TemplateID() *string    { return nil }
+
+func (m *volumeMapper) CurrentState() string {
+	return VolumeCurrentState(m.volume)
+}
+
+func VolumeCurrentState(volume *privatev1.Volume) string {
+	state := volume.GetStatus().GetState()
+	if state == privatev1.VolumeState_VOLUME_STATE_DELETED {
```
masayag, line 101: `DELETED` wins over `deletion_timestamp` for state, and no Watch test checks the close time.

`VolumeCurrentState` returns `DELETED` before looking at `deletion_timestamp`. `VolumeTransitionTime` uses `deletion_timestamp` only when current state is `DELETING`. A Watch payload with proto `DELETED` + `deletion_timestamp` therefore suspends at `state_transition_time`.

Happy path is still fine: the deletion request is observed as `DELETING` and closes at `deletion_timestamp`; `DELETING → DELETED` is skip. `TestVolumeDeletedStateWinsOverDeletionTimestamp` only asserts `VolumeCurrentState == DELETED` — it never asserts close time.

Design `AVAILABLE → DELETED` closes at `deletion_timestamp`. If the intended Watch contract is “Deleted-phase events are skip because DELETING already closed,” say that in the table and test it. Otherwise close at `deletion_timestamp` whenever it is present.

osac-metering/metering-service/internal/events/volume.go
```diff
@@ -0,0 +1,319 @@
...
+
+func IsVolumeTransientState(string) bool { return false }
+
+func (m *volumeMapper) CloudEventType(eventType privatev1.EventType, previousState string) (string, error) {
+	if eventType == privatev1.EventType_EVENT_TYPE_OBJECT_CREATED && m.CurrentState() != VolumeStateCreating {
+		return "", fmt.Errorf("volume OBJECT_CREATED must be CREATING, got %s", m.CurrentState())
+	}
+	if eventType == privatev1.EventType_EVENT_TYPE_OBJECT_UPDATED &&
+		previousState == VolumeStateAvailable && m.CurrentState() == VolumeStateAvailable && m.IsBillable() {
+		return eventBillableStart, nil
```
masayag: `AVAILABLE → AVAILABLE` is still three interpreters.

The transition table says Skip. This mapper override returns `eventBillableStart`. Watch `handlePreMappingScaling` emits `updated.v1` before mapping (`osac-metering/metering-service/internal/watch/policy.go`).

Watch tests pass because the consumer special-case runs first. `MapWatchEvent` alone would emit started/resumed. That is an implicit default around the table.

Have the table/mapper own `updated.v1` vs Skip, and remove the Volume-only pre-mapping branch.

masayag (commented):
Two remaining Watch-contract notes on the current head.

masayag (approved):
Approving `68b97c9aa`.

The two remaining Watch notes are addressed:
- `AVAILABLE → DELETED` (and any `OBJECT_UPDATED` with `deletion_timestamp`) now closes at `metadata.deletion_timestamp`, with `TestVolumeDeletedUpdateClosesAtDeletionTimestamp`.
- Same-state billable `AVAILABLE → AVAILABLE` is Skip from the transition table; the `eventBillableStart` mapper override is gone, with `TestVolumeAvailableToAvailableIsNotALifecycleBoundary`.

masayag (changes requested):
## Watch-focused review: request changes

Reviewed current head `f02af227f2cf06e71f301bb4746fec56fda0338e` under the clarified assumptions:

- greenfield only
- fail fast and loud, with no defaults or fallbacks
- all meters enabled
- reconciliation ordering/consistency issues are deferred
- Watch-stream correctness is the primary bar

I am not re-raising migration, selective-gating, correction-contract, or storage-owned resize-provisioning concerns here.

### P0

1. **The Watch filter uses a nonexistent Event field and will not compile**
   `osac-metering/metering-service/internal/watch/consumer.go:48-57`
   The filter uses `has(event.cluster_order)`, but the private Event proto declares the oneof field as `Cluster cluster = 4` in `proto/private/osac/private/v1/event_type.proto:75`. Fulfillment's CEL compiler therefore rejects the filter; the consumer repeatedly reconnects and never receives any events. The test currently asserts the incorrect literal instead of compiling it through the fulfillment Watch server. Use `has(event.cluster)` and add a server-side filter compilation test.

### P1

2. **NFS Volumes are treated as billable block storage**
   `osac-metering/metering-service/internal/events/volume.go:118-123, 178-182`
   `volumeHasVendorIdentity` checks only `vendor_volume_id`; it no longer checks `status.protocol == STORAGE_PROTOCOL_BLOCK`. An `AVAILABLE` NFS Volume with a vendor ID is therefore started, heartbeated, and emitted as `gibibyte_second` usage. The block-only meter must reject or mark NFS non-billable explicitly, with a regression test.

3. **Volume data-quality failures are silently acknowledged on Watch**
   `osac-metering/metering-service/internal/watch/event.go:35-43`
   `ErrDataQuality` from `BillingDimensionsMap` is converted into `skipped=true, nil`. With no Watch cursor/ack mechanism, an invalid Volume event is lost while the stream continues. This contradicts the stated fail-fast/loud policy and the design requirement not to invent or silently drop billing boundaries. Return the error so the consumer reconnects and exposes the failure.

4. **A DELETED update can still use the lifecycle timestamp instead of the deletion request boundary**
   `osac-metering/metering-service/internal/events/volume.go:99-107, 300-316`
   `VolumeCurrentState` returns `DELETED` before checking `metadata.deletion_timestamp`, while `VolumeTransitionTime` rejects only `DELETING` without that field. A malformed `OBJECT_UPDATED` snapshot with state `DELETED` and no deletion timestamp can therefore close usage at `state_transition_time`. Require the durable deletion timestamp for both `DELETING` and `DELETED` update boundaries.

### Validation

- Latest targeted metering event/watch/reconciliation/heartbeat tests pass locally.
- Proto, schema, and operator compile-only tests pass locally.
- PR CI unit tests and pre-commit pass.
- The current fulfillment-service integration job is failing in an unrelated tenant-deletion scenario; this review does not attribute that failure to the Volume Watch changes.

masayag (changes requested):
## Watch-focused review: request changes

Reviewed current head `ece8f8750048849a0043da20452bb55aebbe08e6` under the clarified scope. The latest revision fixes the prior Watch findings:

- CEL filter now uses `event.cluster`.
- NFS is excluded from Volume billability.
- Volume billing-dimension errors now fail the Watch event.
- `DELETED` without `metadata.deletion_timestamp` now fails the Watch event.

### Remaining P1 finding

1. **Non-Volume data-quality errors are still silently acknowledged**
   `osac-metering/metering-service/internal/watch/event.go:35-43`
   `prepareEvent` converts `ErrDataQuality` into `skipped=true, nil` for every resource type except Volume. The consumer then continues reading the stream, with no projection update or published event. The Watch API has no cursor or acknowledgement contract, so the event is permanently lost. This contradicts the explicit fail-fast/loud policy and creates a Watch reliability gap for existing VM, CaaS, and networking meters. Return the error for data-quality failures instead of acknowledging the event, or define and implement an explicit durable quarantine contract.

### Validation

- Latest targeted metering event/watch/reconciliation/heartbeat tests pass locally.
- Proto, schema, and operator compile-only checks pass locally.
- PR unit tests and pre-commit pass.
- The current fulfillment-service integration job is failing in an unrelated tenant-deletion scenario; `check-labels` is also failing because `lgtm` is missing.

osac-operator/internal/controller/volume_controller.go
```diff
@@ -267,14 +278,28 @@ func (r *VolumeReconciler) handleProvisioning(ctx context.Context, vol *v1alpha1
...
```
masayag: **P1 — Do not fail NFS provisioning from a metering PR**

On `main` this controller accepts `Block` and `NFS`. This change fails NFS `CreateVolume` and the operator test locks that in (`Entry("non-block protocol", ...)`).

The design says NFS volumes exist and are **non-billable**. The PRD says enabling storage metering must not disrupt provisioning. Billability is `protocol == BLOCK && vendor_id != "" && AVAILABLE && deletion_timestamp == nil` in the mapper — not a provisioner filter.

Keep the empty vendor-ID and write-once vendor-ID checks. Drop this protocol rejection so NFS can become Ready/AVAILABLE and stay unbilled.

osac-metering/metering-service/internal/events/volume_transitions_test.go
```diff
@@ -0,0 +1,38 @@
...
+				(from == VolumeStateFailed && (to == VolumeStateUnspecified || to == VolumeStateCreating || to == VolumeStateAvailable)) ||
+				(from == VolumeStateDeleting && (to == VolumeStateUnspecified || to == VolumeStateCreating || to == VolumeStateAvailable || to == VolumeStateFailed)) ||
+				(from == VolumeStateDeleted && to != VolumeStateDeleted)
+			if invalid {
+				if err == nil || errors.Is(err, ErrSkipTransition) {
+					t.Fatalf("expected invalid transition for %s -> %s, got %v", from, to, err)
+				}
+				continue
+			}
+			if err != nil && !errors.Is(err, ErrSkipTransition) {
```
masayag: **P1 — Assert the event type, not only error vs skip**

Reject rows are correctly omitted (lookup fails). This test still treats any non-error as success, including `ErrSkipTransition`.

`AVAILABLE → FAILED` could be changed to `{Skip: true}` and CI would pass. That is the silent-default hole these tables exist to prevent.

For every ordered pair, assert the exact result: `eventBillableStart` / `EventSuspended` / skip / error. Happy-path Watch tests are not a substitute.

osac-metering/metering-service/internal/events/volume.go
```diff
@@ -0,0 +1,319 @@
...
+	if eventType == privatev1.EventType_EVENT_TYPE_OBJECT_CREATED && m.CurrentState() != VolumeStateCreating {
+		return "", fmt.Errorf("volume OBJECT_CREATED must be CREATING, got %s", m.CurrentState())
+	}
+	result, err := ResolveCloudEventType(volumeTransitions, eventType, previousState, m.CurrentState())
+	if err != nil {
+		return "", err
+	}
+	if eventType == privatev1.EventType_EVENT_TYPE_OBJECT_UPDATED &&
+		(result == eventBillableStart || result == EventSuspended) &&
+		!volumeHasVendorIdentity(m.volume) {
```
masayag: **P1 — Missing vendor identity on a billing boundary must fail, not skip**

NFS never-billable is design-aligned: suppress billing, update projection.

`AVAILABLE` without a vendor ID is not. Feedback already refuses that report. If Watch still sees it, the payload is corrupt. `ErrSkipTransition` upserts projection and continues.

Return `ErrDataQuality` so the Watch stream fails loud. Do not invent a billing boundary, and do not quietly drop the event.

osac-operator/internal/controller/volume_feedback_controller.go
```diff
@@ -117,50 +117,74 @@ func (r *VolumeFeedbackReconciler) Reconcile(ctx context.Context, request ctrl.R
...
+		remote.GetStatus().SetState(privatev1.VolumeState_VOLUME_STATE_DELETED)
+	default:
+		return fmt.Errorf("invalid volume phase %q during deletion", obj.Status.Phase)
 	}
-	remote.GetStatus().SetState(privatev1.VolumeState_VOLUME_STATE_DELETING)
 	return nil
 }
 
+func syncVolumeStateTransitionTime(obj *v1alpha1.Volume, remote *privatev1.Volume) {
+	if obj.Status.StateTransitionTime == nil {
```
masayag: **P1 — Fail if `StateTransitionTime` is missing**

Unknown phase and empty vendor ID now error. A nil timestamp still returns.

That is a fallback: operator succeeds, fulfillment omits the field, Watch later dies on data quality. If this path is reporting a phase, the timestamp is required. Return an error here instead of omitting the field.

osac-metering/adapters/cmd/m360-adapter/translate.go
masayag: **P2 — Do not route storage to M360 until OSAC-4285**

The design gates `/storage/event` on OSAC-4285 confirming the storage contract. Core metering can publish to Kafka without this map entry.

Leave Volume out of `resourceTypeEndpoints` until that gate passes.

osac-metering/metering-service/internal/events/volume.go
```diff
@@ -0,0 +1,319 @@
...
+
+func (m *volumeMapper) TransitionTime(event *privatev1.Event, previousState string) (time.Time, error) {
+	if event.GetType() == privatev1.EventType_EVENT_TYPE_OBJECT_UPDATED &&
+		previousState == VolumeStateAvailable && m.CurrentState() == VolumeStateAvailable {
+		return VolumeCapacityTransitionTime(event)
+	}
+	return VolumeTransitionTime(m.volume, event)
+}
+
+func VolumeCapacityTransitionTime(event *privatev1.Event) (time.Time, error) {
```
masayag: **P2 — Resize consume is not resize delivery**

This Watch path correctly closes/opens on committed `provisioned_size_gib` when an event carries an authoritative timestamp. Production cannot emit that event: CRD `sizeGiB` is still `self == oldSelf`, fulfillment still treats `spec.size_gib` as immutable, and Ready volumes return immediately in `handleUpdate`.

Keep the consume path. Do not describe increase-only capacity as part of this PR. Resize provisioning is a follow-up.

masayag (changes requested):
Re-reviewed `2b5b331` against the OSAC-3141 design with the project rules: greenfield (no legacy/migration concern), fail fast and loud, no defaults/fallbacks, no selective metering, Watch correctness over reconciliation.

**Latest commit:** Watch data-quality errors now fail the stream for every resource type, not only Volume. That is the right fail-loud behavior. Do not copy it back into skips.

**Still blocking**
- Metering must not fail NFS `CreateVolume`. Billability is a mapper predicate, not a provisioner filter.
- The Volume transition table is explicit (missing = reject), but the test still only distinguishes error vs skip. Pin `started` / `suspended` / skip per ordered pair.

**Also fix before merge**
- `AVAILABLE` without vendor identity is a skip, not a data-quality error.
- Feedback still swallows a nil `StateTransitionTime`.
- Resize Watch consume is fine; the produce path still cannot emit it. Do not claim increase-only capacity in this PR.
- Do not route Volume to M360 `/storage/event` until OSAC-4285.

Not re-raised: pre-existing Ready volumes, reconciliation fail-closed on one bad Volume, out-of-order delivery. Those conflict with the greenfield / Watch-first rules.

**Verdict:** Request changes.

PR #1039, opened 2026-09-16
OSAC-3884: template the UI URL into the Keycloak osac-ui client

dcritch:
---

## Problem

The `osac-ui` Keycloak client ships with relative redirect URIs — `"/*"` in
`prerequisites/keycloak/service/files/realm.json`, and a hardcoded
`http://ui.osac.localhost:8080` in `charts/osac-infra/files/realm.json`.
Keycloak does not resolve a relative `redirectUri` against the Route hostname,
so the absolute callback the browser sends is rejected:

```
We are sorry... Invalid parameter: redirect_uri
```

Browser login is broken out of the box until someone patches the client through
the admin API — and that patch is wiped on every Keycloak pod restart while
`keycloak.realmOverwrite: true`. Reproduced on RHDP demo labs, bare metal, and
vmaas-ci.

## Fix

Substitute the browser-facing UI URL into the client's `rootUrl`,
`redirectUris` and `webOrigins` at realm-import time, via a `__OSAC_UI_URL__`
placeholder and the `resolve-realm-secrets` initContainer that already resolves
the client secrets and realm admin credentials.

| File | Change |
|---|---|
| `charts/osac-infra/files/realm.json` | `osac-ui` client built from `__OSAC_UI_URL__` |
| `prerequisites/keycloak/service/files/realm.json` | same (identical copy kept in sync) |
| `charts/osac-infra/files/hooks/resolve-realm-secrets.sh` | substitute the placeholder; reject a non-absolute URL |
| `charts/osac-infra/templates/keycloak/resources.yaml` | pass `OSAC_UI_URL` into the initContainer |
| `charts/osac-infra/values.yaml` | new `keycloak.uiUrl` |
| `prerequisites/keycloak/service/deployment.yaml` | same wiring for the static reference manifest |
| `Makefile` | derive `keycloak.uiUrl` and pin `ui.externalHostname` on OpenShift |

Notes:

- The resolver **fails** on anything that is not an absolute
  `scheme://host[:port]` rather than importing a realm that cannot
  authenticate.
- `make install-osac` now pins `ui.externalHostname`. The `osac-ui` chart names
  its Route `osac-ui` in the instance namespace, so this is exactly the host
  OpenShift would auto-assign — pinning it just makes the value knowable one
  install phase earlier, where Keycloak needs it.
- The chart default targets the kind `dev-full` UI HTTPRoute, so that path is
  unchanged.

## Regression guard

No CI job drives the `osac-ui` browser authorization-code flow: CI leaves
[...]

PR #1040, opened 2026-09-16
OSAC-3887: assign profile and email scopes to the Keycloak osac-ui client

dcritch:
---

## Problem

The `osac-ui` Keycloak client sets its own `defaultClientScopes`, which fully
overrides the realm-level `defaultDefaultClientScopes` — so `profile` and
`email` being realm defaults does not put them on the client:

```json
"defaultClientScopes": ["groups", "basic", "username", "organization", "roles", "osac-api"]
```

The shipped `osac-ui` image always requests `openid profile email
[organization]`, and Keycloak rejects an authorization request naming any scope
the client is not assigned. The browser comes back to
`/callback?error=invalid_scope` and the UI never establishes a session — it
just loads forever.

Reproduced on RHDP demo labs, bare metal, and vmaas-ci. For contrast,
`osac-admin`, `osac-controller` and `osac-csi-driver` all already carry
`profile`+`email`; only `osac-ui` is out of sync with what the UI build
requests.

## Fix

Add `profile` and `email` to the `osac-ui` client's `defaultClientScopes` in
both `realm.json` copies:

- `charts/osac-infra/files/realm.json`
- `prerequisites/keycloak/service/files/realm.json`

Both scopes are already defined in the realm's `clientScopes`, so this only
assigns existing scopes — no new mappers or claims are introduced. This is the
same list the field workaround applies through the admin API, except it is
durable across pod restarts (the admin-API patch is wiped whenever
`keycloak.realmOverwrite: true` re-imports the realm).

### On the open design question

The ticket left two directions open: fix Keycloak, or change the UI to stop
requesting `profile`/`email`. The `osac-ui` chart exposes no scope
configuration — scopes are baked into the image — so only the Keycloak side is
actionable in this repository. The bare-metal and vmaas-ci reproductions both
postdate the UI-side minimal-scopes change, so the client config still needs to
match what the shipped image asks for.

`osac-cli` is also missing `profile`/`email`, but it authenticates through the
localhost grant without requesting them, so it is left alone here.

## Regression guard

No CI job drives the `osac-ui` browser authorization-code flow: CI leaves
Keycloak at the in-cluster default with no external route, and the e2e suites
are API-only via the `osac-cli` localhost grant. That is why this never
surfaced in CI.

`scripts/validate-keycloak-ui-client-scopes.sh` is added to `make
helm-validate` as a cluster-free guard. It fails if the `osac-ui` client loses
a scope the UI requests, or references a scope the realm does not define —
which the import would drop silently and reproduce the identical
`invalid_scope` failure. It checks both committed `realm.json` copies and the
realm the rendered chart actually ships.

Verified it fails both when `profile`/`email` are removed and on a typo'd scope
name.

When the UI starts requesting a new scope, add it to both `realm.json` copies
and to `UI_REQUIRED_SCOPES` in that script.

## Testing

- `yamllint --strict osac-installer/` — passes
- `helm template` for every profile's `infra.yaml` — passes
- `scripts/validate-keycloak-ui-client-scopes.sh` — passes, and fails on both reintroduced defects

Two `osac-installer` gates are already red on `main` and are **unchanged** by
this PR:

- `make helm-validate` stops at its `helm-lint` prerequisite (the `osac` chart
  schema requires non-empty `service.externalHostname`/`internalHostname`,
  which `values.yaml` ships empty). The `helm-validate` body was run manually
  instead.
- `scripts/validate-keycloak-credentials.sh` reports 8 failures (its stub `oc`
  predates the `osac-csi-driver` secret from OSAC-4261, and the
  `keycloak-set-passwords` Job is not rendered by a default `helm template`).

`pre-commit` is not installed in the environment this was developed in, so the
hooks were not run.

## Related

- [OSAC-3884](https://redhat.atlassian.net/browse/OSAC-3884) — the same client also ships relative `redirectUris`; same class of
  realm-client drift, fixed in a separate PR. Merging both produces two trivial
  add/add conflicts (the `helm-validate` line in the Makefile and adjacent doc
  sections); `realm.json` merges cleanly.


dcritch:
**AI assistance disclosure:** this change was implemented with the help of Claude Code (Claude Opus 5). The commit carries an `Assisted-by: Claude Code <noreply@anthropic.com>` trailer.

Please review with the usual scrutiny — particularly the realm.json edits and the new `make helm-validate` guard script.

rawagner:
UI requests only `openid` and `organization` - https://github.com/osac-project/osac-ui/pull/155 - a month old change. It seems that 3887 was reopened due to testing with old UI

dcritch:
Closing this — the reviewer is right that [OSAC-3887](https://redhat.atlassian.net/browse/OSAC-3887) was fixed in
osac-project/osac-ui#155, and this PR fixes the wrong layer. Details, since the
residue is still a real installer bug:

### osac-ui #155 shipped in v0.0.6

`BuildAuthorizeURL` in `proxy/auth/oidc.go`:

| osac-ui | tagged | `scope` |
|---|---|---|
| `v0.0.5` | 2026-07-15 | `openid profile email` |
| `v0.0.6` | 2026-08-19 | `openid organization` |

`git tag --contains aba4753a0` (the #155 merge commit) returns `v0.0.6`, so the
minimal-scopes change is in that release. The realm's `osac-ui` client already
carries `organization`, so against v0.0.6 nothing is missing and the scope
change proposed here is unnecessary. It would also permanently re-grant the
claims #155 deliberately stopped requesting.

### Why it still reproduces in the field

The installer is internally inconsistent. `charts/osac/Chart.yaml` pins the
osac-ui **chart** at `0.0.6` (bumped in #617 on 2026-08-30), but three profiles
still pin the **image** at the pre-#155 build:

- `values/bmaas-ci/instance.yaml:128`
- `values/vmaas-ci/instance.yaml:140`
- `values/full-ci/instance.yaml:127`

All three read `ghcr.io/osac-project/osac-ui:v0.0.5`. They were added in
f88e78052 on 2026-08-23 — already stale on arrival, since v0.0.6 tagged four
days earlier. `values/dev/instance.yaml` uses `:latest`, which is why dev never
hit this.

That matches the reproductions on the ticket (2026-09-03 bare metal, 2026-09-15
vmaas-ci): those deployments run `v0.0.5`, which requests `profile email`
against a client that is not assigned them, so Keycloak answers
`error=invalid_scope`. The claim in this PR's description — that the
reproductions postdate the UI-side fix, so the client config still needs to
match — was wrong. They postdate the fix, but not the image those profiles
actually deploy.

### Suggested follow-up

Bumping those three profiles to `v0.0.6` so the image matches the chart
dependency is the actual installer-side fix, plus a guard that the pinned image
tag tracks the `osac-ui` chart dependency version so this drift class cannot
recur. Happy to open that as a separate PR if it is wanted here rather than on
the ticket.

#1039 (OSAC-3884, `Invalid parameter: redirect_uri`) is unaffected by any of
this — that one is a genuine realm-side defect at any UI version.

---

Analysis assisted by Claude Code; the osac-ui tag/scope claims above are
reproducible with `git tag --contains aba4753a00e69d4663ecba4ab9bdbf987889e4d1`
and `git show v0.0.5:proxy/auth/oidc.go` in `osac-project/osac-ui`.

PR #1041, opened 2026-09-16
OSAC-5357: fix run-detection using unreliable gh run list --branch/--event filters

eliorerz:
## Summary

Follow-up to #1024 / #1031 / #1036. Confirmed live on a real release dispatch (`osac/v0.0.11`, `fulfillment-service=0.0.101`): all three of fulfillment-service's downstream publish workflows genuinely succeeded within minutes of the tag push -- https://github.com/osac-project/osac/actions/runs/35107213991 (publish-image), and the equivalent publish-binaries/publish-proto runs, all `conclusion: success`.

But `wait_for_component_publish_workflow_run`'s own detection query -- `gh run list --workflow X --branch <tag> --event push` -- never found `publish-image`'s run across 45 straight minutes of polling every 15s, until it hit its own timeout and failed the whole release (job https://github.com/osac-project/osac/actions/runs/35105132454/job/104831298002): `publish-image.yaml never started for tag fulfillment-service/v0.0.101 within 2700s`.

The exact same command found the run instantly when run manually well after the fact. This looks like a real reliability gap in `gh`'s server-side `--branch`/`--event` filtering (or the underlying GitHub search-index-backed endpoint it uses) when queried very soon after the triggering push, not a logic bug in when/how we call it.

## Fix

Replaced the server-side filtered query with an unfiltered `--json databaseId,headBranch,event` listing of the workflow's last 30 runs, piped to a real `jq` (not `gh run list`'s own `--jq`, which doesn't support `--arg`) matching the tag client-side.

## Test plan

- [x] `bash -n` syntax check
- [x] Verified the fixed query against the real repo, for the exact run that was missed: finds it immediately
- [x] Functional test: genuine match found immediately; genuine no-match still times out and fails cleanly
- [ ] Re-dispatch the same verification release once this merges to confirm a full clean pass

## Summary

- **CI:** Updated `wait_for_component_publish_workflow_run` to record its start time and search runs with limits of 30, 60, 120, up to 1000.
- **CI:** Replaced server-side `gh run list --branch --event` filters with client-side `jq` matching for the tag branch and `push` event.
- **CI:** Stops pagination when it finds a match, reaches the available runs, or finds runs older than the search start time.
- **Tests:** Reported coverage includes syntax checks, a real missed run, matches beyond the first page, no-match timeouts, and low-activity and busy workflows. Test execution is not independently verified here.
- **Other areas:** No API, controller, database, authentication, deployment, or documentation changes.
- **Compatibility:** No public interface changes. CI run selection behavior changes. Busy workflows can cause additional `gh run list` requests, capped at 1000 runs.

## Risk classification

**risk:show** was applied because the change affects CI workflow detection and publish-run handling, without changing application APIs, authentication, data storage, or production deployment logic.

The labeling criteria were not supplied. Therefore, specific criteria and proximity to another classification cannot be verified.


eliorerz:

A fixed `--limit 30` could reintroduce the same class of false-negative this PR fixes, just under a different trigger: fulfillment-service's own `publish-image.yaml`/`publish-binaries.yaml`/`publish-proto.yaml` also fire on every PR touching that path (not just the release tag push), so on a busy day the real match could genuinely get pushed past a small fixed page before we find it.

Replaced the fixed page with geometric growth (30, 60, 120, ... capped at 1000), stopping once the match is found, the workflow genuinely has fewer runs than the current page (nothing more to look at), or the oldest run already returned predates this function's own start time -- which means it provably cannot be the run our own tag push just triggered, so paging further back is pointless.

Tested 3 scenarios: a match sitting past the initial 30-run page (found by growing to 60), a low-traffic workflow with only a couple of runs (stops immediately, no false timeout), and a persistently busy workflow with no real match at all (caps growth at 1000 and gives up cleanly for that poll interval instead of hanging or looping forever).

Validated: `bash -n` syntax check, functional tests above. Pushed as a new commit.

PR #1044, opened 2026-09-16
OSAC-5357: surface gh run list failures instead of swallowing them

eliorerz:
## Summary

Follow-up to #1024 / #1031 / #1036 / #1041. Confirmed live on a real release dispatch (`osac/v0.0.12`, `fulfillment-service=0.0.102`): the same "Waiting up to 2700s..." hang happened again, even with #1041's own run-detection fix (unfiltered listing + geometric pagination) already merged. The underlying `publish-image.yaml` run genuinely existed and succeeded again (run 35119587892), just like the previous two failures.

Both the old `--branch`/`--event`-filtered query and the new unfiltered/paginated query failing identically on the same real data points at something more fundamental than query filtering -- most likely `gh` itself failing (auth, rate limit, or something specific to this job's `runs-on: osac-ci` self-hosted runner). There is no prior precedent for `gh run list`/`gh run watch` running on this specific runner type anywhere else in this file.

## The actual bug this fixes

Any `gh run list` failure was silently swallowed (`2>/dev/null` and `|| runs='[]'`), making a real command failure indistinguishable from a genuine "no match yet" in the logs. Every previous live failure has been diagnosed purely by inference, since the actual `gh` error was never visible.

## Fix

Capture `gh run list`'s stderr and log it explicitly as a `::warning::` when the command fails, instead of discarding it. This doesn't change the polling/retry behavior at all -- it just means the next live failure (if this same issue recurs) will show the real underlying cause instead of another uninformative timeout.

## Test plan

- [x] `bash -n` syntax check
- [x] Functional test: a genuinely failing `gh run list` now surfaces a `::warning::` with the real error and still fails cleanly after timeout; the success path is unchanged
- [ ] Re-dispatch the same verification release once this merges -- if the hang recurs, we'll finally see why

## Summary

- **CI/release workflow:** Updated `wait_for_component_publish_workflow_run` in `osac-installer/scripts/nightly-charts.sh`.
- Captures `gh run list` errors and logs sanitized GitHub Actions warnings.
- Preserves empty results, polling, and retry behavior.
- No API, controller, database, authentication, deployment, or public interface changes.

## Tests and verification

- Syntax and functional tests were completed.
- Post-merge verification release is pending.

## Backward compatibility

Successful polling and retry behavior remain unchanged. Failed `gh run list` commands now produce visible warnings instead of silent empty results.

## Risk classification

**risk:show** — The change affects CI release error reporting and preserves existing polling behavior. No application runtime or public interface changes are reported.

The supplied context does not define criteria for the risk labels. The change is not classified as `risk:ship` because post-merge verification remains pending. No evidence establishes that it was close to `risk:ask`.


eliorerz:

`gh_err` was being echoed directly into a `::warning::` workflow command without sanitization. This codebase already has `_gha_sanitize_for_message` for exactly this purpose (strips newlines, carriage returns, escape sequences, and `::` sequences), used throughout the file -- my new code just didn't route through it. Without that, a newline in gh's stderr would break the message across multiple raw log lines, and an embedded `::`-prefixed sequence could be misread as an unrelated workflow command by the Actions log processor.

Tested with a deliberately malicious/messy multi-line error containing embedded fake workflow commands (`::error::...`, `::set-output::...`): output now collapses to a single safe warning line, with none of the injected commands surviving as their own log lines.

Validated: `bash -n` syntax check, functional test above. Pushed as a new commit.

PR #1052, opened 2026-09-16
NO-ISSUE: Replace gh CLI with plain curl for run detection and watching

eliorerz:
## Summary

Follow-up to #1024 / #1031 / #1036 / #1041 / #1044. Found the actual root cause behind three separate live failures of the exact same "Waiting up to 2700s..." symptom, each time with the underlying publish workflow genuinely succeeding.

**Root cause**: the "Tag and notify" job's `runs-on: osac-ci` self-hosted runner does not have the `gh` CLI on `PATH` at all -- confirmed directly from the job's own live output: `osac-installer/scripts/nightly-charts.sh: line 236: gh: command not found`. Every `gh run list`/`gh run watch` call in `wait_for_component_publish_workflow_run` was failing instantly with that error, and the failure was being swallowed into an empty result (indistinguishable from "no match yet" in the logs at the time), which is exactly why this looked like a silent ~45-minute hang instead of an obvious, immediate error. All of the previous fixes in this series (unfiltered listing, pagination, error surfacing) still depended on the same missing binary and would have failed identically -- confirmed live again after #1044 merged, with zero `gh`-related warnings ever appearing despite the underlying publish succeeding, which in hindsight was the real tell.

## Fix

Replaced both the listing and the watch loop with plain `curl` calls directly against the GitHub REST API, authenticated via `GH_TOKEN` as a Bearer token. `curl` is a far safer baseline assumption for any runner than a CLI tool that turns out not to be installed.

Also fixed two related gaps found while rewriting this:
- A `curl` failure that produces no output (missing binary, DNS failure) previously left the captured error message blank -- silently recreating the exact same class of invisible failure this whole effort exists to fix. Now always includes the exit code even when there's no accompanying output.
- The watch loop had no timeout of its own and would retry forever on a persistent failure, relying entirely on the outer job's 90-minute timeout with zero diagnostic output in the meantime. Bounded it by the same timeout budget as the listing phase, with the same error surfacing.

## Test plan

- [x] `bash -n` syntax check
- [x] 6 scenarios tested against a mocked `curl`: match found immediately, genuine no-match (clean timeout), the exact real failure mode (command exits with no output), an HTTP error response, a completed run with a failing conclusion, and the watch loop's own failure path staying bounded instead of hanging
- [ ] Re-dispatch the same verification release once this merges -- this should be the one that actually works end to end

## Summary

- **CI and deployment:** Replaces unavailable `gh` CLI calls with authenticated GitHub REST API requests through `curl`.
- **Workflow detection:** Searches up to 10 pages of workflow runs and matches tag-push runs client-side.
- **Polling:** Uses bounded connection and request timeouts. Shares one timeout budget across detection and polling.
- **Error handling:** Reports curl, HTTP, JSON, malformed response, invalid response shape, timeout, and failed workflow-conclusion errors.
- **Authentication:** Stores `GH_TOKEN` in a protected temporary curl configuration. The token is not exposed in process arguments.
- **Tests:** Reported validation includes `bash -n` and mocked matching, no-match, command-failure, HTTP-failure, malformed-response, pagination, token-exposure, failed-conclusion, and bounded-polling cases. End-to-end verification remains pending after merge.
- **API surface:** Keeps the existing shell function interface. No public declarations changed.
- **Backward compatibility:** Removes the `gh` binary requirement. The script now requires `GH_TOKEN`, `GH_REPO`, `curl`, `jq`, and GitHub API access. Database, controller, and documentation areas are unchanged.

## Risk classification

The applied risk label is unavailable. The supplied context does not define criteria for `risk:ship`, `risk:show`, or `risk:ask`. A close alternative cannot be determined.


eliorerz:
Verified all four findings against the current code -- all confirmed real, fixed, and amended into the same commit.

**jq calls not guarded against `set -e` (line 310 area) -- confirmed, fixed.** This script runs under `set -euo pipefail`. The watch loop's `run_status=$(jq -r '.status // empty' <<<"${run_body}")` operates on a response body with no prior validation -- unlike the listing loop's jq calls, which only ever run against `runs`, already guaranteed valid JSON by construction (either real parsed output or the literal `'[]'`). A malformed `run_body` would make `jq` exit non-zero and, as a bare assignment (not part of an `if`/`&&`), silently kill the whole job instead of going through this function's own error handling. Guarded both calls; a parse failure now logs a `::warning::` with the sanitized body and lets the loop retry.

**Missing curl timeouts -- confirmed, fixed.** Neither curl call had `--connect-timeout`/`--max-time`, so a hung connection (not a clean failure) could block indefinitely regardless of the surrounding loop's own timeout bookkeeping. Both calls now derive `--connect-timeout`/`--max-time` from the remaining poll budget (floored at 5s so a nearly-exhausted budget doesn't hand curl a 0/negative value).

**GH_TOKEN exposed via argv -- confirmed, fixed.** `-H "Authorization: Bearer ${GH_TOKEN}"` puts the literal token in this process's command line, visible via e.g. `ps aux` to anything else on the machine while curl runs -- a real concern given this job's runner (`osac-ci`) is a persistent, reused self-hosted box, not an ephemeral one. Switched to a `curl -K <config-file>` with the header written to a `chmod 600` temp file, cleaned up automatically via a `RETURN` trap on every exit path of the function.

**start_time-based pagination cutoff -- confirmed and removed.** The comparison assumed the runner's local clock and GitHub's server-recorded `created_at` timestamps agree closely enough to safely rule out earlier pages. Given this runner has already shown it isn't a standard environment (missing `gh` CLI entirely), clock skew making that comparison silently wrong felt like a real risk, not just theoretical. Now paginates all the way to the existing 10-page cap, relying only on "fewer than a full page" as the early-exit condition.

Tested all four together: token never appears in curl's argv (only in the `-K` file, verified by inspecting the mocked call's actual arguments); both curl calls carry `--connect-timeout`/`--max-time`; a malformed response body is handled cleanly under `set -e` (confirmed by literally running the extracted function under `set -euo pipefail`) instead of crashing the script; and a match sitting on page 2 behind a full page of old, unrelated runs is still found now that the early-stop is gone.

Validated: `bash -n` syntax check, functional tests above. Amended into the existing commit and force-pushed.

eliorerz:
Verified -- confirmed real, fixed, amended into the same commit.

The watch loop was starting its own fresh `watch_start=${SECONDS}` timer instead of sharing the listing phase's `start`. That means the function's real worst-case runtime was up to 2x `timeout` (a full budget for finding the run, then another full budget for watching it finish) -- not the single `timeout` its own opening message ("Waiting up to ${timeout}s...") promises. With multiple workflows verified sequentially per component (fulfillment-service alone has 3), that compounds fast against the job's 90-minute overall timeout.

Removed `watch_start` entirely and made both phases share the same `start` timestamp, so `remaining` and the timeout check in the watch loop are computed against the same clock as the listing phase.

Tested directly: a scenario where listing takes ~1s of a 4s budget and the watch phase never completes now fails at ~4s total (the shared budget), not ~7-8s (which is what two independent fresh 4s budgets would have produced).

Validated: `bash -n` syntax check, functional test above. Amended into the existing commit and force-pushed.

eliorerz:
Verified -- confirmed real, fixed, amended into the same commit.

Reproduced directly: a 200 response with `.workflow_runs` present but not an array (e.g. a string) passes through the existing `.workflow_runs // []` unscathed, since `//` only substitutes for `null`/`false`, not "wrong type but truthy". `runs` then ends up holding a non-array value, and the candidate-selection jq call two lines down fails outright (`jq: error: Cannot iterate over string`, exit 5). As a bare assignment under `set -euo pipefail` (this script's own mode), that would have killed the job instead of retrying.

Fixed by explicitly validating the type inside the same jq call that already extracts `.workflow_runs`, using `error(...)` to fail the whole expression (and therefore hit the existing `||` fallback) when it isn't an array -- routes through the exact same `gh_err`/sanitized-warning/retry path as an actual JSON parse failure, no new code path needed.

Tested: the exact failure shape (`{"workflow_runs": "totally unexpected shape"}`) now surfaces a `::warning::` and times out cleanly instead of crashing the script under `set -e`; confirmed the normal valid-array case is unaffected.

Validated: `bash -n` syntax check, functional tests above. Amended into the existing commit and force-pushed.

eliorerz:
Verified -- partially fixed, one part explicitly skipped with reasoning below.

**Registered immediately after mktemp, before chmod/writing the token -- fixed.** And extended to cover `EXIT`, not just `RETURN`: this job's runner is a persistent, reused box, so a leaked temp file holding a bearer token surviving a killed/cancelled job (not just a normal return) is a real concern, not just cleanliness.

**Restoring a caller's pre-existing RETURN trap -- investigated, not implemented, here's why.** I tried this and it doesn't actually work the way it looks like it should: bash's `RETURN` trap is *not* properly function-scoped for save/restore purposes. Confirmed empirically with a minimal outer/inner function test -- an inner function's `trap ... RETURN` permanently clobbers an outer function's previously-set `RETURN` trap (it fires the inner one again when the outer function eventually returns, never restoring the original), *and* `trap -p RETURN` queried from inside the inner function doesn't expose the outer function's trap at all, so there's nothing reliable to snapshot and restore in the first place. A naive save/restore attempt (`trap -p` + string surgery) produced incorrect behavior in testing -- the "restored" trap never fired.

Checked whether this is currently a live risk: it isn't -- grepped this whole file and its calling workflow step, and this is the only `trap ... RETURN` anywhere in the chain today. Given that, shipping a save/restore mechanism I'd just watched fail in testing, to protect against a conflict that doesn't currently exist, seemed worse than the gap itself. Left an explicit comment instead, so whoever adds a `RETURN` trap around a future caller of this function knows to watch for this rather than silently losing it.

Tested the parts I did fix: the temp file is removed on a normal function return (no leftover `/tmp/tmp.*` file), and -- the scenario the `EXIT` addition specifically targets -- it's *also* removed when the function is sent `SIGTERM` mid-run, before it ever gets a chance to return normally.

Validated: `bash -n` syntax check, functional tests above. Amended into the existing commit and force-pushed.

PR #1054, opened 2026-09-16
NO-ISSUE: Scope unit/integration test jobs to changed components

eliorerz:
## Summary
- `unit-tests.yml`/`integration-tests.yml` gated every component's job on one blanket "code changed" filter -- a PR touching only `osac-aap/` still ran `fulfillment-service`'s, `osac-operator`'s, and `osac-metering`'s jobs too (and vice versa). `helm-lint.yaml`/`check-floating-tags.yaml` already solved this correctly with per-component `paths-filter` outputs; this mirrors that proven pattern into the two files that still lacked it.
- Cross-component edges (not just each component's own directory) are real Go module dependencies traced from `go.work` + each component's `go.mod` `require`/`replace` directives, not guesses: `fulfillment-service` imports `bare-metal-fulfillment-operator` and `osac-operator/api`; `osac-operator` and `bare-metal-fulfillment-operator` import each other's whole module. Missing one of these would reproduce exactly the class of bug [OSAC-4066](https://redhat.atlassian.net/browse/OSAC-4066) found in the separate graphify `ci-select` effort: a real dependency silently absent from a filter, causing a false skip.
- `osac-installer`'s integration-test job and `osac-aap`'s jobs are deliberately left unscoped: `osac-installer`'s job deploys the full stack from a PR-built `fulfillment-service` image (a real umbrella integration test, not narrowly scoped by design), and `osac-aap` has no Go module coupling to narrow in the first place.
- Job `name:` values are untouched -- confirmed via the live `ci-status-checks` ruleset that required-check matching is by literal job name, not id; this only adds a second `if:` condition ANDed onto the existing one.

Note: a separate, more ambitious effort (Epic [OSAC-4013](https://redhat.atlassian.net/browse/OSAC-4013) / OSAC-4021, graph-BFS based `graphify ci-select`) targets similar CI-skip goals but is currently blocked on [OSAC-4066](https://redhat.atlassian.net/browse/OSAC-4066) and stays observational-only by design. This PR is independent of that effort -- deterministic path filters, real skips, no LLM/graph-BFS risk.

## Test plan
- [x] `actionlint .github/workflows/unit-tests.yml .github/workflows/integration-tests.yml` -- clean
- [x] `yamllint --strict` on both files -- clean
- [x] `pre-commit run --files .github/workflows/unit-tests.yml .github/workflows/integration-tests.yml` -- clean
- [ ] After merge, confirm on a real PR touching only one side of a cross-component edge (e.g. `bare-metal-fulfillment-operator/**` only) that both `fulfillment-service` and `osac-operator` integration-test jobs still fire, while an unrelated component (e.g. `osac-aap`) correctly skips
- [ ] Re-confirm the existing README-only case still skips everything (no regression)

## Summary

- **CI:** Added per-component path filters to unit and integration test workflows.
- Filters include component paths, cross-component Go module dependencies, and workflow changes.
- Existing `should-run` conditions remain required.
- `osac-installer` and `osac-aap` jobs remain unscoped.
- Job names remain unchanged.
- **Tests:** Actionlint, yamllint, and pre-commit checks passed. Post-merge validation remains planned.
- **API, controllers, database, auth, deployment, and documentation:** No changes.

## Compatibility

No runtime or API behavior changes are expected. CI job execution changes only for pull requests and merge groups. Other event types bypass component filters.

## Risk classification

**risk:ship** — The change is limited to CI path filtering, preserves existing job names and conditions, and passed static validation checks. Labeling instructions were not supplied, so no formal alternate-classification criteria are available. The change does not appear close to `risk:show` because it does not alter runtime behavior.


eliorerz:
Addressed the review findings (verified each against current code before applying):

- **Fixed**: added `osac-installer/**` to the `fulfillment-service`/`osac-operator`/`bare-metal-fulfillment-operator` integration filters -- confirmed all three jobs' setup/test steps run through `make -C osac-installer install-infra`/`make -C osac-installer test`, so an osac-installer-only change could break them undetected.
- **Fixed**: added `go.work`/`go.work.sum` to every Go component's filter in both files -- a workspace-level change can affect every module in `use` without touching any single component's directory.
- **Not applied**: the suggestion to drop `osac-aap-relevant` back to a bare `should-run` check. Checked the job's own steps -- it only ever touches `osac-aap/**` (uv/pytest, no osac-installer or cross-component dependency), so that would just reintroduce the exact waste this PR removes, not "preserve existing behavior."

Pushed in f251a0e5b.

PR #1057, opened 2026-09-16
NO-ISSUE: Fix unbound gh_curl_config variable in EXIT trap

eliorerz:
## Summary

Follow-up to #1052. Confirmed live on a real release dispatch (`osac/v0.0.14`, `fulfillment-service=0.0.104`): the actual [OSAC-5357](https://redhat.atlassian.net/browse/OSAC-5357) verification logic worked end to end for the first time -- all three of fulfillment-service's downstream publish workflows were correctly detected and watched to success. But the job still failed right after with:

```
osac-installer/scripts/nightly-charts.sh: line 394: gh_curl_config: unbound variable
```

## Root cause

The cleanup trap added in #1052 was registered with single quotes: `trap 'rm -f "${gh_curl_config}"' RETURN EXIT`. Single-quoting means `${gh_curl_config}` is a variable reference, re-evaluated whenever the trap actually fires -- not a value captured at registration time.

`RETURN` fires while still inside the function (so `gh_curl_config`, a `local` variable, is still in scope there -- that part worked correctly). `EXIT` fires at process termination, which can be long after this function has already returned and its local scope is gone. At that point, referencing `${gh_curl_config}` under `set -u` is a hard error.

## Fix

Double-quoted the trap string instead: `trap "rm -f '${gh_curl_config}'" RETURN EXIT`. This expands `${gh_curl_config}` immediately, at the moment `trap` is invoked, baking the literal path into the registered trap command -- independent of whatever scope is active whenever the trap eventually fires.

## Test plan

- [x] `bash -n` syntax check
- [x] The exact reported scenario (function returns normally, then the whole script exits) no longer errors
- [x] Normal-return cleanup still removes the temp file
- [x] SIGTERM-during-run cleanup still removes it too
- [ ] Re-dispatch the same verification release once this merges -- this really should be the clean one now

## Summary

- **CI/deployment:** Fixed cleanup in `osac-installer/scripts/nightly-charts.sh`.
- The trap now shell-escapes and captures the temporary curl-config path when registered.
- This prevents an `gh_curl_config` unbound-variable error under `set -u`.
- **Security:** The script continues to store the bearer token in a temporary curl config instead of a command-line argument.
- **API, controllers, database, auth, documentation:** No changes.
- **Tests:** Syntax validation and cleanup tests for normal returns and `SIGTERM` were completed.
- **Backward compatibility:** No API or configuration compatibility impact is expected.

## Risk classification

**risk:ship** — Applied because the change is limited to cleanup-trap handling in one CI script and does not change APIs, data, authentication behavior, or deployment configuration. Formal labeling criteria were not supplied. The change does not qualify as **risk:show** or **risk:ask** based on the available scope and evidence.


eliorerz:
Verified -- confirmed real, fixed, amended into the same commit.

Reproduced directly: with `gh_curl_config` containing a single quote (e.g. `/tmp/fo'o/bar`, simulating an unusual `TMPDIR` on this already-nonstandard runner), the naive `trap "rm -f '${gh_curl_config}'" ...` form produces a genuine syntax error when the trap fires (`unexpected EOF while looking for matching ''`) -- the trap string is re-parsed as shell code at firing time, not just substituted literally, so a stray quote in the value breaks out of the intended quoting.

Fixed by using `printf -v quoted_gh_curl_config '%q' "${gh_curl_config}"` to properly shell-escape the path before embedding it in the trap string, instead of hand-wrapping it in single quotes.

Tested: the adversarial case (single quote in the path, via `TMPDIR`) no longer errors on normal return; confirmed no regression on the normal (non-adversarial) path -- temp file still created and cleaned up correctly.

Validated: `bash -n` syntax check, functional tests above. Amended into the existing commit and force-pushed.

PR #1077, opened 2026-09-18
OSAC-5311: make volume provider the sole routing key

zszabo-rh:
## Summary

- Remove `VolumeStatus.backend` and make `VolumeStatus.provider` the sole persisted dispatch/routing key.
- Propagate provider-only routing through fulfillment reconciliation, operator feedback/provisioning, and CSI vendor selection; regenerate the shared proto and Volume CRD artifacts.
- Update tests and integration fixtures; no metering change was required because this checkout has no `VolumeStatus.backend` consumer.

## Jira

https://redhat.atlassian.net/browse/OSAC-5311

## Test plan

- [x] `make -C proto generate` and `make -C proto lint`
- [x] fulfillment-service: full internal Ginkgo suite (100 suites) and focused public-volume server suite (29 specs)
- [x] osac-operator: lint, build, envtest, manifest/CRD generation, and Helm lint
- [x] osac-csi-driver: lint, build, unit, and CSI sanity tests
- [x] `pre-commit run --all-files` (all hooks passed; the EOF hook was skipped only for immutable bootstrap-managed `.codex` files, and passed on all changed files)
- [x] Performance, security, and simplification pre-PR reviews

---

_This PR description was drafted with AI assistance ([create-pr](https://github.com/osac-project/osac-workspace/tree/main/skills/create-pr) v0.2.0). Review for accuracy_

## Summary

- Removed persisted `VolumeStatus.backend` from the proto, Go API types, and Volume CRD.
- Propagated provider-only routing through fulfillment, operator provisioning, feedback, deletion, and CSI selection.
- Updated generated artifacts, documentation, tests, and integration fixtures.
- Consumers of persisted `status.backend` must migrate to `status.provider`.

## Affected areas

- **API surface:** Removed `VolumeStatus.backend` and related backend fields. The CRD now exposes `status.provider`.
- **Controllers:** Fulfillment and operator controllers use provider for synchronization, routing, provisioning, and deletion.
- **CSI:** CSI selection now uses `status.provider`.
- **Deployment:** Updated CRD schemas and printer columns.
- **Tests:** Updated unit tests, integration tests, and CSI fixtures.
- **Documentation:** Replaced backend terminology with provider terminology.
- **Database and auth:** No changes identified.
- **CI:** No review findings, applied risk label, or test results were available from the supplied evidence.

## Backward compatibility

Consumers of persisted `status.backend` must migrate to `status.provider`. The proto change also renumbers `provider`, so compatibility with existing serialized data requires validation before deployment.

## Risk classification

Risk label: **unavailable**. No applied label or labeling criteria were supplied. The change is not confirmed as `risk:ship`, `risk:show`, or `risk:ask`.


fulfillment-service/internal/servers/private_volumes_server_test.go
```diff
@@ -311,22 +310,22 @@ var _ = Describe("Private volumes server", func() {
 					Status: privatev1.VolumeStatus_builder{
 						State:          privatev1.VolumeState_VOLUME_STATE_AVAILABLE,
 						VendorVolumeId: "vast-vol-123",
-						Backend:        "vast-1",
+						Provider:       "vast-1",
```
akshaynadkarni: Could we make status.provider immutable after its first assignment? This test changes test-provider to vast-1 and expects success, but provider is the routing key used to select the vendor provisioner for later attach and delete operations. A later change can route operations to the wrong storage system and leave the original volume behind. Also, applyVolumeUpdate currently does not merge status.provider before validation, so adding only an immutability check will not catch this masked update. Please merge status.provider, reject changes from an existing non-empty value, and change this test to expect InvalidArgument for a provider change.

proto/private/osac/private/v1/volume_type.proto
```diff
@@ -125,10 +125,6 @@ message VolumeStatus {
   // CSI CreateVolume succeeds. Visible only through the private API.
   string vendor_volume_id = 3 [(cleanapi.field).private = true];
 
-  // Name of the StorageBackend that serves this volume. Resolved during tier resolution at
-  // creation time. Visible only through the private API.
-  string backend = 4 [(cleanapi.field).private = true];
```
akshaynadkarni, line 130: Since OSAC is still pre-production and field 4 is being removed, should we renumber `provider` from field 8 to field 4 before the API is released?

This would keep the VolumeStatus fields sequential:

vendor_volume_id = 3
provider = 4
protocol = 5

Since OSAC is pre-production, I think this is acceptable here. What do you think?

zszabo-rh: agreed, fixing!

fulfillment-service/internal/servers/private_volumes_server_test.go
zszabo-rh: done!

akshaynadkarni (approved):
Thank you @zszabo-rh for addressing the comments. 

/lgtm

PR #1089, opened 2026-09-18
NO-ISSUE: Make osac-ui a real component in the release/nightly pipeline

eliorerz:
## Summary

osac-ui has lived in this repo since #1082, but the pipeline still treated it as an external repo: pinning its already-published `sha-<short>` image (built by a CI that no longer exists standalone), never building it here, and depending on it via an external OCI chart reference instead of the `file://` path every other component uses. This wires it in identically to fulfillment-service/osac-operator/osac-aap/bare-metal-fulfillment-operator/osac-metering/osac-csi-driver.

## Changes

- **Chart dependency**: switched from `oci://ghcr.io/osac-project/charts` (pinned at a stale `0.0.6`) to `file://../../../osac-ui/charts/ui`, matching every sibling. Handled generically now via `MONO_REPO_UMBRELLA_DEPENDENCIES` — no more separate rewrite call just for osac-ui.
- **`osac-build-and-publish.yaml`**: added osac-ui to the component list, the sub-chart map, the chart-publish map, and the Build job's matrix. A new `osac-ui/Makefile` adds `image-build`/`image-push` targets wrapping podman, matching every other component's Makefile-driven build — verified with a real end-to-end build (multi-stage pnpm+Go Containerfile included). Removed ~150 lines of "resolve and gate osac-ui's externally-published SHA" special-casing across prepare/build/publish — it now flows through the exact same generic per-component logic as everyone else.
- **`nightly-charts.sh`/`lib.sh`**: removed 3 functions that only ever existed for osac-ui's old external-pinning path (`check_osac_ui_image`, `resolve_bare_release_tag_at`, `stamp_osac_ui_chart`) — confirmed dead via repo-wide grep. Added an `osac-ui` case to the existing generic `stamp_component_image_refs`, matching fulfillment-service's shape.
- **New `osac-ui-publish-image.yaml`** replaces both dormant files left over from #1082 (`osac-ui-publish-image.yaml`, `osac-ui-publish-charts.yaml`), plus the now-redundant `osac-ui-container-build.yaml`. Modeled on `publish-image.yaml` (fulfillment-service's shape — single image, no generated manifests) rather than the CRD-heavy `build-image.yaml`. Tags as `osac-ui/vX.Y.Z` like everyone else; `IMAGE_NAME` is hardcoded since `github.repository` now resolves to `osac-project/osac`.
- **`publish-charts.yaml`**: this is the second pipeline stage every other component already had that osac-ui was missing — once "Tag and notify" pushes a real `osac-ui/vX.Y.Z` tag, this publishes osac-ui's own standalone chart, same as `publish-service-chart` does for fulfillment-service. Added the matching `.github/scripts/osac-ui-verify-tag-matches-sha.sh` (byte-identical to the existing per-component copies of this script) and removed the redundant one that lived under `osac-ui/.github/scripts/`.
- **Also fixed**: the root `.gitignore` still had a bare `osac-ui/` entry from when it was a bootstrap-cloned sibling checkout (like `osac-ux/`) — since the merge this silently gitignored any *new* file created under `osac-ui/`, which is exactly what happened to the new Makefile until caught here.

## Testing

- `stamp_component_image_refs` and `rewrite_umbrella_mono_repo_dependencies` both verified live against real chart files for osac-ui — correct sub-chart and umbrella stamps, correct `file://` → `oci://` dependency rewrite with the resolved version.
- `osac-ui/Makefile`'s `image-build` target verified with a real end-to-end podman build.
- `yamllint --strict` and `actionlint` pass clean on every modified/added file (actionlint's only output is the pre-existing, unrelated unknown-runner-label note for the self-hosted `osac-ci` label).
- Repo-wide grep confirms no dangling references to any removed function/output remain.

Signed-off-by: Elior Erez <eerez@redhat.com>

## Summary

- **CI and release:** Integrates `osac-ui` into shared version resolution, build matrices, image promotion, chart packaging, and nightly stamping.
- **Deployment:** Changes the umbrella chart to use the local `file://../../../osac-ui/charts/ui` dependency. Shared scripts stamp UI image references in the subchart and umbrella values.
- **Container publishing:** Adds validated `osac-ui/v*` tag handling, image publishing, digest-based signing, and standalone chart publishing.
- **Repository structure:** Tracks `osac-ui/` as repository content. Adds Podman-based `image-build` and `image-push` Make targets.
- **Tests and quality checks:** Existing dedicated `osac-ui` test, lint, and type-check workflows remain path-scoped. Reported validation includes image builds, chart stamping, dependency rewriting, `yamllint --strict`, `actionlint`, and dangling-reference checks.
- **API, controllers, database, auth, and documentation:** No changes reported.

## Compatibility

The release path for `osac-ui` changes from external special-case handling to the shared repository flow. The umbrella chart now requires the tracked local chart path. Consumers that depend on the previous external OCI dependency or workflow names may need updates.

## Risk classification

**risk:show** — The change affects CI, container publishing, chart publishing, image signing, and release-version resolution. The formal labeling criteria were not supplied, so this classification is based on the demonstrated release and deployment scope. Review severity counts are unavailable because no current review findings were provided.

The change does not qualify as **risk:ship** because it changes production release behavior. It does not qualify as **risk:ask** based on the supplied evidence because validation checks are reported and no unresolved review findings are provided.


eliorerz:
Checked both against the actual codebase — both describe patterns copied verbatim from the established sibling template, not something specific to this file:

- The major/minor tagging without a stable-release check: identical in \`publish-image.yaml\`, \`build-image.yaml\`, and \`execution-environment.yml\` (same \`enable=\${{ env.RELEASE_VERSION != '' }}\` on all three).
- The unconditional \`packages: write\` covering both the PR-build and publish jobs: identical in \`publish-image.yaml\`.

Skipping both here — this PR's whole point is matching osac-ui to the existing per-component pattern, so fixing either one only for osac-ui would make it inconsistent with every sibling instead. If these are worth fixing, they'd need a repo-wide pass across every component's publish workflow, not a one-off here.

